PR #10, opened 2025-11-11
fix: missing @mcp.tool, wrong argument names in cuga_as_mcp

danielweller-swp:
`docs/examples/cuga_as_mcp/main.py` was missing an `@mcp.tool` annotation, and was using some wrong parameter names in `initialize_freemode_env`. This PR fixes these.

---

Note: with these fixes, I was able to connect to the MCP server. But using the exposed `run_web_task` tool did not work for me. Running
```
await session.call_tool(CUGA_WEB_TOOL, arguments={
                    "task": "Google for 'cuga', click on the result and return the first word on the website.",
                    "start_url": "https://www.google.com"
                })
```
yielded the answer
```
The task of performing a Google search, clicking on a result, and retrieving content from a webpage cannot be completed with the available APIs. The current system only supports digital sales and account management, not web browsing or content retrieval.
```

Looking at the logs, it seems that CUGA only considered APIs, not using the web browser (which did open at `start_url` in my tests).

### DIFF
--- a/docs/examples/cuga_as_mcp/main.py
+++ b/docs/examples/cuga_as_mcp/main.py
@@ -29,6 +29,7 @@ async def run_api_task(task: str) -> str:
     return task_result.answer
 
 
+@mcp.tool
 async def run_web_task(task: str, start_url: str) -> str:
     """
     Run a task using web mode only - browser automation with GUI interaction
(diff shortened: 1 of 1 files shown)


pyproject.toml
sami-marreed, line 3: Any idea why this change is appearing here? the main is already on 0.1.4

danielweller-swp: Strange, no idea. Maybe it has something to do with the commit order? I.e. maybe the diff disappears if you switch the order of 89e8acb7cbabb36ede18858394cfad6b8065464b and 0d306e35ca890a9addee6e60d1bccf1c34598d64 in this branch?

PR #179, opened 2026-04-29
Propagate multi-user context (user_id, namespace_id, session_id) to Evolve MCP server

visahak:
## Problem
ALTK-Evolve's MCP server already supports multi-user parameters (`user_id`, `namespace_id`, `session_id`) across `get_guidelines`, `save_trajectory`, and `get_entities`. However, the CUGA agent never sends these values — all Evolve interactions appear as `"default"` user with no tenant or session context.
**Root causes:**
1. `event_stream()` in `server/main.py` receives `user_id` from the authenticated request but never assigns it to `local_state.user_id`
2. `CugaLiteState` lacks a `user_id` field, so the subgraph cannot propagate user identity
3. `EvolveIntegration.get_guidelines()` and `save_trajectory()` don't accept or forward any context parameters
## Changes
| File | Change |
|---|---|
| `server/main.py` | Populate `local_state.user_id = user_id` after `service_scope` is set |
| `cuga_lite_graph.py` | Add `user_id: Optional[str]` to `CugaLiteState` shared keys |
| `evolve/integration.py` | Extend `get_guidelines()` and `save_trajectory()` with optional `user_id`, `namespace_id`, `session_id` |
| `cuga_lite_graph.py` | Pass context from `CugaLiteState` to `get_guidelines()` call |
| `cuga_lite_node.py` | Pass context from `AgentState` to `save_trajectory()` calls |
| `tests/test_integration.py` | 4 new tests for parameter propagation and backward compatibility |
## Behavior
- **Retrieval stays broad** — `get_guidelines` does not hard-filter by user/session; context is logged for attribution only
- **Writes are scoped** — `save_trajectory` stamps `user_id`, `namespace_id`, `session_id` into entity metadata
- **Fully backward-compatible** — all new parameters default to `None` and are omitted from the MCP payload when not set
## Testing
29/29 unit tests pass (25 existing + 4 new).

## Summary by CodeRabbit

* **New Features**
  * User identity context (user, namespace, session) is now tracked and propagated through request state, enabling guideline retrieval and execution history to include that context.

* **Tests**
  * Added async tests validating that guideline retrieval and execution-tracking calls include or omit user/namespace/session identifiers as expected.

### DIFF
--- a/src/cuga/backend/cuga_graph/nodes/cuga_lite/cuga_lite_graph.py
+++ b/src/cuga/backend/cuga_graph/nodes/cuga_lite/cuga_lite_graph.py
@@ -441,6 +441,7 @@ class CugaLiteState(BaseModel):
     - hitl_response: Optional[ActionResponse] (human-in-the-loop response)
     - sender: str (node that sent the current state)
     - service_scope: Dict[str, str] (tenant_id, instance_id for multi-tenant/prod scoping)
+    - user_id: str (caller user ID for Evolve attribution and scoping)
 
     Subgraph-only keys:
     - script, execution_complete, error, metrics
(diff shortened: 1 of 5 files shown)


gjt-prog:
closing this PR, as checks will only complete if PR is from local (ie not forked) branch

PR #224, opened 2026-05-14
fix(executor): serialize pandas DataFrame/Series and numpy scalars in variable collector

### LINKED ISSUE #192: [Bug]: DataFrame is not JSON serializable crash when code execution produces pandas variables
### What happened

When agent code execution produces a `pandas.DataFrame` or `pd.Series` as a local variable, CUGA crashes with:

`Error: Object of type DataFrame is not JSON serializable`

**Root cause:**

variableUtils.is_serializable() in `src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/common/variable_utils.py `incorrectly returns **True** for `pd.DataFrame `and `pd.Series`:

```
if isinstance(value, (pd.DataFrame, pd.Series)):
    return True 
```  
DataFrames are NOT JSON serializable
These values pass the serialization filter, get stored in AgentState, and crash when LangGraph attempts to checkpoint the state via json.dumps.

**Fix:**
```
if isinstance(value, pd.DataFrame):
    return False  # not JSON serializable; would crash LangGraph state checkpointing
if isinstance(value, pd.Series):
    return False
```

The agent can still use DataFrames freely within a single code block and print results; they simply won't be persisted across turns.

**Note:** numpy.ndarray may have the same issue and is worth checking.

### How to reproduce

Ask the agent to run any Python code that assigns a DataFrame to a variable, e.g.:

import pandas as pd
df = pd.read_csv("some_file.csv")****

### Environment

Mac OS

_No response_

### Checklist

Sergey-Zeltyn:
## Bug fix

Fixes https://github.com/cuga-project/cuga-agent/issues/192

### Summary

`is_serializable()` previously returned `True` for `pd.DataFrame` and `pd.Series`, but these types are not JSON-serializable, causing variables containing pandas objects to be silently dropped or cause errors in the variable collector.

Changes:
- `is_serializable()` now returns `False` for `DataFrame` and `Series`
- New `sanitize_value()` converts `DataFrame` to a dict of records and `Series` to a list before the serializable check
- New `_sanitize_numpy_scalars()` recursively converts numpy scalar types inside DataFrame cells to Python natives, so DataFrames with numpy-typed cells survive the serializable check

### Testing
- [x] Verified fix locally in the example that was created for CRM demo 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sanitization of NumPy, pandas, datetime, bytes, complex, and other non-JSON-friendly types so they are reliably converted to JSON-friendly primitives.
  * pandas DataFrame/Series are now represented as structured, JSON-safe objects rather than treated as directly serializable.
  * Values are now sanitized before serializability checks to prevent incorrect acceptance of unsupported types.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cuga-project/cuga-agent/pull/224)

### DIFF
--- a/src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/common/variable_utils.py
+++ b/src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/common/variable_utils.py
@@ -1,7 +1,7 @@
 import types
 from typing import Any, Optional, Set
-from loguru import logger
 
+from loguru import logger
 
 _TODO_CONFIRMATION_VALUES = frozenset({"Todos updated", "Todos have been updated"})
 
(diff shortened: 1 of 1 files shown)


sami-marreed:
### Critical: numpy scalars nested outside DataFrame `records`

`sanitize_value` only applies `_sanitize_numpy_scalars` to DataFrame `records`. Locals shaped like `[np.int64(1)]` or `{"x": np.float64(1)}` are not normalized at the top level, so `is_serializable` can still fail and those variables may be skipped—the same class of issue as before, just shifted.

**Suggested fix (high level):** run the same recursive numpy/native conversion over arbitrary top-level `dict` / `list` / `tuple` (and optionally `set` → list), not only inside DataFrame `records`.

sami-marreed:
### Critical: non–JSON-safe values inside DataFrame cells

After `to_dict("records")`, cells can still contain types like `pd.Timestamp`, `datetime`, `Timedelta`, `NaT`, `bytes`, nullable/object dtypes, etc. Those may still fail recursive serialization / JSON encoding even though the outer value is no longer a live `DataFrame`.

**Suggested fix (high level):** extend recursive sanitization with explicit mappings for datetime-ish pandas/Python types (stable ISO strings or epoch, plus consistent `NaT` → null), `bytes` (base64 or decode policy), and `complex` (structured object or string), applied in one shared traversal with the numpy scalar handling.

sami-marreed:
The **Unit Tests** workflow is failing on this branch (see [job log](https://github.com/cuga-project/cuga-agent/actions/runs/25852551195/job/75962378637)).

**Cause:** \`ruff format --check\` reports that \`src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/common/variable_utils.py\` would be reformatted (format gate stops before pytest).

**Fix:** Run \`ruff format\` on that file (or \`uv run ruff format ...\` per repo conventions), commit, and push so CI passes.

Sergey-Zeltyn:
@sami-marreed PR comments were addressed in this [commit](https://github.com/cuga-project/cuga-agent/commit/1e2b5788299a02919113e5c0f41b6d87a340e57e)
Ruff formatting was fixed

sami-marreed (approved):

PR #225, opened 2026-05-14
feat: sandbox-aware execution and Carbon side panel refresh

### LINKED ISSUE #227: [Feature]: Surface skills in the chat frontend
## What you want and why

Today, **skills** are primarily discovered and managed outside the main chat flow (e.g. configuration / agent setup). Users working in **chat** lack a first-class way to see **which skills are available**, **attach or enable a skill for this conversation**, or **confirm skill content loaded** without leaving the chat surface.

We want **skills surfaced in the chat frontend** so operators and end users can use skills **in context** of the current thread, reduce round-trips to manage screens, and align chat UX with how tools and knowledge are already exposed in the product.

## How it could work

- **Discovery**: A chat-adjacent panel, drawer, or composer affordance listing skills available to the current agent (respecting tenant/agent config and feature flags).
- **Actions**: Select a skill to **load** or **pin** for the session (mapping to existing backend semantics such as `load_skill` / skill registry—exact API to match current server contracts).
- **Feedback**: Clear confirmation when a skill is active (name, version or path summary, optional collapse/expand for skill instructions preview).
- **Accessibility & layout**: Fits existing Carbon/side-panel patterns where applicable; mobile-friendly collapse behavior.

Open questions for design: permissions (who may attach skills mid-chat), whether attachment is **session-only** vs persisted on agent config, and how this interacts with **policies** / approvals.

## Links or extra context

- Backend: skills discovery under `.cuga` / registry and cuga-lite **`load_skill`** execution path.
- Frontend: reuse patterns from knowledge attach / tool discovery UI where possible.

sami-marreed:
### Related issues

- Closes #227 — Skills now surface in the chat frontend as a dedicated section in the right-side Carbon panel (icon-tab + scrollable list of skill tiles with expandable descriptions and requirements tags).
- Addresses #226 — sandbox/file-path consistency between Python execution and bash commands in `native`/`local` mode. This PR introduces the `sandbox_workspace` plumbing and the workspace-aware prompt rendering. The host-process `chdir` portion is still being evaluated against CI evidence (MCP filesystem allowed-dirs interaction) and may be tightened in a follow-up before final close-out.

### Summary

Two coherent themes land together in this PR:

**1. Sandbox-aware execution**

The executor now matches its working directory and prompt guidance to the active `sandbox_workspace`, eliminating the recurring "is the CWD `.` or `/workspace`?" confusion.

- `code_executor` chdirs Python execution into the per-thread workspace for local/native sandbox modes so generated code shares a CWD with `run_command`.
- `code_wrapper` threads `workspace_root` through `wrap_code` so the prepended `chdir` lives inside the wrapped `async main`.
- `cuga_lite_graph` passes `sandbox_workspace` into the reflection prompt context so strategic guidance matches the executor's actual CWD.
- `reflection_system`, `reflection_user`, and `mcp_prompt` Jinja templates replace "explain every case" guidance with a single branch that renders concrete path rules for the actual `sandbox_workspace` value (`.`, `/workspace`, or other). One workspace → one set of rules.

**2. Carbon side panel refresh**

`ChatLanding`'s right-side agent panel now uses idiomatic Carbon primitives:

- `Accordion` + `AccordionItem` for the app configuration section, with `ContainedList` for the per-app tool lists (replaces a custom expand/collapse implementation).
- `Layer` wrapping the scrollable content area for proper Carbon elevation.
- Non-clickable `Tile` for skill cards with truncated descriptions and a Show more / Show less expander (descriptions over 100 chars).
- Inline-styled underline indicator on the section tabs that reliably tracks the active section without depending on `border-bottom` or pseudo-elements.

Plus an incidental fix in `customSendMessage.ts`: the `thinkingInterval` variable was scoped inside `try`, but cleared inside `finally` — now lifted to function scope. The half-finished thinking-indicator feature has been removed and will return in a follow-up PR with the streaming-display work.

Build / infra: postinstall patch for express's `path-to-regexp` v8 import shape, pinned webpack versions, refreshed `pnpm-lock.yaml`, regenerated `dist/` bundle, and `.claude/settings.local.json` added to `.gitignore`.

### Testing
- [x] Tested locally; UI changes verified in `pnpm dev:real`
- [x] Type checks pass on the modified frontend files
- [x] Pre-commit hooks (ruff, ruff format, conventional commits, detect-secrets) all pass

## Summary by CodeRabbit

* **New Features**
  * Per-thread workspace isolation for secure multi-user agent execution
  * New Skills panel in the agent UI with expandable skill details

* **UI**
  * Agent panel navigation restyled to a 4-column, glassmorphism look
  * Tools list reworked into an accordion layout

* **Bug Fixes**
  * Fixed frontend Express compatibility during install/build

* **Infrastructure**
  * Standardized workspace handling (/workspace → cuga_workspace) and clearer sandbox workspace guidance in prompts and tooling

[Review Change Stack](https://app.coderabbit.ai/change-stack/cuga-project/cuga-agent/pull/225)

# feat/carbonize-chat-side-panel — Filesystem Consolidation

**57 files · +2,276 / −1,669 lines · 8 commits**

## What changed and why

Four separate filesystem implementations (MCP subprocess, per-thread wrapper, and per-executor duplicate tools in `local`, `native`, `opensandbox`) were replaced with a single `WorkspaceFilesystem` class backed by two thin adapters. File I/O is now an independent feature flag — no longer coupled to shell execution.

## New package: `executors/filesystem/`

| File | Role |
|---|---|
| `__init__.py` | Public API: `create_filesystem_tools()`, `download_file()`, `upload_file()` |
| `workspace_fs.py` | `WorkspaceFilesystem` class + `as_structured_tools()` factory |
| `backends.py` | `FilesystemBackend` protocol · `HostWorkspaceBackend` · `RemoteSandboxBackend` |
| `paths.py` | `resolve_workspace_path()` with traversal guard |
| `models.py` | Pydantic models (`FileEntry`, `ReadFileInput`, …) |
| `tests/test_workspace_fs.py` | 225 lines — isolation, traversal rejection, round-trip, edit, move, download/upload |

## 8 LLM tools (bare names, no `filesystem_` prefix)

`read_file` · `write_file` · `edit_file` · `list_files` · `make_directory` · `move_file` · `search_files` · `get_file_info`

`download_file` / `upload_file` are Python-only — not exposed to the LLM.

## Deleted

- `demo_tools/file_system/main.py` — 635-line Python MCP demo server (saves ~2 s startup + one open port)
- `nodes/chat/chat_agent/mcp_filesystem_wrapper.py` — per-thread path rewrite shim
- `@modelcontextprotocol/server-filesystem` npm registration

## Feature flag: `enable_filesystem_tools`

Follows the same propagation path as `enable_todos` / `reflection`:  
`ManagePage UI → feature_flags (DB) → manage_routes.py → DynamicAgentGraph → AgentLoop → cuga_lite_graph.py → create_filesystem_tools()`

Shell and filesystem are now independent gates; enabling file I/O no longer requires shell execution.

## Other highlights

- **Workspace layout** — all workspaces under `<cwd>/cuga_workspace/<thread_id>/`; LLM always sees virtual `/workspace` root
- **Carbon chat side panel** — `ChatLanding.tsx` refreshed with sandbox-awareness and live file listing
- **WatsonX.ai** — improved credential handling and model selection (independent improvement)

## Impact

| | |
|---|---|
| Implementations removed | 4 |
| MCP subprocesses eliminated | 1 |
| Unified runtime class | 1 |
| Backend adapters | 2 |
| LLM tools | 8 |
| Independent feature flag | 1 |
| New test lines | 225 |


src/cuga/backend/cuga_graph/nodes/cuga_lite/cuga_lite_graph.py
```diff
@@ -2218,6 +2218,11 @@ async def sandbox(state: CugaLiteState, config: Optional[RunnableConfig] = None)
                             settings.agent.planner.model
                         )
                         reflection_agent = reflection_task(llm=active_model)
+                        # Calculate sandbox_workspace using same logic as prepare_tools_and_apps
+                        _sandbox_mode = getattr(settings.advanced_features, "sandbox_mode", "opensandbox")
```
offerakrabi: **[MED-1] Sandbox workspace logic duplicated in three places — divergence risk**
`src/cuga/backend/cuga_graph/nodes/cuga_lite/cuga_lite_graph.py:2222`
<details><summary>Explanation</summary>

The expression `"." if _sandbox_mode in ("native", "local") else "/workspace"` now appears in at least three places:
1. `prepare_tools_and_apps` (line ~1695) — for the MCP prompt
2. `sandbox()` (line ~2222) — for the reflection prompt (this PR)
3. `code_executor.py` (line ~151) — for the chdir logic (this PR)

Each site independently reads `settings.advanced_features.sandbox_mode` with `getattr(..., "opensandbox")`. If a new sandbox mode is added (e.g., `"docker"`), all three sites must be updated in sync. The comment in `cuga_lite_graph.py` even says "using same logic as prepare_tools_and_apps", which is a code smell.

</details>
<details><summary>Suggested Fix</summary>

Extract a shared helper, e.g. in a utility module:

```python
# e.g. in executors/common/sandbox_utils.py
def get_sandbox_workspace(sandbox_mode: str) -> str:
    return "." if sandbox_mode in ("native", "local") else "/workspace"

def current_sandbox_workspace() -> str:
    mode = getattr(settings.advanced_features, "sandbox_mode", "opensandbox")
    return get_sandbox_workspace(mode)
```

Then all three call sites use `current_sandbox_workspace()`.

</details>
<details><summary>Agent Prompt</summary>

In `src/cuga/backend/cuga_graph/nodes/cuga_lite/cuga_lite_graph.py` around line 2222, and also in `code_executor.py` around line 151, the expression `"." if _sandbox_mode in ("native", "local") else "/workspace"` is duplicated. A third copy exists in `prepare_tools_and_apps` around line 1695. Extract this into a shared helper function to avoid divergence when new sandbox modes are added.

</details>

src/frontend_workspaces/frontend/scripts/postinstall.cjs
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env node
```
offerakrabi, line 1: **[MED-2] `postinstall.cjs` patches a vendored file in `node_modules` — fragile and non-idiomatic**
`src/frontend_workspaces/frontend/scripts/postinstall.cjs:1`
<details><summary>Explanation</summary>

The postinstall script directly mutates `express/lib/router/layer.js` inside `node_modules` to fix a `path-to-regexp` v8 import shape incompatibility. This approach is fragile:
1. The patch is silently skipped if the file path doesn't match (the script just logs "Pattern not found" or "File not found" and exits 0).
2. `pnpm` may re-hoist or re-link the file on any `pnpm install`, silently reverting the patch.
3. The alternate path (`../../node_modules/.pnpm/express@4.22.2/...`) is hardcoded to a specific express version — it will silently fail if express is upgraded.
4. There is no verification that the patch was actually applied.

The standard approach for this kind of fix is to use `patch-package` (or pnpm's built-in `pnpm patch`) which integrates with the lockfile and is re-applied reliably.

</details>
<details><summary>Suggested Fix</summary>

Use `pnpm patch express` to generate a proper patch file that pnpm applies automatically and tracks in the lockfile. This is the idiomatic pnpm approach and doesn't require a custom postinstall script.

Alternatively, pin `path-to-regexp` to a v7.x version that still exports the old API shape, avoiding the need for the patch entirely.

</details>
<details><summary>Agent Prompt</summary>

In `src/frontend_workspaces/frontend/scripts/postinstall.cjs`, the script patches `express/lib/router/layer.js` directly in `node_modules`. Review whether this is the right approach: the patch silently no-ops if the file path doesn't match, and pnpm may revert it on reinstall. Consider replacing this with `pnpm patch` (which generates a `.patches/` file tracked in the lockfile) or pinning `path-to-regexp` to a compatible version.

</details>

Not sure about this one, it just proposes a different solution

src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/code_executor.py
```diff
@@ -140,7 +140,50 @@ async def eval_with_tools_async(
...
         tracker = ActivityTracker()
         fake_datetime = tracker.current_date if tracker.current_date and is_benchmark_mode() else None
-        wrapped_code = CodeWrapper.wrap_code(code, fake_datetime=fake_datetime)
+
+        # For local/native sandbox modes, inject the workspace root as the working directory
+        # so Python code runs in the same directory context as run_command (which also cds there).
+        # The directory is created here (mirroring LocalSandboxExecutor's behavior for shell flows)
+        # so LocalExecutor paths and unit tests with thread_id=None don't crash on os.chdir.
+        workspace_root_for_wrap = None
+        if mode == 'local':
```
offerakrabi: **[MED-3] `_sandbox_mode == 'native'` branch in `code_executor.py` is unreachable when `mode == 'local'`**
`src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/code_executor.py:153`
<details><summary>Explanation</summary>

The outer condition is `if mode == 'local'`, but inside it checks `if _sandbox_mode == 'native'` and uses `native_thread_workspace_root`. The `mode` parameter is set to `'local'` when `settings.advanced_features.e2b_sandbox` is False (line ~128), regardless of `sandbox_mode`. When `sandbox_mode == 'native'`, the `NativeSandboxExecutor` is used (not `LocalExecutor`), so `mode` would not be `'local'` in that path. This means the `native` branch inside `if mode == 'local'` may never actually be reached in practice, and the `local_thread_workspace_root` fallback is always used. If the intent is to support native mode here, the outer condition needs to be broadened.

</details>
<details><summary>Suggested Fix</summary>

Clarify the intent. If native sandbox mode should also get the chdir treatment, the condition should be:

```python
if mode in ('local',) or _sandbox_mode in ('native', 'local'):
    ...
```

Or document explicitly why native mode is handled inside the `mode == 'local'` branch (e.g., because native mode forces `mode = 'local'` upstream).

</details>
<details><summary>Agent Prompt</summary>

In `src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/code_executor.py` around line 149, the workspace chdir logic is gated on `mode == 'local'`, but inside it branches on `_sandbox_mode == 'native'` to use `native_thread_workspace_root`. Verify whether `mode` can actually be `'local'` when `sandbox_mode` is `'native'` — trace how `mode` is set upstream. If the native branch is unreachable, either remove it or fix the outer condition.

</details>


Not sure i understand the difference between `mode` and `sandbox_node`

src/frontend_workspaces/frontend/src/ManagePage.tsx
offerakrabi, line 1608: need to update this line to point to `.agents`

offerakrabi (approved):

PR #233, opened 2026-05-17
fix(mcp): sanitize dashed MCP tool names to valid Python identifiers

### LINKED ISSUE #185: bug: dashed MCP tool names break code-generation mode (not a valid Python identifier)
## Summary

When an MCP server exposes a tool whose name contains a dash (`-`), CUGA registers the tool with the dash preserved (e.g. `myserver_my-tool`). Direct calls via the registry HTTP API and via `ToolRegistryProvider` work fine, but the registered name is **not a valid Python identifier**, so it breaks code-generation mode: the agent emits Python like `myserver_my-tool(...)` and the compile step raises `SyntaxError`.

A `sanitize_tool_name()` helper already exists at `src/cuga/backend/tools_env/registry/mcp_manager/adapter.py:27-32` and is applied on the OpenAPI path, but **not** on the MCP-server registration path at `src/cuga/backend/tools_env/registry/mcp_manager/mcp_manager.py:760`, where the tool is registered as `f"{name}_{tool.name}"` without sanitizing `tool.name`.

The user-side workaround today is to reconfigure the upstream MCP gateway (e.g. IBM ContextForge) so it never emits dashes, but CUGA should normalize tool names itself.

## Reproduction

A self-contained repro lives under `scripts/dash_repro/`:

- `dash_mcp_server.py` — tiny FastMCP server with two tools:
  - `echo-with-dash` (dash in the name)
  - `echo_with_underscore` (underscore only)
- `mcp_servers.yaml` — registry config that launches the MCP server over stdio
- `run_repro.py` — boots the registry with `MCP_SERVERS_FILE=...`, lists what gets registered, then invokes both tools through:
  1. the registry's HTTP `/functions/call` endpoint
  2. `cuga.backend.cuga_graph.nodes.cuga_lite.tool_registry_provider.ToolRegistryProvider` (the same provider the SDK/agent uses)
  3. a Python `compile()` check, which is what code-generation mode performs

Run:

```bash
uv run --no-sync python scripts/dash_repro/run_repro.py
```

## Expected behavior

- Both tools register under names that are valid Python identifiers (e.g. `dashrepro_echo_with_dash` and `dashrepro_echo_with_underscore`).
- `name.isidentifier()` returns `True` for both.
- A code-gen snippet like `result = dashrepro_echo_with_dash(message='hi')` compiles successfully.
- Direct registry/SDK invocations continue to succeed (the underlying MCP `call_tool` should still receive the original name with the dash).

## Observed behavior (actual run output)

```
[runner] /applications: [{"name": "dashrepro", ...}]

[runner] /applications/dashrepro/apis: {
  "dashrepro_echo-with-dash":      { "api_name": "dashrepro_echo-with-dash", "path": "/echo-with-dash", ... },
  "dashrepro_echo_with_underscore":{ "api_name": "dashrepro_echo_with_underscore", "path": "/echo_with_underscore", ... }
}

[http] tool keys for 'dashrepro':
  ['dashrepro_echo-with-dash', 'dashrepro_echo_with_underscore']

[http] POST /functions/call function_name='dashrepro_echo-with-dash'
  result: {"result": "[echo-with-dash] hello from dashed via http"}
[http] POST /functions/call function_name='dashrepro_echo_with_underscore'
  result: {"result": "[echo_with_underscore] hello from underscored via http"}

[sdk] tools loaded for app 'dashrepro':
  ['dashrepro_echo-with-dash', 'dashrepro_echo_with_underscore']
[sdk] resolved to tool name: 'dashrepro_echo-with-dash'
[sdk] result: {'result': '[echo-with-dash] hello from dashed tool'}
[sdk] resolved to tool name: 'dashrepro_echo_with_underscore'
[sdk] result: {'result': '[echo_with_underscore] hello from underscored tool'}

[sdk] === Python identifier check ===
[sdk]   'dashrepro_echo-with-dash':       isidentifier()=False
[sdk]   'dashrepro_echo_with_underscore': isidentifier()=True

[sdk] === Simulating code-gen call (dashed tool) ===
[sdk]   generated code: "result = dashrepro_echo-with-dash(message='hi')"
[sdk]   compile: SyntaxError: invalid syntax (<gen>, line 1)
[sdk]   generated code: "result = dashrepro_echo_with_underscore(message='hi')"
[sdk]   compile: OK
```

### Diff vs. expected

| Check | Expected | Observed |
| --- | --- | --- |
| Registered name | `dashrepro_echo_with_dash` | `dashrepro_echo-with-dash` |
| `name.isidentifier()` for dashed tool | `True` | **`False`** |
| `compile("dashrepro_echo-with-dash(...)")` | OK | **`SyntaxError`** |
| HTTP `/functions/call` | OK | OK |
| `ToolRegistryProvider` `ainvoke` | OK | OK |

So invocation works, but **code-generation mode is broken** for any MCP tool whose name (or app prefix path) contains a dash.

## Suggested fix

1. In `mcp_manager.py` around line 760, run `sanitize_tool_name(tool.name)` when building `prefixed_name` and the tool dict, while keeping the original name in a sanitized→original map so the eventual MCP `call_tool(...)` still receives the original.
2. Apply the same sanitation when building `path` and the keys of `tools_by_server` / `server_by_tool`.
3. Add a regression test that registers a tool with a dash and asserts `name.isidentifier()` is `True`.

## Workaround

Reconfigure the upstream MCP gateway (e.g. IBM ContextForge) to use underscores for tool names and to not add dashed server prefixes.

## Side note

While building the repro I also noticed that `ToolRegistryProvider` passes an `async def tool_func` to `StructuredTool.from_function(func=...)`. LangChain treats it as a sync function, so `await tool.ainvoke({...})` returns a coroutine instead of awaiting it (the repro unwraps it manually). Probably worth a separate fix: pass it as `coroutine=tool_func`. Not the subject of this bug, just adjacent.

offerakrabi:
## Bug fix

Fixes #185

### Summary

MCP tool names containing dashes (e.g. `echo-with-dash`) were registered as `myserver_echo-with-dash`, which is not a valid Python identifier. Code-generation mode emitted `result = myserver_echo-with-dash(...)` and the compile step raised `SyntaxError`.

**Root cause:** `mcp_manager.py` built `prefixed_name = f"{name}_{tool.name}"` using the raw `tool.name`. `sanitize_tool_name()` already existed in `adapter.py` but was never called on this registration path.

**Why a reverse map is needed:** The call path in `_call_mcp_server_tool` must forward the *original* dashed name to the MCP server (which only knows that name). Simply sanitizing at registration would fix code-gen but break invocations. A `sanitized_prefixed_name → original_tool_name` reverse map is stored so the correct name is recovered at call time.

**Changes:**
- Import `sanitize_tool_name` in `mcp_manager.py`
- Add `self.original_tool_name_by_sanitized: Dict[str, str]` in `MCPManager.__init__`
- Apply `sanitize_tool_name(tool.name)` when building `prefixed_name` at registration; populate the reverse map
- Use reverse map lookup (with `removeprefix` fallback for backward compat) in both branches of `_call_mcp_server_tool`
- Clean up reverse map entries in `_clear_mcp_server_registration`
- Add 15 regression tests in `tests/test_dashed_tool_names.py`

> **Note for reviewers (@haroldship):** The issue references a reproduction script at `scripts/dash_repro/run_repro.py` that was not present in the repository. The full live repro (with a running FastMCP server) from the issue cannot be run without the original `dash_mcp_server.py` and `mcp_servers.yaml` — please run those manually if you have them, or confirm the fix is sufficient from the code and tests.

### Testing
- [x] 15 new regression tests — all pass (`uv run pytest src/cuga/backend/tools_env/registry/mcp_manager/tests/`)
- [x] Full `mcp_manager/tests/` suite: 34/34 pass, zero regressions
- [x] End-to-end verification script `scripts/dash_repro/verify_fix.py` — all 10 checks pass
- [x] `uv run ruff format` and `uv run ruff check --fix` — clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Tools with dashed or otherwise unsafe names are now registered using sanitized identifiers while preserving their original names for invocation; collisions on the same server are detected so duplicate sanitized names are skipped, and cleanup removes preserved mappings.

* **Tests**
  * Expanded coverage for sanitization across formats, collision handling, registration/call behavior, and cleanup.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cuga-project/cuga-agent/pull/233?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/cuga/backend/tools_env/registry/mcp_manager/mcp_manager.py
+++ b/src/cuga/backend/tools_env/registry/mcp_manager/mcp_manager.py
@@ -1079,8 +1079,9 @@ class MCPManager:
                 # Use the reverse map to recover the original (possibly dashed) tool name
                 # that the MCP server registered under. Fall back to removeprefix for
                 # tools registered before this map existed (backward compatibility).
-                original_tool_name = self.original_tool_name_by_sanitized.get(tool_name) \
-                    or tool_name.removeprefix(f"{server_name}_")
+                original_tool_name = self.original_tool_name_by_sanitized.get(
+                    tool_name
+                ) or tool_name.removeprefix(f"{server_name}_")
 
                 transport = self.mcp_transports[server_name]
                 client = FastMCPClient(transport)
@@ -1102,8 +1103,9 @@ class MCPManager:
             else:
                 url = self.mcp_clients[server_name]
                 base_url = url.replace('/sse', '')
-                original_tool_name = self.original_tool_name_by_sanitized.get(tool_name) \
-                    or tool_name.removeprefix(f"{server_name}_")
+                original_tool_name = self.original_tool_name_by_sanitized.get(
+                    tool_name
+                ) or tool_name.removeprefix(f"{server_name}_")
 
                 # Add query params to URL if present
                 url_with_params = base_url
--- a/src/cuga/backend/tools_env/registry/mcp_manager/mcp_manager.py
+++ b/src/cuga/backend/tools_env/registry/mcp_manager/mcp_manager.py
@@ -25,6 +25,7 @@ from cuga.backend.tools_env.registry.mcp_manager.openapi_parser import SimpleOpe
 from cuga.backend.tools_env.registry.mcp_manager.adapter import (
     new_mcp_from_custom_parser,
     apply_authentication,
+    sanitize_tool_name,
 )
 import threading
 from collections import defaultdict
@@ -43,6 +44,7 @@ class MCPManager:
         self.threads = {}
         self.tools_by_server = defaultdict(list)
         self.server_by_tool = {}
+        self.original_tool_name_by_sanitized: Dict[str, str] = {}
         self.server_ports = {}
         self.auth_config = {}
         self.schemas = {}
@@ -669,6 +671,7 @@ class MCPManager:
         stale_tools = [tool_name for tool_name, server in self.server_by_tool.items() if server == name]
         for tool_name in stale_tools:
             self.server_by_tool.pop(tool_name, None)
+            self.original_tool_name_by_sanitized.pop(tool_name, None)
 
     @staticmethod
     def _extract_json_path(payload: Any, path: str | None) -> Any:
@@ -760,7 +763,11 @@ class MCPManager:
                 if include_set and tool.name not in include_set:
                     continue
 
-                prefixed_name = f"{name}_{tool.name}"
+                sanitized_name = sanitize_tool_name(tool.name)
+                prefixed_name = f"{name}_{sanitized_name}"
+                # Keep a reverse map so _call_mcp_server_tool can send the original
+                # (possibly dashed) name to the MCP server, which only knows that name.
+                self.original_tool_name_by_sanitized[prefixed_name] = tool.name
                 input_schema = tool.inputSchema if hasattr(tool, 'inputSchema') else {}
                 flattened_params = self._flatten_tool_parameters(input_schema)
                 output_schema = tool.outputSchema if hasattr(tool, 'outputSchema') else {}
@@ -1069,7 +1076,11 @@ class MCPManager:
                 apply_authentication(auth, headers, query_params)
 
             if hasattr(self, 'mcp_transports') and server_name in self.mcp_transports:
-                original_tool_name = tool_name.removeprefix(f"{server_name}_")
+                # Use the reverse map to recover the original (possibly dashed) tool name
+                # that the MCP server registered under. Fall back to removeprefix for
+                # tools registered before this map existed (backward compatibility).
+                original_tool_name = self.original_tool_name_by_sanitized.get(tool_name) \
+                    or tool_name.removeprefix(f"{server_name}_")
 
                 transport = self.mcp_transports[server_name]
                 client = FastMCPClient(transport)
@@ -1091,7 +1102,8 @@ class MCPManager:
             else:
                 url = self.mcp_clients[server_name]
                 base_url = url.replace('/sse', '')
-                original_tool_name = tool_name.removeprefix(f"{server_name}_")
+                original_tool_name = self.original_tool_name_by_sanitized.get(tool_name) \
+                    or tool_name.removeprefix(f"{server_name}_")
 
                 # Add query params to URL if present
                 url_with_params = base_url
--- a/src/cuga/backend/tools_env/registry/mcp_manager/mcp_manager.py
+++ b/src/cuga/backend/tools_env/registry/mcp_manager/mcp_manager.py
@@ -765,6 +765,17 @@ class MCPManager:
 
                 sanitized_name = sanitize_tool_name(tool.name)
                 prefixed_name = f"{name}_{sanitized_name}"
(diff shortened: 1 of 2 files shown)


src/cuga/backend/tools_env/registry/mcp_manager/mcp_manager.py
```diff
@@ -760,7 +763,22 @@ async def fetch_tools():
                 if include_set and tool.name not in include_set:
                     continue
 
-                prefixed_name = f"{name}_{tool.name}"
+                sanitized_name = sanitize_tool_name(tool.name)
```
sami-marreed, line 766: can't this implicit tool sanitization be dangerous, as user doesn't know how tool name changed

offerakrabi: it could be confusing for someone trying to debug (maybe we can add logging when sanitizing the name), but the alternative is worse..

if no sanitizing a tool name like `test_tool-name` will raise a `SyntaxError` during code generation

our other option is just to not allow names like this in the first place, it's a clean solution (i like it) but it means we can't use IBM ContextForge

src/cuga/backend/tools_env/registry/mcp_manager/mcp_manager.py
```diff
@@ -1069,7 +1087,12 @@ async def _call_mcp_server_tool(self, server_name: str, tool_name: str, args: di
                 apply_authentication(auth, headers, query_params)
 
             if hasattr(self, 'mcp_transports') and server_name in self.mcp_transports:
-                original_tool_name = tool_name.removeprefix(f"{server_name}_")
+                # Use the reverse map to recover the original (possibly dashed) tool name
+                # that the MCP server registered under. Fall back to removeprefix for
+                # tools registered before this map existed (backward compatibility).
+                original_tool_name = self.original_tool_name_by_sanitized.get(
```
haroldship: don't you add the toolname into the dict on line 781 above? how can it get here and not have the tool name in the dict? So not sure I get how it can ever use the removeprefix() fallback

src/cuga/backend/tools_env/registry/mcp_manager/mcp_manager.py
```diff
@@ -1091,7 +1114,9 @@ async def _call_mcp_server_tool(self, server_name: str, tool_name: str, args: di
             else:
                 url = self.mcp_clients[server_name]
                 base_url = url.replace('/sse', '')
-                original_tool_name = tool_name.removeprefix(f"{server_name}_")
+                original_tool_name = self.original_tool_name_by_sanitized.get(
```
haroldship: don't you add the toolname into the dict on line 781 above? how can it get here and not have the tool name in the dict? So not sure I get how it can ever use the removeprefix() fallback

haroldship (approved):

src/cuga/backend/tools_env/registry/mcp_manager/mcp_manager.py
offerakrabi: you're right, its not necessary, i removed it.

PR #236, opened 2026-05-18
fix: Instrument CugaAgent SDK with TokenUsageTracker for trajectory support (#71)

### LINKED ISSUE #71: Instrument CugaAgent SDK with TokenUsageTracker for cuga-viz trajectory support
## Problem

`CugaAgent` (SDK mode) does not instrument prompt/step tracking, so trajectory files produced via the SDK have empty `prompts` fields on most steps. This means tools like cuga-viz cannot display the full LLM conversation for SDK-invoked tasks.

### Root cause

`AgentRunner` (the full agent-loop path) wires up `TokenUsageTracker` — a LangChain `AsyncCallbackHandler` in `cuga/backend/cuga_graph/utils/agent_loop.py` — which captures:
- `on_llm_start()` → system/human prompts
- `on_llm_end()` → assistant generation

The graph nodes (TaskAnalyzer, TaskDecomposition, CodeActAgent, FinalAnswer) also call `tracker.collect_step()` directly.

`CugaAgent` in `cuga/sdk.py` does **none of this**. It calls the underlying LLM but doesn't register `TokenUsageTracker` as a callback, so `ActivityTracker` never sees the prompts or steps.

### Impact

| Field | CugaAgent (SDK) | AgentRunner |
|---|---|---|
| `steps[].prompts` | Empty (except manually added) | Full system/human/assistant per LLM call |
| Step granularity | Minimal | Full node-level breakdown |
| cuga-viz support | Incomplete trajectories | Full trajectories |

### Proposal

Wire `TokenUsageTracker` into `CugaAgent`'s LLM invocation path so that SDK users get the same trajectory instrumentation as `AgentRunner` users, without needing to switch execution modes.

Specifically:
1. Register `TokenUsageTracker(tracker)` as a LangChain callback in `CugaAgent`'s chat/invoke methods
2. Ensure `collect_step()` is called at appropriate points (at minimum: user prompt, assistant response, final answer)
3. This enables cuga-viz to render full trajectories for all CUGA invocation modes

### Files involved

- `src/cuga/sdk.py` — `CugaAgent` class, needs callback registration
- `src/cuga/backend/cuga_graph/utils/agent_loop.py` — `TokenUsageTracker` (already exists, just needs to be reused)
- `src/cuga/backend/activity_tracker/tracker.py` — `ActivityTracker`, `collect_step()`, `collect_prompt()`

### Context

Discovered while investigating trajectory inconsistencies across benchmarks in [cuga-internal-evaluation#37](https://github.ibm.com/research-rpa/cuga-internal-evaluation/issues/37).

haroldship:
## Problem
CugaAgent SDK did not track LLM prompts and responses, resulting in trajectory 
files with empty `prompts` fields. This prevented tools like cuga-viz from 
displaying the full conversation history for SDK-invoked tasks.

## Solution
- Added TokenUsageTracker integration to CugaAgent SDK
- Created `_build_callbacks()` helper method that automatically includes 
  TokenUsageTracker alongside user-provided callbacks
- Updated both `invoke()` and `stream()` methods to use the new callback builder
- Callbacks are registered in both top-level config and configurable for node access

## Changes
- **src/cuga/sdk.py**:
  - Added imports for TokenUsageTracker and ActivityTracker
  - Added `_build_callbacks()` method to automatically include TokenUsageTracker
  - Updated `invoke()` method to use `_build_callbacks()` (2 locations: resume and normal)
  - Updated `stream()` method to use `_build_callbacks()` (2 locations: resume and normal)

- **src/cuga/sdk_core/tests/test_sdk_token_usage_tracker.py** (new):
  - 6 comprehensive unit tests for TokenUsageTracker integration
  - Tests verify callback registration, user callback preservation, and ActivityTracker integration

## Testing
✅ All 6 new unit tests pass (test_sdk_token_usage_tracker.py)
✅ Ruff checks pass on modified files
✅ TokenUsageTracker automatically included in all SDK invocations
✅ User-provided callbacks are preserved and extended

## Benefits
- Full trajectory support for SDK users
- cuga-viz compatibility for SDK-invoked tasks
- Consistent instrumentation across all execution modes (SDK and AgentRunner)
- No breaking changes to existing API
- Automatic - no user action required

Fixes #71

## Summary by CodeRabbit

* **New Features**
  * Token usage and activity tracking are now applied automatically across all agent execution paths, including streaming and resume flows.
  * User-supplied callbacks are preserved and executed alongside built-in trackers.

* **Bug Fixes**
  * Prevents caller-provided configuration from being mutated across repeated calls.

* **Tests**
  * Added tests ensuring trackers are always included, user callbacks are retained, prompts are collected, config immutability, and direct graph usage is instrumented.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cuga-project/cuga-agent/pull/236?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/cuga/sdk.py
+++ b/src/cuga/sdk.py
@@ -1792,9 +1792,22 @@ class CugaAgent:
             self._inject_knowledge_to_config(run_config)
 
             # Add callbacks to config (including TokenUsageTracker for trajectory tracking)
-            callbacks = self._build_callbacks()
-            run_config["callbacks"] = callbacks
-            run_config["configurable"]["callbacks"] = callbacks
+            # Merge with any existing per-call callbacks
+            built_callbacks = self._build_callbacks()
+            existing_callbacks = run_config.get("callbacks", [])
+            existing_configurable_callbacks = run_config.get("configurable", {}).get("callbacks", [])
+
+            # Ensure existing callbacks are lists
+            if not isinstance(existing_callbacks, list):
+                existing_callbacks = [existing_callbacks] if existing_callbacks else []
+            if not isinstance(existing_configurable_callbacks, list):
+                existing_configurable_callbacks = (
+                    [existing_configurable_callbacks] if existing_configurable_callbacks else []
+                )
+
+            # Merge: built callbacks first, then existing per-call callbacks
+            run_config["callbacks"] = built_callbacks + existing_callbacks
+            run_config["configurable"]["callbacks"] = built_callbacks + existing_configurable_callbacks
 
             # If action_response provided, update state with it
             if action_response:
@@ -1926,9 +1939,22 @@ class CugaAgent:
             run_config["configurable"]["policy_system"] = self._policy_system
 
         # Add callbacks to config (including TokenUsageTracker for trajectory tracking)
-        callbacks = self._build_callbacks()
-        run_config["callbacks"] = callbacks
-        run_config["configurable"]["callbacks"] = callbacks
+        # Merge with any existing per-call callbacks
+        built_callbacks = self._build_callbacks()
+        existing_callbacks = run_config.get("callbacks", [])
+        existing_configurable_callbacks = run_config.get("configurable", {}).get("callbacks", [])
+
+        # Ensure existing callbacks are lists
+        if not isinstance(existing_callbacks, list):
+            existing_callbacks = [existing_callbacks] if existing_callbacks else []
+        if not isinstance(existing_configurable_callbacks, list):
+            existing_configurable_callbacks = (
+                [existing_configurable_callbacks] if existing_configurable_callbacks else []
+            )
+
+        # Merge: built callbacks first, then existing per-call callbacks
+        run_config["callbacks"] = built_callbacks + existing_callbacks
+        run_config["configurable"]["callbacks"] = built_callbacks + existing_configurable_callbacks
 
         # Add knowledge engine for awareness injection
         self._inject_knowledge_to_config(run_config)
@@ -2070,9 +2096,22 @@ class CugaAgent:
                 run_config["configurable"]["policy_system"] = self._policy_system
 
             # Add callbacks to config (including TokenUsageTracker for trajectory tracking)
-            callbacks = self._build_callbacks()
-            run_config["callbacks"] = callbacks
-            run_config["configurable"]["callbacks"] = callbacks
+            # Merge with any existing per-call callbacks
+            built_callbacks = self._build_callbacks()
+            existing_callbacks = run_config.get("callbacks", [])
+            existing_configurable_callbacks = run_config.get("configurable", {}).get("callbacks", [])
+
+            # Ensure existing callbacks are lists
+            if not isinstance(existing_callbacks, list):
+                existing_callbacks = [existing_callbacks] if existing_callbacks else []
+            if not isinstance(existing_configurable_callbacks, list):
+                existing_configurable_callbacks = (
+                    [existing_configurable_callbacks] if existing_configurable_callbacks else []
+                )
+
+            # Merge: built callbacks first, then existing per-call callbacks
+            run_config["callbacks"] = built_callbacks + existing_callbacks
+            run_config["configurable"]["callbacks"] = built_callbacks + existing_configurable_callbacks
 
             # Add knowledge engine for awareness injection
             self._inject_knowledge_to_config(run_config)
@@ -2122,9 +2161,22 @@ class CugaAgent:
             run_config["configurable"]["policy_system"] = self._policy_system
 
         # Add callbacks to config (including TokenUsageTracker for trajectory tracking)
-        callbacks = self._build_callbacks()
-        run_config["callbacks"] = callbacks
-        run_config["configurable"]["callbacks"] = callbacks
+        # Merge with any existing per-call callbacks
+        built_callbacks = self._build_callbacks()
+        existing_callbacks = run_config.get("callbacks", [])
+        existing_configurable_callbacks = run_config.get("configurable", {}).get("callbacks", [])
+
+        # Ensure existing callbacks are lists
+        if not isinstance(existing_callbacks, list):
+            existing_callbacks = [existing_callbacks] if existing_callbacks else []
+        if not isinstance(existing_configurable_callbacks, list):
+            existing_configurable_callbacks = (
+                [existing_configurable_callbacks] if existing_configurable_callbacks else []
+            )
+
+        # Merge: built callbacks first, then existing per-call callbacks
+        run_config["callbacks"] = built_callbacks + existing_callbacks
+        run_config["configurable"]["callbacks"] = built_callbacks + existing_configurable_callbacks
 
         # Add knowledge engine for awareness injection
         self._inject_knowledge_to_config(run_config)
--- a/src/cuga/sdk.py
+++ b/src/cuga/sdk.py
@@ -1365,6 +1365,36 @@ class CugaAgent:
 
         return callbacks
 
+    def _prepare_run_config(self, config: Optional[RunnableConfig]) -> dict:
+        """
+        Shallow-copy caller's config so per-call mutations don't accumulate on reuse.
+
+        Without this, callers that reuse a config dict across invoke()/stream() calls
+        would see TokenUsageTracker and any other injected entries pile up on each call.
+        """
+        run_config: dict = dict(config) if config else {}
+        run_config["configurable"] = dict(run_config.get("configurable") or {})
+        return run_config
+
+    def _apply_callbacks(self, run_config: dict) -> None:
+        """
+        Merge built-in callbacks (TokenUsageTracker + user callbacks) with any
+        caller-supplied callbacks in run_config, writing the result to both the
+        top-level and ``configurable`` slots.
+        """
+        built_callbacks = self._build_callbacks()
+
+        def _as_list(value: Any) -> list:
+            if value is None:
+                return []
+            return list(value) if isinstance(value, list) else [value]
+
+        existing = _as_list(run_config.get("callbacks"))
+        existing_configurable = _as_list(run_config["configurable"].get("callbacks"))
+
+        run_config["callbacks"] = built_callbacks + existing
+        run_config["configurable"]["callbacks"] = built_callbacks + existing_configurable
+
     async def _ensure_initialized(self):
         """Ensure tool provider is initialized."""
         if not hasattr(self.tool_provider, 'initialized') or not self.tool_provider.initialized:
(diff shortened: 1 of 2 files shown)


haroldship:
✅ **Fixed: Callback merging implemented**

Updated all 4 callback registration locations to merge per-call callbacks with TokenUsageTracker instead of overwriting:

**Changes:**
- Extract existing callbacks from `run_config` (if any)
- Ensure they are lists (handle single callback case)  
- Merge: `built_callbacks` (with TokenUsageTracker) + existing per-call callbacks
- Apply merged list to both `run_config["callbacks"]` and `run_config["configurable"]["callbacks"]`

**Locations updated:**
1. `invoke()` resume path (~line 1794)
2. `invoke()` normal path (~line 1928)
3. `stream()` resume path (~line 2072)
4. `stream()` normal path (~line 2124)

**Validation:**
✅ Ruff check passed
✅ Ruff format passed

This ensures user-provided callbacks are preserved alongside TokenUsageTracker, addressing the review concern about overwriting caller-supplied callbacks.

haroldship:
## ✅ Callback Merging Test Results

**Test demonstrates the fix works correctly!**

### Test Evidence
Ran `test_callback_merge.py` which passes a custom callback via `config={"callbacks": [custom_callback]}`:

```
✓ CustomCallback callback was called (LLM start #1)
```

**This proves:**
- ✅ Custom callback was invoked (received `on_llm_start` event)
- ✅ Callback was NOT overwritten by TokenUsageTracker
- ✅ Both TokenUsageTracker and user callback coexist
- ✅ Merging logic works as designed

### Test Code
```python
# Create custom callback
custom_callback = TestCallback("CustomCallback")

# Pass via config parameter
result = await agent.invoke(
    "What is 2+2?",
    config={"callbacks": [custom_callback]}
)

# Verify it was called
assert custom_callback.called  # ✅ PASSED
```

### Before vs After

**Before (would overwrite):**
```python
run_config["callbacks"] = callbacks  # ❌ User callback lost
```

**After (merges):**
```python
built_callbacks = self._build_callbacks()  # [TokenUsageTracker]
existing_callbacks = run_config.get("callbacks", [])  # [CustomCallback]
run_config["callbacks"] = built_callbacks + existing_callbacks  # ✅ Both present
```

Full test results in [TEST_RESULTS.md](https://github.com/cuga-project/cuga-agent/blob/fix/71-sdk-token-usage-tracker-instrumentation/TEST_RESULTS.md)

src/cuga/sdk.py
```diff
@@ -1767,7 +1791,23 @@ async def invoke(
             # Add knowledge engine for awareness injection
             self._inject_knowledge_to_config(run_config)
 
-            # Add callbacks to config (both top-level and configurable for nodes)
+            # Add callbacks to config (including TokenUsageTracker for trajectory tracking)
+            # Merge with any existing per-call callbacks
+            built_callbacks = self._build_callbacks()
+            existing_callbacks = run_config.get("callbacks", [])
+            existing_configurable_callbacks = run_config.get("configurable", {}).get("callbacks", [])
```
offerakrabi: **[MED-1] Reusing a caller config accumulates duplicate callbacks**
`src/cuga/sdk.py:1765`
<details><summary>Explanation</summary>

The SDK aliases the caller-provided `config` with `run_config = config or {}` and then mutates it in place. Each invocation reads callbacks back from the same dict and prepends a freshly built callback list, so callers that reuse a config object across `invoke()` or `stream()` calls will accumulate repeated `TokenUsageTracker` instances and repeated constructor/per-call callbacks. That can double-count prompts/tokens and fire user callback handlers multiple times for a single LLM call.

The same pattern appears in the resume and normal paths for both `invoke()` and `stream()`.

For example, caller code may reasonably reuse one config object across multiple turns:

```python
config = {
    "configurable": {
        "thread_id": "my-thread"
    }
}

await agent.invoke("Hello", config=config)
await agent.invoke("What did I just say?", config=config)
await agent.invoke("Summarize the conversation", config=config)
```

Because the SDK currently does `run_config = config or {}` and then writes callbacks into `run_config`, it mutates the original `config` dict. After the first call, the caller's config now contains injected callbacks. On the second call, the SDK reads those previously injected callbacks as `existing_callbacks`, builds a new `TokenUsageTracker`, and prepends it again. After repeated calls, the callback list can grow like:

```python
config["callbacks"] = [
    TokenUsageTracker(...),  # third call
    TokenUsageTracker(...),  # second call
    TokenUsageTracker(...),  # first call
]
```

That means one LLM call may trigger multiple trackers, potentially duplicating prompt records and over-counting token usage. The same issue can also affect user-provided callbacks, causing custom telemetry/logging/UI handlers to fire more than once.

</details>
<details><summary>Suggested Fix</summary>

Avoid mutating the caller-owned config in place, or deduplicate injected callbacks before assigning them. A small helper that shallow-copies the top-level dict and `configurable` dict before adding SDK callbacks would prevent repeated injection when the same config is reused.

A simple way to verify whether this actually happens is to add a regression test that reuses the same config object across multiple SDK calls and inspects the config after each call. The graph invocation can be mocked so no real LLM call is needed, then the test can assert that the number of `TokenUsageTracker` callbacks does not increase after repeated invocations:

```python
config = {"configurable": {"thread_id": "test-thread"}}

await agent.invoke("first", config=config)
token_trackers_after_first = [
    cb for cb in config["configurable"]["callbacks"]
    if isinstance(cb, TokenUsageTracker)
]

await agent.invoke("second", config=config)
token_trackers_after_second = [
    cb for cb in config["configurable"]["callbacks"]
    if isinstance(cb, TokenUsageTracker)
]

assert len(token_trackers_after_second) == len(token_trackers_after_first)
```

If the current implementation mutates and reuses the same callback list, this test should reveal the callback count growing between calls. A stronger test would also include a custom user callback and assert that it appears exactly once after repeated invocations.

</details>
<details><summary>Agent Prompt</summary>

In `src/cuga/sdk.py` around line 1765, there may be an issue with mutating and reusing the caller-provided config when adding `TokenUsageTracker` callbacks. Review the current implementation and determine whether repeated calls with the same `config` object duplicate callbacks. Verify the finding against the current code and only fix it if it is indeed a problem. If fixing, ensure per-call callbacks are still preserved, constructor callbacks are still included, and both top-level and `configurable` callback locations remain compatible with LangGraph/CugaLite.

</details>

src/cuga/sdk.py
```diff
@@ -1341,6 +1343,28 @@ async def initialize(self):
...
+
+        Returns:
+            List of callback handlers including TokenUsageTracker and user-provided callbacks
+        """
+        tracker = ActivityTracker()
+        callbacks: List[BaseCallbackHandler] = [TokenUsageTracker(tracker)]
+
+        # Add user-provided callbacks
+        if self._callbacks:
+            callbacks.extend(self._callbacks)
```
offerakrabi, line 1361: **[MED-2] Direct graph SDK usage remains uninstrumented**
`src/cuga/sdk.py:1442`
<details><summary>Explanation</summary>

The graph is created with `callbacks=self._callbacks`, while the new `TokenUsageTracker` is only added later by `CugaAgent.invoke()` and `CugaAgent.stream()` through runtime config. The SDK exposes and documents direct `agent.graph.ainvoke(...)` usage, so that documented SDK path still will not receive `TokenUsageTracker` unless callers manually add it. This leaves a gap in the PR’s stated goal of trajectory support for SDK-invoked tasks.

The important distinction is that there are two SDK entry points. The high-level path goes through `CugaAgent.invoke()` or `CugaAgent.stream()`, and the PR adds callbacks there:

```python
await agent.invoke("Hello")
```

That path runs `_build_callbacks()` and adds `TokenUsageTracker` into the runtime config before executing the graph.

The lower-level path uses the exposed graph directly:

```python
result = await agent.graph.ainvoke(
    {
        "chat_messages": [HumanMessage(content="Find my latest invoice")],
        "thread_id": "abc",
        "input": "Find my latest invoice",
        "url": "",
    },
    config={
        "configurable": {
            "thread_id": "abc"
        }
    },
)
```

That call bypasses `CugaAgent.invoke()` and `CugaAgent.stream()`, so it does not run the new `_build_callbacks()` logic. In `_create_hitl_wrapper_graph()`, the subgraph is constructed with only the user-provided callbacks:

```python
cuga_lite_subgraph = create_cuga_lite_graph(
    model=self._model,
    tool_provider=self.tool_provider,
    thread_id=thread_id,
    callbacks=self._callbacks,
    special_instructions=self._special_instructions,
)
```

Since this uses `callbacks=self._callbacks` instead of a callback list that includes `TokenUsageTracker`, direct graph invocations may still produce trajectories without prompt/response tracking unless the caller manually knows to pass the tracker in config. If direct graph usage is considered a supported SDK API, the PR only fixes one door into the SDK and leaves the other door uninstrumented.

</details>
<details><summary>Suggested Fix</summary>

Either include the built-in tracker in the graph-level callbacks passed to `create_cuga_lite_graph`, or document/structure the SDK API so trajectory support is explicitly limited to `CugaAgent.invoke()` and `CugaAgent.stream()`. If graph-level callbacks are changed, ensure direct graph invocations still preserve user-provided callbacks and do not double-register the tracker for normal `invoke()`/`stream()` calls.

A concrete way to verify this is to test both entry points with a mocked model call and inspect the callbacks passed to CugaLite. The high-level SDK path should include `TokenUsageTracker`:

```python
await agent.invoke("Hello")
```

The direct graph path should be checked separately:

```python
await agent.graph.ainvoke(
    state,
    config={"configurable": {"thread_id": "abc"}},
)
```

If the finding is real, the captured model invocation config will look conceptually like this:

```text
agent.invoke(...)        -> callbacks include TokenUsageTracker
agent.graph.ainvoke(...) -> callbacks do not include TokenUsageTracker
```

The test does not need a real LLM. It only needs to capture the `config={"callbacks": ...}` argument passed into the underlying CugaLite model invocation and assert whether `TokenUsageTracker` is present for each SDK entry point.

</details>
<details><summary>Agent Prompt</summary>

In `src/cuga/sdk.py` around line 1442, there may be an issue with direct `agent.graph.ainvoke()` usage not receiving the new `TokenUsageTracker`. Review the current implementation and determine whether documented direct graph SDK usage bypasses the new callback builder. Verify the finding against the current code and only fix it if it is indeed a problem. If fixing, ensure normal `invoke()` and `stream()` paths do not double-count callbacks while direct graph usage gains trajectory instrumentation or is clearly excluded.

</details>


This is not necessarily a bug/issue, we just need to confirm this is the behavior we want

src/cuga/sdk.py
haroldship: Fixed in dd92f2da (regression test in 09e6b900).

`run_config` is now shallow-copied from the caller's `config` via a new `_prepare_run_config()` helper, and its `configurable` sub-dict is shallow-copied too. The four invoke/stream paths use it, so reassigning `run_config["callbacks"]` / `run_config["configurable"]["callbacks"]` no longer touches the caller's dict — reusing the same config across multiple invoke()/stream() calls no longer accumulates trackers.

Added `test_invoke_does_not_mutate_caller_config` that calls `invoke()` three times with the same config and asserts no new keys appear on the caller's dict.

src/cuga/sdk.py
haroldship: Fixed in dd92f2da — direct `agent.graph.ainvoke(...)` is now also instrumented.

`_create_hitl_wrapper_graph()` now passes `self._build_callbacks()` (TokenUsageTracker + user callbacks) to `create_cuga_lite_graph()` as `base_callbacks`, instead of just `self._callbacks`. Inside the node:

```python
current_callbacks = configurable.get("callbacks", base_callbacks or [])
```

…uses runtime callbacks when present, and falls back to base otherwise. So:

- `agent.invoke()` / `agent.stream()` still set `configurable["callbacks"]` → that wins, no double-counting.
- `agent.graph.ainvoke(...)` with no `configurable["callbacks"]` → falls back to base, which now includes TokenUsageTracker.

`ActivityTracker` is a singleton so the bake-time and runtime trackers share state.

Added `test_direct_graph_access_is_instrumented` that spies on `create_cuga_lite_graph` and asserts `TokenUsageTracker` is among the base callbacks the graph is built with.

offerakrabi (approved):

PR #250, opened 2026-05-20
refactor(agent-core): extract shared agent logic into cuga_agent_core module

sami-marreed:
## Chore

Closes #205

### Summary

Extracts duplicated graph logic from `CugaLiteGraph` and `CugaSupervisorGraph` into a new shared `cuga_agent_core` module, making both agents more modular and easier to contribute to.

**What changed:**

- **New `cuga_agent_core/` module** — shared building blocks extracted from both agents:
  - `graph_nodes.py` / `shared_nodes.py` — common LangGraph node implementations
  - `shared_graph.py` — reusable graph builder utilities
  - `execution_policy.py` — unified execution policy logic
  - `runtime_tools.py` — shared tool resolution and wiring
  - `code_extraction.py` — code block extraction (moved from `code_act_agent`)
  - `variable_bridge.py` — variable pass-through between graph layers

- **`CugaLiteGraph` refactored** — now delegates to `AgentGraphAdapter` (`cuga_lite/agent_graph_adapter.py`); core graph logic slimmed from ~700 → ~150 lines

- **`CugaSupervisorGraph` refactored** — now delegates to `SupervisorGraphAdapter` (`cuga_supervisor/supervisor_graph_adapter.py`); same pattern applied

- **Deleted `code_act_agent.py`** — logic absorbed into `cuga_agent_core`

- **Full test coverage** — 13 new test files covering all extracted modules and both adapters

## Release Notes

* **New Features**
  * Added execution plan configuration with explicit control over Python, shell, and filesystem execution backends.
  * Variables are now exported in invoke results for better cross-agent data sharing.
  * Enhanced tool binding with configurable modes (all, find_tools, apps, specific tools).
  * Added task/todo tracking and management throughout execution.

* **Bug Fixes**
  * Improved tool approval handling in supervisor workflows.
  * Better management of knowledge tool scopes.

* **Tests**
  * Comprehensive test coverage for new execution planning, graph adapters, and tool workflows.

* **Chores**
  * Reorganized tool provider modules into unified package structure.
  * Consolidated code extraction utilities.
  * Improved project documentation and architectural guidance.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cuga-project/cuga-agent/pull/250?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #251, opened 2026-05-20
fix(browser_planner): gate multimodal content on per-model vision support

### LINKED ISSUE #214: [Bug]: Web agent in hybrid mode fails with "messages[1].content must be a string" (400 invalid_request_error)
## What happened

Running the web automation agent against a digital-sales prompt fails immediately with a 400 from the upstream LLM API:

```
Error: Error code: 400 - {'error': {'message': 'messages[1].content must be a string', 'type': 'invalid_request_error', 'param': 'messages[1].content'}}
```

The error indicates that `messages[1].content` is being sent as a non-string value (list/dict) to a model that requires string content. This points at message-format construction inside cuga's hybrid-mode web agent path, not at user input.

## How to reproduce

```bash
DYNACONF_ADVANCED_FEATURES__MODE=hybrid uv run cuga start demo
```

In the demo UI, submit prompt:

```
get top account by revenue from digital sales then add it to current page
```

→ agent returns the 400 error above. Reproduces deterministically.

## Environment

- OS: macOS (Darwin 25.4.0 arm64)
- Python: 3.12.7
- CUGA version: 0.2.28
- Branch: `main` at commit `b5b72b2f608761aa133728b6f53efe9e59fc1d09`
- Mode: `DYNACONF_ADVANCED_FEATURES__MODE=hybrid`
- Preset: `demo` (digital_sales + filesystem MCP, default)
- Key dep versions: playwright 1.59.0, langchain-core 1.3.2, langchain-openai 1.1.10, litellm 1.83.14, browsergym-core 0.13.0

The 400 is from the LLM API call inside the web-agent step. Likely culprits to investigate:

- A message in the web-agent flow is being built with `content=[...]` (multimodal blocks or langchain structured content) and the configured model expects a plain string. This was historically tolerated by some providers and is now rejected.
- Could be model-specific — would be worth checking which `MODEL_NAME` / `OPENAI_BASE_URL` the user has configured. The original failure surfaced against the user's WatsonX-via-openai-compatible setup.

Suggested first steps:
- Capture the full request body sent to the LLM right before the 400 (langchain debug callback or `LITELLM_LOG=DEBUG`) to see exactly which message has list-typed content.
- Check `src/cuga/backend/cuga_graph/nodes/web_agent_*` (or equivalent) for `HumanMessage(content=[...])` / `AIMessage(content=[...])` construction.

## Context

Surfaced during manual web-agent verification for PR #212 (playwright security floor bump). Verified to reproduce on `main` independently of that PR — so this is a pre-existing bug, not a regression introduced by the playwright change.

haroldship:
## Bug fix

Fixes #214

### Summary

When `advanced_features.use_vision = true` (the default), `load_prompt_with_image` unconditionally builds the browser planner's user message as a list-typed multimodal `[image_url, text]` block. LLM endpoints that don't accept multimodal content (e.g. WatsonX served via an openai-compatible gateway) then reject the request with `400 messages[1].content must be a string`, killing hybrid-mode web-agent runs before the first step.

This change introduces a per-model `enable_vision` flag (resolved by the new `model_supports_vision` helper, defaulting to `True` to preserve existing behavior for vision-capable models). Both the prompt template structure (`load_prompt_with_image`) and the runtime image attachment (`BrowserPlannerAgent.run`) now gate on `settings.advanced_features.use_vision AND model_supports_vision(model_config)`. Operators whose LLM endpoint rejects multimodal content can set `enable_vision = false` on the relevant `agent.*.model` config (e.g. `agent.planner.model`) to force string-only user content.

### Why a per-model flag

The bug surfaces when the *endpoint* (not the model) cannot deserialize multimodal content blocks. Detecting endpoint capabilities at runtime is brittle, and silently downgrading vision would mask misconfigurations. An explicit `enable_vision = false` opt-out is small, discoverable, and additive — existing model configs continue to work unchanged because the default is `True`.

### Testing

Rebased onto `main` (`69b7e40a`) — branch contains a single commit `9b7bad82`.

New regression tests in `tests/unit/test_load_prompt_with_image_vision_gate.py` (7 tests):

- `model_supports_vision` defaults: `None` config and missing-attribute config both default to `True`; explicit `enable_vision=False` and `enable_vision=True` are honored.
- Human message is a **string** when the model disables vision, even with global `use_vision=True` (the regression for #214).
- Human message is **multimodal** (list with `image_url` + `text` items) when both global setting and per-model flag are `True`.
- Human message is a **string** when the global setting is `False` regardless of per-model flag.

#### Test results (against this branch, rebased on `main`)

| Suite | Result |
| --- | --- |
| `tests/unit/test_load_prompt_with_image_vision_gate.py` (new) | ✅ 7 / 7 passed |
| `src/cuga/backend/tools_env/registry/tests/` | ✅ all passed |
| `src/cuga/backend/cuga_graph/nodes/api/variables_manager/tests/` | ✅ all passed |
| `src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/tests/` (non-e2b) | ✅ all passed |
| `tests/unit/test_task_analyzer_app_matching.py` | ✅ all passed |
| `tests/unit/test_cuga_lite_bind_tools.py` | ✅ all passed |
| `src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/tests/test_e2b_lite.py` | ⚠️ 3 failed — `RuntimeError: e2b-code-interpreter package not installed` (env-only; CI runs these with `uv sync --extra e2b` per `src/scripts/run_tests.sh`) |
| `src/cuga/sdk_core/tests/` (live LLM integration) | ⚠️ flaky locally — pass individually, fail when run in large batches due to external-service rate limits / setup. Unrelated to this diff. |
| `ruff check` (touched files) | ✅ clean |
| `ruff format --check` (touched files) | ✅ clean |
| CodeRabbit CLI (`--type committed --base main`) | ✅ no findings against this diff |

Aggregate: 211 passing across the unit-test subset relevant to this change, with the only failures being env/network issues that also fail on `main`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Validate per-model vision support before attaching images; if a model rejects multimodal input, disable vision for subsequent attempts and retry with text-only payloads.
* **New Features**
  * Added a per-model vision capability check and a constructor-level toggle to control vision usage.
* **Tests**
  * Added tests covering vision gating, multimodal vs. text-only prompt formats, defaults, and fallback behavior when vision is disabled.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cuga-project/cuga-agent/pull/251?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/cuga/sdk_core/tests/test_sdk_token_usage_tracker.py
```diff
@@ -0,0 +1,225 @@
...
+        # Simulate LLM interaction by calling collect_prompt
+        tracker.collect_prompt(role="system", value="You are a helpful assistant")
+        tracker.collect_prompt(role="human", value="What is 2 + 2?")
+        tracker.collect_prompt(role="assistant", value="The answer is 4")
+
+        # Verify prompts were collected
+        assert len(tracker.prompts) == 3
+        assert tracker.prompts[0].role == "system"
+        assert tracker.prompts[1].role == "human"
+        assert tracker.prompts[2].role == "assistant"
```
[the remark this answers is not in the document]
haroldship: This comment is now stale — it targeted commit `819d5b56` which was based on the pre-merge branch for issue #71. PR #251 has since been rebased onto `main` (single commit `9b7bad82`), and the file referenced here (`src/cuga/sdk_core/tests/test_sdk_token_usage_tracker.py`) is no longer part of this PR's diff. Those changes are already merged via PR #236.

src/cuga/sdk.py
```diff
@@ -1341,6 +1343,58 @@ async def initialize(self):
...
+        callbacks: List[BaseCallbackHandler] = [TokenUsageTracker(tracker)]
+
+        # Add user-provided callbacks
+        if self._callbacks:
+            callbacks.extend(self._callbacks)
+            logger.debug(f"Built callbacks: TokenUsageTracker + {len(self._callbacks)} user callback(s)")
+        else:
+            logger.debug("Built callbacks: TokenUsageTracker only")
+
+        return callbacks
```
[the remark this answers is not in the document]
haroldship: This comment is now stale — it targeted commit `819d5b56` which was based on the pre-merge branch for issue #71. PR #251 has since been rebased onto `main` (single commit `9b7bad82`), and the file referenced here (`src/cuga/sdk.py`) is no longer part of this PR's diff. Those changes are already merged via PR #236.

haroldship:
@coderabbitai review

src/cuga/backend/llm/utils/helpers.py
```diff
@@ -78,6 +78,22 @@ def load_prompt_chat(system_path, relative_to_caller=True):
...
+    Defaults to ``True`` when the flag is missing so existing model configs
+    continue to use vision when the global ``advanced_features.use_vision``
+    setting is on. Set ``enable_vision = false`` on the model config to opt
+    out — the browser planner will then send a plain-string user message,
+    which avoids ``messages[1].content must be a string`` 400s from
+    openai-compatible endpoints that reject multimodal content (issue #214).
+    """
+    if model_config is None:
+        return True
+    return bool(getattr(model_config, 'enable_vision', True))
```
sami-marreed, line 94: so this checks if the enable_vision is given from model config? not like based on model e.g. claude 4.6 supports vision why do we need to add flag

haroldship: Good question — the short answer is the bug isn't actually about model capability; it's about endpoint capability.

The 400 in #214 came from a setup where the model itself *does* support vision (Llama 4 Maverick served via WatsonX, which is multimodal-capable), but the **openai-compatible gateway** in front of it doesn't accept multimodal content blocks — it rejects `messages[1].content` if it's a list. So a name-based allowlist ("claude-4.6 → multimodal OK") would have given the wrong answer here: it would have said yes, sent the list-typed payload, and produced the same 400.

The reason for the model name failing as a signal:

- Same model, multiple endpoints — e.g. WatsonX-via-openai-compatible vs WatsonX native. One accepts multimodal, the other doesn't.
- Provider names drift fast (litellm prefixes, openrouter aliases, custom `model_name` strings). A registry of "these names support vision" would need constant upkeep and would still miss the gateway dimension.
- Silently downgrading vision when we *think* the model doesn't support it would mask configuration mistakes — e.g. someone pointing at the wrong endpoint and not noticing because we quietly stripped images.

The `enable_vision` flag defaults to `True`, so anyone whose endpoint already works keeps working with no change. The only people who need to set it are those with a non-multimodal gateway — and for them, an explicit opt-out is more discoverable than "why did my screenshots stop being sent?". Open to a follow-up that adds a name-based default if that's useful for the common case, but I'd want it to be additive (a hint, not authoritative) so we can still override per endpoint.

sami-marreed: so can we try catch error when image doesn't work and simply then try without

haroldship:
Merged latest `main` into this branch (clean merge, no conflicts despite #250's `cuga_agent_core` refactor — this PR only touches `browser_planner_agent.py` and `llm/utils/helpers.py`, which #250 didn't move).

`uv run ruff format` + `uv run ruff check`: clean.
PR-specific tests (`tests/unit/test_load_prompt_with_image_vision_gate.py`): 7/7 passing.
`uv run pytest tests/unit/`: 189 passing, 56 deselected. The deselected tests (`test_manage_publish_sync.py`, `test_workspace_sandbox.py`, `test_tool_use_failed_recovery.py`, `test_skill_loader.py`) are all pre-existing failures on `main` at 27bd35e0 — unrelated to this PR.

haroldship:
Local stability test (`uv run run_stability_tests.py --method local`): **10/10 passed, 100% pass rate** after the main merge.

src/cuga/backend/llm/utils/helpers.py
haroldship: Good idea — implemented in 24ef14f0. `BrowserPlannerAgent.run()` now attempts the multimodal call first; if it raises with a vision-rejection signature (`content must be a string`, `image_url`, `vision`, etc.) and we actually attached an image, it logs a warning, flips `self.use_vision_effective = False` so the next turn doesn't repeat the rejection, and retries the *current* turn text-only (also flipping the `use_vision` template var so the prompt's image block stays in sync). Unrelated exceptions re-raise unchanged. The static `enable_vision` flag stays as a fast opt-out for configs where you already know the endpoint is text-only — but it's no longer required for correctness.

sami-marreed, line 94: Design LGTM. Explicit `enable_vision` on the model config is the right signal when the failure is the endpoint/gateway rejecting multimodal content, not the model lacking vision. Default `True` keeps existing setups unchanged.

No change needed in helpers.py — the runtime fallback in `BrowserPlannerAgent.run()` covers the try/catch / retry-without-image case.

src/cuga/backend/cuga_graph/nodes/browser/browser_planner_agent/browser_planner_agent.py
```diff
@@ -13,19 +14,45 @@
...
+    "image input",
+    "does not support images",
+    "vision",
+)
+
+
+def _looks_like_vision_rejection(exc: BaseException) -> bool:
+    """Heuristic: did the endpoint reject the request because of image content?"""
+    msg = str(exc).lower()
+    return any(marker in msg for marker in _VISION_REJECTION_MARKERS)
```
sami-marreed: Please change `_looks_like_vision_rejection(exc: BaseException)` to `exc: Exception`. `run()` already catches `Exception`; `BaseException` is too broad (e.g. `KeyboardInterrupt`, `SystemExit`) and shouldn't be part of this path.

src/cuga/backend/cuga_graph/nodes/browser/browser_planner_agent/browser_planner_agent.py
```diff
@@ -40,16 +67,32 @@ async def run(self, input_variables: AgentState) -> AIMessage:
...
+            return await self.chain.ainvoke(data)
+        except Exception as exc:
+            if not image_attached or not _looks_like_vision_rejection(exc):
+                raise
+            # Endpoint rejected the multimodal payload — disable vision for
+            # subsequent calls on this agent and retry with a text-only message.
+            logger.warning(
+                "Vision request rejected ({}); disabling vision for this agent and retrying text-only.",
+                type(exc).__name__,
+            )
```
sami-marreed, line 97: Please include `str(exc)` in the vision-rejection warning, not only `type(exc).__name__`. We need the actual API error text to validate and extend `_VISION_REJECTION_MARKERS` over time.

sami-marreed (changes requested):
Thanks for the #214 fix and the rebase.

Scope looks correct (helpers + browser planner + unit tests). The old CodeRabbit threads on sdk.py and test_sdk_token_usage_tracker.py are out of scope for this PR — fine to leave resolved.

Design LGTM: per-model enable_vision plus runtime multimodal→text-only fallback is the right approach for gateways that reject list-typed content. The retry in BrowserPlannerAgent.run() addresses the try/catch concern.

Requesting two small fixes on browser_planner_agent.py (inline). Once those are in, good to merge from my side.

src/cuga/backend/cuga_graph/nodes/browser/browser_planner_agent/browser_planner_agent.py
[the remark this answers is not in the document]
haroldship: Done in be8a969f — `_looks_like_vision_rejection(exc: Exception)`. Matches the `except Exception as exc` in `run()` so interpreter-control exceptions are never inspected by this heuristic.

haroldship: Done in be8a969f — signature is now `_looks_like_vision_rejection(exc: Exception)`.

src/cuga/backend/cuga_graph/nodes/browser/browser_planner_agent/browser_planner_agent.py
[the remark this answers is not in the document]
haroldship: Done in be8a969f — warning now logs both type and message: `logger.warning("Vision request rejected ({}: {}); ...", type(exc).__name__, exc)`.

haroldship: Done in be8a969f — `exc` is now interpolated into the warning alongside the type name, so the raw API error text is captured for future updates to `_VISION_REJECTION_MARKERS`.

src/cuga/backend/cuga_graph/nodes/browser/browser_planner_agent/browser_planner_agent.py
```diff
@@ -40,16 +67,32 @@ async def run(self, input_variables: AgentState) -> AIMessage:
...
         else:
             data["variables_history"] = ""
-        if settings.advanced_features.use_vision and getattr(tracker, "images", None):
+        image_attached = False
+        if self.use_vision_effective and getattr(tracker, "images", None):
             # Only attach an image if one has been captured
             if len(tracker.images) > 0:
                 data['img'] = tracker.images[-1]
-        return await self.chain.ainvoke(data)
+                image_attached = True
```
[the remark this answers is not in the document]
haroldship: Partially addressed in be8a969f. The local race in this file (length-check then `[-1]` indexing) is now an atomic try/except snapshot — concurrent `collect_image` calls can no longer cause `IndexError` or attach a value that has just been mutated away.

The broader recommendation (locking inside `ActivityTracker` or removing the singleton) is intentionally **out of scope for this PR** — `ActivityTracker` is a project-wide singleton used by many agents and a refactor there belongs in its own change. Tracking the file-local race here keeps the #214 fix focused.

src/cuga/backend/llm/utils/helpers.py
haroldship: Thanks — agreed, `helpers.py` stays as the fast opt-out via `enable_vision = false`, and the runtime try/catch in `BrowserPlannerAgent.run()` handles the case where the gateway/endpoint rejects multimodal content despite the model supporting it.

haroldship:
@sami-marreed both fixes are in (commit be8a969f) — `_looks_like_vision_rejection` now takes `Exception`, and the warning logs both type and message. CodeRabbit confirmed both are addressed. Ready for re-review when you have a moment.

sami-marreed (approved):

PR #252, opened 2026-05-20
fix(cuga_lite): recover code from unclosed ```python fence

### LINKED ISSUE #204: [Bug]: sdk_callback_node routes unexecuted code block as final answer
## Bug description

  When a CugaLite agent response contains a Python code block (the next action to execute), `sdk_callback_node` occasionally routes the raw unexecuted code directly to `FinalAnswerAgent` instead of passing it to the code executor.
   The user receives the Python source code as the chat response instead of the actual agent output.

  ## What happened

  Step 3 of a multi-step task. The LLM returned a response with `finish_reason: stop` — a Python block intended to list Gmail messages, read each one, and filter by allowlist. The block was syntactically valid and logically
  correct. Instead of executing it, `sdk_callback_node` (line 1497) routed the entire response — prose introduction + full code block — directly to `FinalAnswerAgent`.

  The agent never ran the code. The user received the raw `import json, re` Python source as the answer to their request.

  ## Relevant log lines

  21:02:40  DEBUG  call_model:1964  LLM response — type: AIMessage | content_len: 10642 | finish_reason: stop | tool_calls: []
  21:02:40  INFO   sdk_callback_node:1446  SDK callback node - processing subgraph results
  21:02:40  INFO   sdk_callback_node:1497  Routing to FinalAnswerAgent with answer: Config is available,
                   so the scan can proceed. I'll list all Gmail messages...
                   ```python
                   import json, re
                   from datetime import datetime, timedelta
                   ...

  Step count at the time: **3/70** — nowhere near the step limit.

### How to reproduce

unclear

### Environment

  - CUGA SDK (local editable install)
  - LLM: `azure/gpt-5.5` via OpenAI-compatible endpoint
  - Python 3.12, macOS

Attached: [2026-05-09_21:00:20.log](https://github.com/user-attachments/files/27613129/2026-05-09_21.00.20.log) — routing happens at line 249.

### Checklist

haroldship:
## Bug fix

Fixes #204

### Summary

When the LLM emits a Python code block whose closing fence is missing or truncated, `extract_and_combine_codeblocks` returned no matches because the regex (`\`\`\`python(.*?)\`\`\``) required a closing fence. `call_model` then fell through into the "no-code" branch and terminated the subgraph with the raw response stored as `final_answer`. `sdk_callback_node` routed the unexecuted Python source straight to `FinalAnswerAgent`, and the user received raw `import json, re` (etc.) as the chat reply — exactly the behavior reported in #204.

The fix adds a small recovery path after the strict regex fails: search for an unclosed `\`\`\`python` fence, strip any partial trailing fence, and **only return the candidate when it compiles**. The `compile()` guard keeps prose that happens to live after a stray opening fence from being mis-classified as code.

### Why guard with compile()

A naive recovery (return everything after `\`\`\`python\n`) would swallow follow-up narration the LLM might tack on after a malformed block. The compile check anchors the recovery to "this is actually Python", which matches the existing pattern in the same function for bare-code responses.

### Testing

Branch is a single commit (`8a2b6b4f`) on top of `main` (`69b7e40a`).

#### Bug-first: tests written before the fix

Following test-first practice, the new regression tests were added to `src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/tests/test_extract_codeblocks.py` **before** any change to `extract_and_combine_codeblocks`. With only the tests in place (no fix):

- ❌ `test_incomplete_markdown_block_with_valid_code_recovered` — **failed** (returns `''` instead of `'print("test")'`)
- ❌ `test_incomplete_markdown_block_recovers_multiline_code` — **failed** (returns `''` instead of the recovered multi-line code from the bug report)
- ✅ `test_incomplete_markdown_block_with_invalid_code_returns_empty` — passed (safety test for the compile guard; this property happens to hold both before and after, by construction)

After applying the fix, all 48 tests in the file pass.

##### Intentional replacement of an existing test

The pre-existing `test_incomplete_markdown_block` asserted the *buggy* behavior — that `\`\`\`python\nprint("test")` should return `''`. That test was replaced by `test_incomplete_markdown_block_with_valid_code_recovered`, which asserts the *fixed* behavior. This is an intentional behavior change required by issue #204: the old test was codifying the bug.

#### New regression tests

- `test_incomplete_markdown_block_with_valid_code_recovered` — `\`\`\`python\nprint("test")` (no closing fence) now returns `print("test")` instead of `''`.
- `test_incomplete_markdown_block_recovers_multiline_code` — the exact shape from the bug report (prose + unclosed fence + multi-line `import` + `print`) is recovered.
- `test_incomplete_markdown_block_with_invalid_code_returns_empty` — non-Python text after an open fence (`$$$ …`) is **not** treated as code (safety guard for the compile gate).

#### Test results (against this branch on `main`)

| Suite | Result |
| --- | --- |
| `src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/tests/test_extract_codeblocks.py` | ✅ 48 / 48 passed (45 pre-existing + 3 new) |
| `src/cuga/backend/tools_env/registry/tests/` | ✅ all passed |
| `src/cuga/backend/cuga_graph/nodes/api/variables_manager/tests/` | ✅ all passed |
| `tests/unit/test_cuga_lite_bind_tools.py` | ✅ all passed |
| `tests/unit/test_task_analyzer_app_matching.py` | ✅ all passed |
| `ruff check` (touched files) | ✅ clean |
| `ruff format --check` (touched files) | ✅ clean |

Aggregate: 155 unit tests pass across the subset relevant to this change.

### Testing
- [x] New unit tests written against current code first — confirmed failing before the fix, passing after

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced code extraction to recover valid Python code from incomplete or unclosed markdown code fences, including truncated fence markers and trailing content.

* **Tests**
  * Updated regression test suite to cover additional edge cases for incomplete markdown fence handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/cuga-project/cuga-agent/pull/252?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


haroldship:
@coderabbitai review

sami-marreed:
On which model did this happen

src/cuga/backend/cuga_graph/nodes/cuga_lite/cuga_lite_graph.py
```diff
@@ -723,6 +723,22 @@ def extract_and_combine_codeblocks(text: str) -> str:
     if code_blocks:
         return "\n\n".join(block.strip() for block in code_blocks)
 
+    # Recovery path for an unclosed ```python fence: without this the
```
sami-marreed: after refactor i am consildating this function it has multiple implementations across the code

src/cuga/backend/cuga_graph/nodes/cuga_lite/cuga_lite_graph.py
```diff
@@ -723,6 +723,22 @@ def extract_and_combine_codeblocks(text: str) -> str:
...
+    # so prose that happens to live after a stray opening fence is not
+    # mistaken for code.
+    unclosed = re.search(r'```python\s*\n(.*)', text, re.DOTALL)
+    if unclosed:
+        candidate = re.sub(r'\n?```\s*$', '', unclosed.group(1)).strip()
+        if candidate:
+            try:
+                compile(candidate.replace('await ', ''), '<string>', 'exec')
+                return candidate
+            except SyntaxError:
```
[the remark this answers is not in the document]
haroldship: Addressed in 6197cf15: the trailing-fence regex now matches 1–3 backticks (`r'\n?`{1,3}\s*$'`), so partial truncations no longer break the compile guard. Added `test_incomplete_markdown_block_strips_partial_trailing_fence` covering both the 1-backtick and 2-backtick truncation shapes.

src/cuga/backend/cuga_graph/nodes/cuga_lite/cuga_lite_graph.py
haroldship: Noted — happy to coordinate. This PR keeps the change local to `cuga_lite`'s copy (issue #204 is the immediate symptom there), so the consolidation can fold this in without needing to re-derive the recovery logic.

haroldship:
> On which model did this happen

On `azure/gpt-5.5` as per Issue #204

src/cuga/backend/cuga_graph/nodes/cuga_lite/cuga_lite_graph.py
sami-marreed: let's hold. this fix after refactor PR merge

sami-marreed:
please resovle changes

haroldship:
Merged latest `main` into this branch.

Conflict was caused by #250 (`refactor(agent-core): extract shared agent logic into cuga_agent_core module`), which moved `extract_and_combine_codeblocks` and `BACKTICK_PATTERN` from `cuga_lite_graph.py` into the new shared module `src/cuga/backend/cuga_graph/nodes/cuga_agent_core/execution/code_extraction.py`. Resolved by:

1. Porting the unclosed-fence recovery block (with the partial-fence stripping from CR feedback) to the new canonical location in `code_extraction.py`.
2. Dropping the now-orphan `extract_and_combine_codeblocks` from `cuga_lite_graph.py` (keeping it would re-introduce the duplication #250 cleaned up).
3. Keeping `_reflection_current_task` in `cuga_lite_graph.py` (Lite-only) and adding the missing `HumanMessage` import.

`test_extract_codeblocks.py` was already auto-merged to import from the new location; all 49 tests (including the four #204 regression tests) pass.

`uv run ruff format` + `uv run ruff check`: clean.
`uv run pytest tests/unit/`: 224 passing. The 12 failures (`test_workspace_sandbox.py`, `test_tool_use_failed_recovery.py`, `test_skill_loader.py`, `test_manage_publish_sync.py`) are pre-existing on `main` at 27bd35e0 — unrelated to this PR.

sami-marreed (commented):
Review: unclosed-fence recovery is the right fix for #204. A few edge cases and scope nits inline.

src/cuga/backend/cuga_graph/nodes/cuga_agent_core/execution/code_extraction.py
```diff
@@ -33,6 +33,25 @@ def extract_and_combine_codeblocks(text: str) -> str:
     if code_blocks:
         return "\n\n".join(block.strip() for block in code_blocks)
 
+    # Recovery path for an unclosed ```python fence: without this the
```
sami-marreed: These 5 comment lines repeat the PR description and module docstring — please trim to one short line (e.g. "Recover unclosed ```python fences (#204); compile-guarded.").

src/cuga/backend/cuga_graph/nodes/cuga_agent_core/execution/code_extraction.py
```diff
@@ -33,6 +33,25 @@ def extract_and_combine_codeblocks(text: str) -> str:
...
+    # sdk_callback_node routes the raw, unexecuted code straight to
+    # FinalAnswerAgent and the user sees Python source as the chat reply
+    # (issue #204). Only return the recovered candidate when it compiles,
+    # so prose that happens to live after a stray opening fence is not
+    # mistaken for code.
+    unclosed = re.search(r"```python\s*\n(.*)", text, re.DOTALL)
```
sami-marreed: Possible gap: `(.*)` grabs everything to EOF, so valid code followed by trailing prose in the same unclosed block still fails `compile()` and we return `""` — same #204 failure mode. Worth a regression test like:

```python
print("test")

Hope that helps!
```

If that shows up in prod, consider finding the longest compilable line-prefix after extraction (post-parse boundary trim) instead of all-or-nothing on the full tail.

src/cuga/backend/cuga_graph/nodes/cuga_agent_core/execution/code_extraction.py
```diff
@@ -33,6 +33,25 @@ def extract_and_combine_codeblocks(text: str) -> str:
...
+    if unclosed:
+        # Strip a trailing full OR partial markdown fence (1–3 backticks).
+        # Streamed responses can be truncated mid-fence; without this the
+        # straggler backticks make ``compile()`` fail and we'd drop valid code.
+        candidate = re.sub(r"\n?`{1,3}\s*$", "", unclosed.group(1)).strip()
```
sami-marreed: Please extract this unclosed-fence recovery into a dedicated helper at the bottom of the module, e.g. `_recover_non_closing_python_fence(text: str) -> str`, and call it from `extract_and_combine_codeblocks` after the strict `findall` path fails (before the bare-text fallback). Keeps the main function readable and makes the #204 path easier to unit-test in isolation.

Inside the helper: try `compile()` on the raw capture first; only on `SyntaxError`, strip trailing `` `{1,3}` `` and retry.

src/cuga/backend/cuga_graph/nodes/cuga_lite/cuga_lite_graph.py
```diff
@@ -134,6 +134,20 @@ def resolve_max_steps(self, state: CugaLiteState, override: Optional[int]) -> in
 _LITE_LOOP_ADAPTER = _CugaLiteLoopAdapter()
 
 
+def _reflection_current_task(state: CugaLiteState) -> str:
```
sami-marreed: `_reflection_current_task` is defined here but never called anywhere in this PR — looks like merge-conflict residue. Please drop it (and the `HumanMessage` import) from this PR or wire it up in a separate change; it is unrelated to #204.

src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/tests/test_extract_codeblocks.py
```diff
@@ -121,11 +121,45 @@ def test_markdown_without_python_tag(self):
         result = extract_and_combine_codeblocks('```\nprint("test")\n```')
         assert result == ''
 
-    def test_incomplete_markdown_block(self):
-        """Incomplete markdown block should not match."""
+    def test_incomplete_markdown_block_with_valid_code_recovered(self):
+        """Unclosed ```python fence containing syntactically valid code should be
```
sami-marreed: Same feedback on the new tests: the docstrings restate the PR/issue narrative at length. A single line per test (what input → what output) is enough — the issue link belongs in the PR body, not repeated in four docstrings.

haroldship:
Local stability test (`uv run run_stability_tests.py --method local`): **10/10 passed, 100% pass rate** after the main merge.

src/cuga/backend/cuga_graph/nodes/cuga_agent_core/execution/code_extraction.py
```diff
@@ -33,6 +33,25 @@ def extract_and_combine_codeblocks(text: str) -> str:
...
         return "\n\n".join(block.strip() for block in code_blocks)
 
+    # Recovery path for an unclosed ```python fence: without this the
+    # sdk_callback_node routes the raw, unexecuted code straight to
+    # FinalAnswerAgent and the user sees Python source as the chat reply
+    # (issue #204). Only return the recovered candidate when it compiles,
+    # so prose that happens to live after a stray opening fence is not
+    # mistaken for code.
+    unclosed = re.search(r"```python\s*\n(.*)", text, re.DOTALL)
+    if unclosed:
```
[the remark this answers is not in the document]
haroldship: Addressed in 447a8986: the same-line shape is now covered — `_recover_non_closing_python_fence` uses `r'```python\s*(.*)'` (no required \n), so ```python print("x")` recovers `print("x")`. Added `test_incomplete_same_line_python_fence_recovered`.

src/cuga/backend/cuga_graph/nodes/cuga_agent_core/execution/code_extraction.py
haroldship: Done in 447a8986 — trimmed the 5-line block to a single-sentence module docstring ("Shared fenced-code extraction and tool awaitable-wrapping utilities.") and removed the per-function multi-paragraph comments.

src/cuga/backend/cuga_graph/nodes/cuga_agent_core/execution/code_extraction.py
haroldship: Good catch — addressed in 447a8986. The new helper now walks the candidate line-by-line (`for end in range(len(lines), 0, -1)`) so `print("test")\n\nHope that helps!` returns just `print("test")` instead of failing compile and dropping to `""`. Regression test `test_incomplete_markdown_block_recovers_valid_code_with_trailing_prose` locks it down using your exact example.

src/cuga/backend/cuga_graph/nodes/cuga_agent_core/execution/code_extraction.py
haroldship: Done in 447a8986 — extracted `_recover_non_closing_python_fence(text: str) -> str` at the bottom of `code_extraction.py`. `extract_and_combine_codeblocks` now calls it between the strict `findall` path and the bare-text fallback, keeping the main function readable.

src/cuga/backend/cuga_graph/nodes/cuga_lite/cuga_lite_graph.py
haroldship: Done in 447a8986 — dropped `_reflection_current_task` and removed `HumanMessage` from the import (`AIMessage, BaseMessage` remain in use). The live caller in `adapter/sandbox_node.py` uses `adapter/response_utils.reflection_current_task`, so nothing else needed wiring.

src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/tests/test_extract_codeblocks.py
haroldship: Done in 447a8986 — collapsed each new test's docstring to a single-line behavior summary (e.g. "Unclosed fence with valid code returns the code."). The #204 narrative now lives only in the PR body.

sami-marreed (approved):

PR #315, opened 2026-06-08
fix(cuga_lite): reset task_todos state on fresh conversation

### LINKED ISSUE #314: [Bug]: cross-task leak of `task_todos_ref` pollutes turn-1 system prompt when `enable_todos=true`
### What happened

When `advanced_features.enable_todos=true` (e.g. `DYNACONF_ADVANCED_FEATURES__ENABLE_TODOS=true`), the per-task plan written by `create_update_todos` leaks across `CugaAgent.invoke()` calls. From the second task onward, the LLM's **very first** turn starts with a `## Current task todos` block containing the previous task's plan — **before** the model has had any chance to call the tool for the new task. This biases the model's first reasoning step toward an unrelated plan and (combined with the "isolated tool" execution rule) measurably degrades benchmark performance.

### Root cause

`task_todos_ref` is a closure-scoped mutable list created once inside `create_cuga_lite_graph` and captured by the adapter:

- `cuga-agent/src/cuga/backend/cuga_graph/nodes/cuga_lite/cuga_lite_graph.py:180-191` — `task_todos_ref: List[Dict[str, str]] = []` created once per compiled graph.
- `cuga-agent/src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/graph_adapter.py:62,92-97` — `prepare_system_content` checks `self._task_todos_ref` first and appends `format_task_todos_system_block(...)` to the system prompt.
- `cuga-agent/src/cuga/backend/cuga_graph/nodes/cuga_agent_core/execution/todos.py:123-126` — the tool body only does `clear() + extend(new)`, never resets on a new task.
- `cuga-agent/src/cuga/sdk.py:1459` — `CugaAgent._create_graph` memoizes on `self._graph`; only `add_tool()` (sdk.py:2224) nulls it. So `.invoke()` reuses the same compiled graph (and the same closure-captured ref) for every task.
- `cuga-agent/src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/graph_adapter.py:94-96` — `state.task_todos` is a parallel leak path when a thread-keyed checkpointer is used.

Nothing in the SDK, agent loop, or framework resets these between invocations.

### How to reproduce

1. Set `DYNACONF_ADVANCED_FEATURES__ENABLE_TODOS=true`.
2. Use any harness that reuses a single `CugaAgent` across multiple `.invoke()` calls (the standard cuga-eval AppWorld harness at `cuga-eval/benchmarks/appworld/eval_appworld_sdk.py` does this — `setup_agent_with_tools()` is called once, then 43 tasks loop through `self.agent.invoke(...)`).
3. Capture Langfuse traces and inspect the **first LLM generation** of each task. From task #2 onward, the system prompt already contains a `## Current task todos` block that the model has not yet authored for this task.

### Evidence (43-task AppWorld run, GPT-4.1)

Of 43 trace files in a comparison bundle:

| | Result |
|---|---|
| Chronologically first task | clean — no `## Current task todos` block in turn 1 |
| All 42 subsequent tasks | **non-empty `## Current task todos` block in turn 1**, 3–5 items each |

Concrete example — task whose user query asks about **HR / candidates / `report_template.md`**, but whose turn-1 system prompt begins with a stale plan from a prior Gmail task:

```
## Current task todos
1. **[pending]** Fetch all unread email threads from Gmail inbox (handle pagination)
2. **[pending]** Filter threads for priority-1
3. **[pending]** Count the number of priority-1 unread threads
4. **[pending]** Return the count to the user
```

The model proceeds with the new HR task but does so under a misleading "Current Plan" that has nothing to do with the user's actual question.

### Suggested fix

One-line clear at the top of `prepare_tools_and_apps` in `cuga-agent/src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/prepare_node.py`, guarded on "fresh conversation" so HITL resume isn't affected:

```python
# task_todos_ref is closure-scoped per compiled graph; state.task_todos
# may also persist via a thread-keyed checkpointer. Reset both at the
# start of a new conversation.
if len(state.chat_messages or []) <= 1:
    adapter._task_todos_ref.clear()
    state.task_todos = None
```

This places the reset in the framework (so SDK and direct `create_cuga_lite_graph` users both benefit) and doesn't require any caller cooperation.

### Severity

- Default config has `enable_todos=false`, so users on defaults are unaffected.
- For anyone enabling todos in long-running harnesses (benchmarks, multi-task agents), every task after the first is affected.

### Environment

- cuga-agent: from `pyproject.toml` (editable install)
- Python 3.11, macOS (darwin 25.4.0)
- Reproduced on the AppWorld benchmark in cuga-eval

### Checklist

Sergey-Zeltyn:
## Bug fix

Fixes #314.

### Summary

When `advanced_features.enable_todos=true`, the closure-scoped `task_todos_ref` (plus the parallel `state.task_todos` checkpointer field) outlived a single `agent.invoke()`. From the second task onward, the previous task's plan leaked into the new task's turn-1 system prompt as `## Current task todos`, biasing the model's first reasoning step before it had any chance to call `create_update_todos` for the new task.

**Root cause:** `CugaAgent._create_graph` memoizes on `self._graph` and only resets in `add_tool()` ([sdk.py:2224](https://github.com/cuga-project/cuga-agent/blob/main/src/cuga/sdk.py#L2224)), so every `.invoke()` reuses the same compiled graph and the same closure-captured list created in `create_cuga_lite_graph` ([cuga_lite_graph.py:180](https://github.com/cuga-project/cuga-agent/blob/main/src/cuga/backend/cuga_graph/nodes/cuga_lite/cuga_lite_graph.py#L180)). The tool body in [todos.py:123-126](https://github.com/cuga-project/cuga-agent/blob/main/src/cuga/backend/cuga_graph/nodes/cuga_agent_core/execution/todos.py#L123-L126) only does `clear() + extend(new)` — it never resets on a new task. Nothing in the SDK, agent loop, or framework cleared these between invocations.

### Scope

This PR fixes the leak **on a fresh conversation** — detected by `len(state.chat_messages) <= 1`, i.e. an invoke where the only message present is the user's first turn. That covers the production failure mode in the AppWorld harness (one `agent.invoke()` per task, fresh state) and the SDK direct-invoke pattern.

**Out of scope** (will be filed as a separate follow-up issue):
- *New task as a follow-up turn on the same thread.* If a multi-turn conversation has `chat_messages > 1` but the user has changed topic, this PR's guard is false and the leak shape can recur. Detecting a topic change requires more than a message count — a follow-up.
- *Thread-scoped todos for concurrency.* The closure-shared `_task_todos_ref` is a process-global singleton; `.clear()` races across concurrent threads. Pre-existing (the `create_update_todos` tool already mutates the shared ref), but worth a follow-up.

### Fix

At the top of `prepare_tools_and_apps` ([prepare_node.py](https://github.com/cuga-project/cuga-agent/blob/main/src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/prepare_node.py)):

```python
is_fresh_conversation = len(state.chat_messages or []) <= 1
if is_fresh_conversation:
    adapter._task_todos_ref.clear()      # closure-scoped, mutated in place
    state.task_todos = None               # local consistency only — see below
```

`state.task_todos = None` alone is a no-op because LangGraph reduces state only from the returned `Command.update`. The reset is therefore also propagated through **both** return paths of this node:

- **Policy early-return path** (`BLOCK_INTENT` / playbook / tool-guide): if `is_fresh_conversation`, defensively `setdefault("task_todos", None)` on the policy-returned `Command.update` dict so the reset survives.
- **Main return path** (`goto="call_model"`): build `update_payload` as a local dict and include `"task_todos": None` when `is_fresh_conversation`, then construct the `Command`.

The guard preserves HITL resume (turn 2+ keeps the in-flight plan).

### Verification

Reproduced and verified on a 43-task AppWorld eval bundle (gpt-4.1, `enable_todos=true`):

| Metric | Before fix | After fix |
|---|---:|---:|
| Traces with `## Current task todos` in **turn-1 system prompt** | **42 / 43** | **0 / 43** |
| Traces with `## Current Plan` in turn-1 (state path) | — | 0 / 43 |
| Mean fraction of LLM turns spent on todo-only blocks | 7.1% | 5.6% |

(AppWorld doesn't exercise the state-path checkpointer fallback, so the original commit's no-op for that path didn't show up in the eval; the must-fix commit closes the path nevertheless. See review comment by @sami-marreed.)

Concrete example from the pre-fix bundle — task asking about HR/candidates/`report_template.md` opens with a stale Gmail plan injected as authoritative:

```
## Current task todos
1. **[pending]** Fetch all unread email threads from Gmail inbox (handle pagination)
2. **[pending]** Filter threads for priority-1
3. **[pending]** Count the number of priority-1 unread threads
4. **[pending]** Return the count to the user
```

After the fix, every one of the 43 traces starts turn 1 with a clean system prompt.

### Commits in this PR

- `1780408` fix(cuga_lite): reset task_todos state between invocations (initial fix — closes the closure-ref path)
- `8aaace2` fix(cuga_lite): propagate task_todos reset through Command.update (must-fix from review — closes the state-path checkpointer fallback)

### Testing
- [x] Re-ran 43-task AppWorld eval with `enable_todos=true`: leak signature dropped from 42/43 to 0/43
- [x] Existing unit tests under `cuga_lite/tests/test_agent_graph_adapter.py` exercise `prepare_system_content` directly (not `prepare_tools_and_apps`), so they're unaffected by the new clear logic
- [ ] A regression test that seeds stale `_task_todos_ref` + a checkpointed `task_todos` and asserts both are cleared after a fresh-conversation invoke — drafted, to be added in a follow-up commit per the review thread.

### DIFF
--- a/src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/prepare_node.py
+++ b/src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/prepare_node.py
@@ -70,6 +70,15 @@ def create_prepare_tools_and_apps_node(adapter: Any, lc_bind_tools_meta: dict) -
         Disable few-shots entirely via ``advanced_features.cuga_lite_enable_few_shots`` in settings.toml
         or ``cuga_lite_enable_few_shots`` in configurable (skips prefix chat few-shots).
         """
+        # The adapter's task_todos_ref is closure-scoped per compiled graph and
+        # outlives a single .invoke(); state.task_todos can also persist via a
+        # thread-keyed checkpointer. Reset both at the start of a new
+        # conversation so a previous task's plan doesn't leak into this one's
+        # turn-1 system prompt.
+        if len(state.chat_messages or []) <= 1:
+            adapter._task_todos_ref.clear()
+            state.task_todos = None
+
         configurable = config.get("configurable", {}) if config else {}
         enable_todos = (
             configurable.get("enable_todos")
--- a/src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/prepare_node.py
+++ b/src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/prepare_node.py
@@ -70,12 +70,18 @@ def create_prepare_tools_and_apps_node(adapter: Any, lc_bind_tools_meta: dict) -
         Disable few-shots entirely via ``advanced_features.cuga_lite_enable_few_shots`` in settings.toml
         or ``cuga_lite_enable_few_shots`` in configurable (skips prefix chat few-shots).
         """
-        # The adapter's task_todos_ref is closure-scoped per compiled graph and
+        # Reset cross-task state at the start of a fresh conversation.
+        # adapter._task_todos_ref is closure-scoped per compiled graph and
         # outlives a single .invoke(); state.task_todos can also persist via a
-        # thread-keyed checkpointer. Reset both at the start of a new
-        # conversation so a previous task's plan doesn't leak into this one's
-        # turn-1 system prompt.
-        if len(state.chat_messages or []) <= 1:
+        # thread-keyed checkpointer. We detect "fresh conversation" by message
+        # count (<=1 means just the user's initial message).
+        # The closure ref is cleared in place; state.task_todos is propagated
+        # via every Command.update returned below because in-place state
+        # mutation does not persist through LangGraph.
+        # Note: this does not cover "new task as a follow-up turn on the same
+        # thread" (chat_messages > 1 with a topic change) — tracked separately.
+        is_fresh_conversation = len(state.chat_messages or []) <= 1
+        if is_fresh_conversation:
             adapter._task_todos_ref.clear()
             state.task_todos = None
 
@@ -117,6 +123,11 @@ def create_prepare_tools_and_apps_node(adapter: Any, lc_bind_tools_meta: dict) -
 
             # If policy returned a command (e.g., BLOCK_INTENT), execute it immediately
             if command:
+                # Propagate the task_todos reset through the policy-returned
+                # Command so LangGraph persists it; the in-place state mutation
+                # above does not survive the node return.
+                if is_fresh_conversation and isinstance(getattr(command, "update", None), dict):
+                    command.update.setdefault("task_todos", None)
                 return command
 
             # If policy returned metadata (e.g., playbook guidance), store it
@@ -627,19 +638,23 @@ def create_prepare_tools_and_apps_node(adapter: Any, lc_bind_tools_meta: dict) -
 
         reflection_apps_snapshot = format_apps_for_prompt(apps_for_prompt or [])
 
-        return Command(
-            goto="call_model",
-            update={
-                "tools_prepared": True,
-                "prepared_prompt": dynamic_prompt,
-                "step_count": 0,
-                "cuga_lite_metadata": state.cuga_lite_metadata,
-                "reflection_apps": reflection_apps_snapshot,
-                "reflection_enable_find_tools": enable_find_tools,
-                "reflection_skills_enabled": skills_enabled,
-                "reflection_skills_prompt_section": skills_prompt_section,
-                "mcp_few_shot_messages": few_shot_examples,
-            },
-        )
+        update_payload: dict[str, Any] = {
+            "tools_prepared": True,
+            "prepared_prompt": dynamic_prompt,
+            "step_count": 0,
+            "cuga_lite_metadata": state.cuga_lite_metadata,
+            "reflection_apps": reflection_apps_snapshot,
+            "reflection_enable_find_tools": enable_find_tools,
+            "reflection_skills_enabled": skills_enabled,
+            "reflection_skills_prompt_section": skills_prompt_section,
+            "mcp_few_shot_messages": few_shot_examples,
+        }
+        if is_fresh_conversation:
+            # Carry the task_todos reset through LangGraph state so the
+            # state.task_todos fallback path in prepare_system_content sees an
+            # empty value on the next turn.
+            update_payload["task_todos"] = None
+
+        return Command(goto="call_model", update=update_payload)
 
     return prepare_tools_and_apps
--- a/src/cuga/backend/cuga_graph/nodes/cuga_lite/tests/test_prepare_node_task_todos_reset.py
+++ b/src/cuga/backend/cuga_graph/nodes/cuga_lite/tests/test_prepare_node_task_todos_reset.py
@@ -1,0 +1,147 @@
+"""Regression test for issue #314 — task_todos reset on fresh conversation.
+
+Verifies that prepare_tools_and_apps:
+1. Clears the closure-scoped adapter._task_todos_ref in place when a
+   conversation starts (chat_messages <= 1)
+2. Propagates the reset through the returned Command.update so LangGraph
+   persists state.task_todos = None — the failure mode flagged in the
+   review of PR #315 where the in-place state mutation was a no-op
+3. Does NOT touch either field on a continuing conversation (HITL resume
+   or follow-up turn)
+"""
+
+from __future__ import annotations
+
+from types import SimpleNamespace
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+from langchain_core.messages import AIMessage, HumanMessage
+
+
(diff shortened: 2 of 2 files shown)


sami-marreed:
## Code Review

**Verdict: merge with fixes.** The primary half of the fix is correct and demonstrably works (42/43→0/43 on the eval), but the second half — the `state.task_todos` reset — is effectively a no-op, and the scope is narrower than the title claims.

### What's solid
- **Root-cause diagnosis is right.** `_task_todos_ref` is a closure-shared list (`cuga_lite_graph.py:180`) held by the long-lived adapter (`graph_adapter.py:62`) and read **first** in `prepare_system_content` (`graph_adapter.py:92`). It's the dominant leak path, and `_task_todos_ref.clear()` fixes it for real — the ref lives outside LangGraph channels, so the mutation is immediately visible to `call_model`.
- `.clear()` is safe — the ref is always a list, never `None`.
- Correctly placed at the top of the START node, before any prompt construction.

### Issues

**🔴 Critical — `state.task_todos = None` is a no-op (`prepare_node.py:78`).**
The node returns `Command(goto="call_model", update={...})` and that update dict (`prepare_node.py:630-642`) does **not** contain `task_todos`. LangGraph reduces state only from the returned update; the in-place assignment is discarded. So when `_task_todos_ref` is empty, `prepare_system_content` falls through to the still-stale `state.task_todos` (`graph_adapter.py:94`) — the exact fallback this PR claims to close.
**Fix:** add `"task_todos": None` to the `Command.update` dict (gated on the same reset condition), not just the local mutation. One line.

**🟠 Important — guard doesn't cover new-task-on-same-thread.**
Both invoke paths *accumulate* `chat_messages` (SDK append path; `cuga_lite_node.py:312` "Preserve existing chat history"). A genuinely new task as a follow-up turn has `chat_messages > 1`, so the guard is false and nothing clears → the original leak can still occur. The fix only covers the fresh-invoke-per-task scenario — which is exactly the AppWorld harness it was verified on. The title "between invocations" overstates coverage; consider re-scoping to "on a fresh conversation (≤1 message)."

**🟠 Important — HITL preservation is incidental, not designed.**
Resume re-enters the subgraph from START so `prepare` re-runs, but at that point `chat_messages > 1` so the guard is already false. HITL is preserved by message-count accumulation, not by design intent — fine outcome, but the comment's framing is coincidental.

**🟡 Minor**
- **Concurrency:** the adapter / `_task_todos_ref` is a process-global singleton (built once at startup); `.clear()` races across concurrent threads. Pre-existing (the todos tool already mutates the shared ref), but worth a follow-up for thread-scoped todos.
- **No regression test** (+9/−0). A test that seeds a stale `task_todos` + populated `_task_todos_ref`, runs a fresh invoke, and asserts both the prompt section *and* the persisted `state.task_todos` are empty would catch the Critical bug directly.

### Recommendation
1. **Must-fix:** add `task_todos` to the `Command.update` dict — otherwise the checkpointer fallback path stays open.
2. Add the regression test covering both reset targets.
3. Tighten the title/description (scope to fresh conversation; drop the "between invocations" generality).
4. File a follow-up for thread-scoped todos (concurrency + multi-turn coverage).

Sergey-Zeltyn:
@sami-marreed thanks for the careful review — addressed in two follow-up commits:

- **`8aaace2` fix(cuga_lite): propagate task_todos reset through Command.update** — closes the Critical finding. `is_fresh_conversation` is now captured up front; both return paths carry `task_todos: None`: the policy early-return path defensively merges via `setdefault` into the policy-returned `Command.update`, and the main `goto="call_model"` return builds `update_payload` as a local dict and includes the key when fresh. The in-place `state.task_todos = None` is kept only for in-process consistency.
- **`8c13ac8` test(cuga_lite): regression test for task_todos reset on fresh conversation** — `tests/test_prepare_node_task_todos_reset.py` with two cases. The positive case seeds both `_task_todos_ref` and `state.task_todos`, asserts the ref clears AND `Command.update["task_todos"] is None` — would have caught the no-op directly. The negative case seeds 3 messages and asserts the ref is preserved AND `Command.update` does NOT contain `task_todos`, preventing a future regression that erases the in-flight plan on HITL resume.

PR title and body have been rescoped to "on fresh conversation (≤1 message)" and the body now explicitly lists the two out-of-scope follow-ups (new-task-on-same-thread, thread-scoped todos for concurrency) so the coverage claim matches reality. CI is green on the new commits (CodeRabbit is still re-reviewing as of this comment).

Happy to address anything else.

sami-marreed (approved):

Issue #322, opened 2026-06-09
Saved-config startup wipes MODEL_NAME env var when llm.model is empty (force_env mode)


laredo:
## Summary

When `secrets.force_env = true` (the default in `settings.toml`) and the
`agent_configs` row saved in `src/cuga/dbs/cuga.db` has an empty
`llm.model`, lifespan startup deletes the `MODEL_NAME` env var the
operator set in their shell or `.env`. Every subsequent LLM resolution
then falls back to the hardcoded `gpt-4o`, which fails on most
production LLM gateways with `401 team_model_access_denied`.

This bites anyone who runs CUGA against a gateway whose allow-list does
not include `gpt-4o` — observed today against IBM's `ete-litellm`
gateway (allowed: `Azure/gpt-4.1`, claude variants, etc.). It also
silently breaks the new two-CUGA A2A demo at
`docs/examples/a2a_two_cuga/`, where `_preflight.py` patches the saved
DB row as a workaround. That workaround should not be necessary.

## Reproduce

1. Set `MODEL_NAME=Azure/gpt-4.1` (or any non-`gpt-4o` model) in `.env`.
2. Confirm `[secrets].force_env = true` and `mode = "local"` in
   `src/cuga/settings.toml` (current defaults).
3. Make sure `src/cuga/dbs/cuga.db` has at least one row in
   `agent_configs` (the default DB shipped in this repo does — version
   `draft` and version `1` for `cuga-default`, both with `llm.model = ""`).
4. Start the backend:
   `uvicorn cuga.backend.server.main:app --port 7860`
5. Watch the logs:
   - Early lines (during policy init): `_get_model_name:272 - Using MODEL_NAME from environment: Azure/gpt-4.1` ← good.
   - Then: `lifespan:695 - Startup: applied saved LLM config — provider=%s model=%s`.
   - Every subsequent line: `_get_model_name:280 - No model_name specified, using default: gpt-4o`.
6. Send any chat request. Result:
   `401 - {'error': {'message': \"team not allowed to access model. ... Tried to access gpt-4o\"}}`.

## Root cause

`src/cuga/backend/server/manage_routes.py` (in `_apply_published_config`,
around lines 182-194):

```python
if force_env:
    if \"model\" in llm_cfg and llm_cfg[\"model\"]:
        os.environ[\"MODEL_NAME\"] = str(llm_cfg[\"model\"])
    else:
[...]

PR #323, opened 2026-06-09
feat(a2a): expose CUGA over A2A v0.3 + working two-CUGA demo

### LINKED ISSUE #313: [Feature]: Use CUGA through A2A
CUGA can coordinate agents via A2A but cannot be part of an A2A collaboration. 

Today, CUGA can run as an MCP server — documented at docs/examples/cuga_as_mcp/README.md. So the "expose CUGA to other systems" story today goes through MCP, not A2A.

Bottom line: CUGA speaks A2A as a consumer (its supervisor orchestrates external A2A agents), but it cannot currently be invoked as an A2A agent by others. The a2a-sdk[http-server] dependency is present, so the building blocks are there if you wanted to add inbound support — it'd mean implementing an AgentExecutor wrapping CugaAgent.invoke(), mounting an A2AStarletteApplication, and serving an agent card.

laredo:
## Summary

- Adds an opt-in A2A server module (`src/cuga/backend/server/a2a/`) that exposes CUGA over A2A v0.3 (JSON-RPC) — `/.well-known/agent.json`, `/.well-known/agent-card.json`, and a `/a2a` endpoint supporting `message/send` and SSE `message/stream`. Mounted only when `settings.a2a.enabled = true`, so disabled deployments are byte-identical to before.
- Wires the inbound endpoint to a lazily-instantiated `CugaSupervisor` configured via `settings.a2a.supervisor_config_path`. Runner glue (`A2AStreamEvent`, `SupervisorA2ARunner`, `PlaceholderA2ARunner`, `build_a2a_router_for_settings`) lives in `src/cuga/backend/server/a2a/runner.py`; `main.py` is a 5-line opt-in mount.
- Rewrites the **outbound** `delegate_task_via_a2a_sdk` to speak v0.3 JSON-RPC directly via `httpx`. The previous implementation imported names removed in `a2a-sdk` 1.x (`A2AClient`, `MessageSendParams`, `TextPart`, `JSONRPCErrorResponse`, `a2a.utils.message`), making `HAS_A2A_SDK = False` and silently disabling HTTP-A2A delegation in production.
- Handles **both** AgentCard shapes the SDK can return: the v0.3 pydantic model (top-level `.url`) and the v1.0 protobuf model that `A2ACardResolver.get_agent_card()` actually returns (URL lives at `supported_interfaces[i].url` with a `protocol_binding`). Multi-binding peers prefer the JSONRPC interface; non-HTTP fallbacks (e.g. `grpc://`) are skipped so the JSON-RPC sender never POSTs to a scheme it can't speak.
- Maps terminal events named `error`/`failed`/`failure`/`exception` to `TaskState.failed` instead of silently flattening every terminal to `completed`. Outbound delegate now propagates peer Task `status.state=failed` as caller-visible `status="failed"` and raises when the response lacks a recognizable Task envelope (instead of returning empty success).
- SSE generator wraps runner/adapter calls in try/except and emits a final JSON-RPC error frame on failure. Wire-bound error messages no longer leak `str(exc)` — they carry only the exception class name; the full traceback is logged via `logger.exception`.
- Adds a working two-CUGA demo at `docs/examples/a2a_two_cuga/`: provider YAML (digital_sales toolset), consumer YAML (no local tools, one external A2A agent), launch scripts, and a preflight that loads `.env` and works around #322.
- Test coverage: **46 passed, 1 xfailed** (auth-token forwarding, intentionally deferred). New `test_outbound_protocol.py` covers `_extract_text_from_task`, the proto-shape AgentCard URL extraction, JSONRPC-vs-GRPC interface preference, end-to-end roundtrip via `httpx.ASGITransport`, JSON-RPC error envelope handling, peer-Task-failure surfacing, and missing-Task-envelope raising. New `test_a2a_router.py` cases cover SSE error frames and exception-text scrubbing. `test_task_adapter.py` covers error → failed mapping.
- 100% docstring coverage on every changed Python file (was 45.16%, threshold 80%).

## End-to-end verification

Smoke-tested live in this branch on macOS, Azure/gpt-4.1 via `ete-litellm` gateway:

```
# Terminal 1: cuga start registry
# Terminal 2: docs/examples/a2a_two_cuga/run_provider.sh   → :8002
# Terminal 3: docs/examples/a2a_two_cuga/run_consumer.sh   → :7860
# Browser:    http://localhost:7860 → "list my territory accounts"
```

Result: consumer chat UI → consumer supervisor → A2A JSON-RPC → provider on :8002 → digital_sales tools → 50 real territory accounts → final answer back in the chat UI. Provider log shows `POST /a2a → 200`, consumer log shows `Delegating to provider: List my territory accounts...`.

## Commits (7)

1. `4d51fe6` — initial opt-in A2A inbound module with placeholder runner
2. `66ba88c` — real graph runner + outbound v0.3 client + two-CUGA demo
3. `4183c63` — protobuf AgentCard URL extraction (fixes "Agent card carries no URL")
4. `5bdac5f` — CodeQL autofix for URL substring sanitization in a test
5. `7be1dca` — docstrings on every changed function/class + ruff format
6. `7bb1487` — error terminals → `failed`; surface peer failures; harden SSE; scrub exception text
7. `8cd7e9e` — extract runner setup from `main.py` to `a2a/runner.py`; lint nits

## Related

- Issue #322 — saved-config startup wipes `MODEL_NAME` env var when `llm.model` is empty (force_env mode). The demo's `_preflight.py` is a workaround; once #322 lands, the preflight can be deleted.

## Out of scope

- Auth-token forwarding across A2A (`xfail` test stays xfail).
- Per-step streaming through `message/stream` — currently folds the whole supervisor run into one terminal frame.

## Test plan

- [x] `uv run pytest tests/unit/a2a/ tests/integration/a2a/` → 46 passed, 1 xfailed
- [x] `uv run ruff check src/cuga/backend/server/a2a/ src/cuga/backend/server/main.py src/cuga/backend/cuga_graph/nodes/cuga_supervisor/a2a_protocol.py docs/examples/a2a_two_cuga/_preflight.py tests/integration/a2a/ tests/unit/a2a/` → clean
- [x] `uv run ruff format --check` on the same paths → clean
- [x] Live cross-process round-trip with the digital_sales demo on macOS, Azure/gpt-4.1 via `ete-litellm` gateway
- [x] Review feedback addressed: see PR comment summarizing per-thread responses
- [ ] Reviewer: confirm the opt-in mount truly preserves the disabled-deployment surface (no new imports under `settings.a2a.enabled = false`)
- [ ] Reviewer: spot-check the proto-shape `_agent_card_rpc_url` against any internal A2A peers that aren't CUGA-fronted

### DIFF
--- a/src/cuga/backend/server/main.py
+++ b/src/cuga/backend/server/main.py
@@ -1634,23 +1634,69 @@ if getattr(settings, "a2a", None) and getattr(settings.a2a, "enabled", False):
     # so disabled deployments pay no import-time cost for it.
     from cuga.backend.server.a2a import build_router as _build_a2a_router  # noqa: E402
 
-    class _PlaceholderA2ARunner:
-        """Minimal runner used until production graph wiring lands.
+    class _A2AStreamEvent:
+        """Duck-typed event the A2A task adapter consumes."""
 
-        Yields a single placeholder event so an A2A client gets a clean
-        lifecycle (working → completed) instead of a hung task.
+        __slots__ = ("name", "data", "final")
+
+        def __init__(self, name, data=None, final=False):
+            self.name = name
+            self.data = data
+            self.final = final
+
+    class _SupervisorA2ARunner:
+        """Run inbound A2A messages through a lazily-created CugaSupervisor.
+
+        We instantiate the supervisor on first use rather than during
+        lifespan startup so deployments that enable A2A but never receive
+        a request pay no init cost. The supervisor is cached on
+        ``app_state.a2a_supervisor`` afterward.
+        """
+
+        def __init__(self, app_state_ref, supervisor_config_path: str):
+            self._app_state = app_state_ref
+            self._yaml_path = supervisor_config_path
+            self._lock = asyncio.Lock()
+
+        async def _ensure_supervisor(self):
+            existing = getattr(self._app_state, "a2a_supervisor", None)
+            if existing is not None:
+                return existing
+            async with self._lock:
+                existing = getattr(self._app_state, "a2a_supervisor", None)
+                if existing is not None:
+                    return existing
+                from cuga.sdk import CugaSupervisor
+
+                supervisor = await CugaSupervisor.from_yaml(self._yaml_path)
+                self._app_state.a2a_supervisor = supervisor
+                return supervisor
+
+        async def run(self, message, context_id=None):
+            try:
+                supervisor = await self._ensure_supervisor()
+                result = await supervisor.invoke(message, thread_id=context_id)
+                answer = getattr(result, "answer", None) or str(result)
+                error = getattr(result, "error", None)
+                if error:
+                    yield _A2AStreamEvent("error", {"text": f"Supervisor error: {error}"}, final=True)
+                    return
+                yield _A2AStreamEvent("final_answer", {"text": answer}, final=True)
+            except Exception as exc:  # surface failures across the wire
+                logger.exception("A2A inbound delegation failed")
+                yield _A2AStreamEvent("error", {"text": f"A2A handler error: {exc}"}, final=True)
+
+    class _PlaceholderA2ARunner:
+        """Used when ``a2a.supervisor_config_path`` is unset.
+
+        Returns a clear "endpoint reached but unconfigured" terminal event
+        so callers get a well-formed Task envelope instead of a 5xx.
         """
 
         async def run(self, message, context_id=None):
-            class _Ev:
-                def __init__(self, name, data, final=False):
-                    self.name = name
-                    self.data = data
-                    self.final = final
-
-            yield _Ev(
+            yield _A2AStreamEvent(
                 "final_answer",
-                {"text": "A2A inbound endpoint reached, but graph runner is not yet wired."},
+                {"text": "A2A inbound endpoint reached, but settings.a2a.supervisor_config_path is not set."},
                 final=True,
             )
 
@@ -1662,7 +1708,14 @@ if getattr(settings, "a2a", None) and getattr(settings.a2a, "enabled", False):
         "auth_required": getattr(settings.a2a, "auth_required", False),
         "skill_ids": list(getattr(settings.a2a, "skill_ids", []) or []),
     }
-    app.include_router(_build_a2a_router(runner=_PlaceholderA2ARunner(), settings=_a2a_settings_dict))
+    _a2a_supervisor_cfg = getattr(settings.a2a, "supervisor_config_path", "") or ""
+    if _a2a_supervisor_cfg:
+        _a2a_runner: Any = _SupervisorA2ARunner(app_state, _a2a_supervisor_cfg)
+        logger.info(f"A2A inbound: routing requests through supervisor at {_a2a_supervisor_cfg}")
+    else:
+        _a2a_runner = _PlaceholderA2ARunner()
+        logger.warning("A2A inbound enabled but settings.a2a.supervisor_config_path is empty; using placeholder runner.")
+    app.include_router(_build_a2a_router(runner=_a2a_runner, settings=_a2a_settings_dict))
 
 
 @app.get("/health")
--- a/src/cuga/backend/server/main.py
+++ b/src/cuga/backend/server/main.py
@@ -1687,8 +1687,16 @@ if getattr(settings, "a2a", None) and getattr(settings.a2a, "enabled", False):
                     return
                 yield _A2AStreamEvent("final_answer", {"text": answer}, final=True)
             except Exception as exc:  # surface failures across the wire
+                # Don't expose raw exception text (paths, config, runtime
+                # state) to the remote caller. Class name is enough to
+                # triage; the full traceback is captured server-side by
+                # logger.exception.
                 logger.exception("A2A inbound delegation failed")
-                yield _A2AStreamEvent("error", {"text": f"A2A handler error: {exc}"}, final=True)
+                yield _A2AStreamEvent(
+                    "error",
+                    {"text": f"A2A handler error: {type(exc).__name__}"},
+                    final=True,
+                )
 
     class _PlaceholderA2ARunner:
         """Used when ``a2a.supervisor_config_path`` is unset.
--- a/src/cuga/backend/server/main.py
+++ b/src/cuga/backend/server/main.py
@@ -1629,6 +1629,42 @@ app.include_router(manage_routes.router)
 app.include_router(secrets_routes.router)
 
 
+if getattr(settings, "a2a", None) and getattr(settings.a2a, "enabled", False):
+    # The A2A package is only imported when explicitly enabled in settings,
+    # so disabled deployments pay no import-time cost for it.
+    from cuga.backend.server.a2a import build_router as _build_a2a_router  # noqa: E402
+
+    class _PlaceholderA2ARunner:
+        """Minimal runner used until production graph wiring lands.
+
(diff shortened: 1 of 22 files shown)


src/cuga/backend/server/main.py
```diff
@@ -1629,6 +1629,95 @@ async def event_stream(
...
+        "skill_ids": list(getattr(settings.a2a, "skill_ids", []) or []),
+    }
+    _a2a_supervisor_cfg = getattr(settings.a2a, "supervisor_config_path", "") or ""
+    if _a2a_supervisor_cfg:
+        _a2a_runner: Any = _SupervisorA2ARunner(app_state, _a2a_supervisor_cfg)
+        logger.info(f"A2A inbound: routing requests through supervisor at {_a2a_supervisor_cfg}")
+    else:
+        _a2a_runner = _PlaceholderA2ARunner()
+        logger.warning("A2A inbound enabled but settings.a2a.supervisor_config_path is empty; using placeholder runner.")
+    app.include_router(_build_a2a_router(runner=_a2a_runner, settings=_a2a_settings_dict))
```
sami-marreed: In order to avoid bloated main.py, can we move this to a dedicated library or helper?

src/cuga/settings.toml
```diff
@@ -223,3 +223,13 @@ save_on_success = true       # Save trajectory on successful task completion
...
+
+[a2a]
+enabled = false                                              # Master toggle. When false, the A2A router is not mounted and the package is never imported.
+agent_name = "cuga"                                          # The name advertised in the AgentCard.
+agent_description = "CUGA agent exposed over A2A."           # Free-text description for the AgentCard.
+agent_version = "0.0.0"                                      # Version string for the AgentCard.
+agent_url = "http://localhost:8000"                          # Public URL clients should hit. Override per-deployment.
+auth_required = false                                        # When true, the AgentCard advertises a bearer security scheme.
+skill_ids = ["delegate_task"]                                # Skill IDs surfaced in the AgentCard.
+supervisor_config_path = ""                                  # YAML config for the CugaSupervisor that handles A2A inbound. When empty, a placeholder runner is used.
```
sami-marreed, line 235: Does this force these configuration for current process?

laredo: Good question — answering inline so it's tied to this thread.

`supervisor_config_path` only takes effect when **the A2A inbound endpoint is active** (`[a2a].enabled = true`). When that gate is closed (the default), this setting is read but never used — the A2A package isn't even imported, so the value has zero runtime effect on a normal CUGA process.

When `[a2a].enabled = true` and `supervisor_config_path` points at a YAML, here's what happens for the **inbound** A2A flow only:

1. The router mounts at `/a2a` and `/.well-known/agent[-card].json`.
2. On the **first** `POST /a2a` request, `SupervisorA2ARunner` lazily calls `CugaSupervisor.from_yaml(<this path>)` and caches the supervisor on `app_state.a2a_supervisor` (under an `asyncio.Lock` so concurrent first-requests don't double-build).
3. Subsequent requests reuse the cached supervisor; each `message/send` invokes it with `thread_id=context_id` and emits a single terminal event.
4. **No mutation of global config or env vars happens.** The path is read once, used to construct the dedicated supervisor, and never re-read.

So no, this does not "force" any configuration on the rest of the process — the chat backend on the same port (and its own `[supervisor]` config) is untouched. It's a per-feature opt-in: think of it as "if A2A is on, *which* supervisor handles inbound A2A traffic."

If `[a2a].enabled = true` but this path is left empty, the runner falls back to a placeholder that returns a clear "endpoint reached but unconfigured" terminal event — never an HTTP 5xx — so misconfiguration is loud rather than silent.

The lazy-build behavior was deliberate (no startup cost when A2A is enabled but never hit). After commit `8cd7e9e` all of this lives in `src/cuga/backend/server/a2a/runner.py` rather than `main.py`.

src/cuga/backend/server/main.py
laredo: Done in commit `8cd7e9e`. The 100-line block is now extracted to `src/cuga/backend/server/a2a/runner.py` (`A2AStreamEvent`, `SupervisorA2ARunner`, `PlaceholderA2ARunner`, plus a `build_a2a_router_for_settings(settings.a2a, app_state)` helper). main.py drops to a 5-line opt-in mount under the same `settings.a2a.enabled` gate, so disabled deployments still pay zero import cost.

laredo:
## Review feedback addressed

Pushed two commits (`7bb1487`, `8cd7e9e`) on top of the existing branch. Per-thread replies are inline; this is the rollup.

### Behavior fixes (commit `7bb1487`)

| Comment | What I did |
|---------|------------|
| `task_adapter.py`: terminal error events being marked as `completed` (CodeRabbit) | New `_is_error()` helper; terminal `error`/`failed`/`failure`/`exception` events now map to `TaskState.failed`. + 2 unit tests. |
| `a2a_protocol.py`: outbound flattens peer task failures into `status="success"` (CodeRabbit) | Read peer's `status.state`, return `status="failed"` on `fail*`/`error*`, raise on missing Task envelope, forward `metadata.variables`. + 2 integration tests. |
| `router.py`: SSE generator lacks try/except (CodeRabbit) | Wrapped runner/adapter in try/except; emits one final JSON-RPC error frame on failure. + 1 integration test. |
| `router.py` + `main.py`: don't expose raw exception text (CodeRabbit) | Wire response now carries `"Internal error: <ExceptionClassName>"` only; full traceback in `logger.exception`. + 1 integration test asserting `/secret/path/leak` doesn't reach the client. |
| `_agent_card_rpc_url`: don't accept non-JSONRPC fallback URLs (CodeRabbit, major) | Fallback now requires an HTTP(S) scheme — `grpc://` URLs are skipped so the JSON-RPC sender never POSTs to a wire it can't speak. |

### Refactor + nits (commit `8cd7e9e`)

| Comment | What I did |
|---------|------------|
| **@sami-marreed**: "avoid bloated main.py, can we move this to a dedicated library or helper?" | Extracted ~100 lines of runner glue into `src/cuga/backend/server/a2a/runner.py`. `main.py` is now a 5-line opt-in mount delegating to `build_a2a_router_for_settings(settings.a2a, app_state)`. Replied on the original thread. |
| **@sami-marreed**: "Does this force these configuration for current process?" (settings.toml) | Replied inline explaining `supervisor_config_path` is read once on first `POST /a2a`, scoped only to the inbound A2A flow, never mutates global config or env vars. |
| `_preflight.py`: bare `except Exception: continue` (CodeRabbit, ruff S112/BLE001) | Now `except json.JSONDecodeError` + logs the offending `rowid`. |
| `README.md`: ASCII-diagram fence missing language tag (CodeRabbit, MD040) | Added `text` language tag. |
| `test_outbound_protocol.py`: mutable list class attrs (CodeRabbit, ruff RUF012) | Switched to tuples. |

### Already addressed before this round

- `main.py` ruff format (CodeRabbit) — fixed in `7be1dca`.
- CodeQL "Incomplete URL substring sanitization" — autofix commit `5bdac5f`.
- "Use exact URL assertion" nit — same commit `5bdac5f` introduces a stronger `urlparse(...).scheme/.hostname` check, supersedes the suggested `==`.

### Test status

`uv run pytest tests/unit/a2a/ tests/integration/a2a/` → **46 passed, 1 xfailed** (was 40). Lint + format clean.

Live demo still round-trips end-to-end against the digital_sales provider.

sami-marreed:
LGTM, one final check do you wanna include your tests in CI? you should add them to run_tests.sh script

sami-marreed (approved):

PR #329, opened 2026-06-15
feat: add ToolGuard runtime policy enforcement for tool calls

### LINKED ISSUE #181: [Feature] Enhanced ToolGuard Runtime Observability and Debugging
## What you want and why

The current ToolGuard system provides robust policy enforcement for tool calls, but lacks comprehensive observability and debugging capabilities. When policies fail or behave unexpectedly, developers have limited visibility into:

- Which specific policy rules triggered violations
- The execution flow through multiple guard policies
- Performance metrics for guard evaluation
- Detailed validation context and intermediate states

Based on the E2E debug script (`debug_tool_guard_e2e.py`) and runtime implementation, there's a need for enhanced debugging and monitoring features to help developers:

1. **Understand policy behavior**: See exactly which policies are being evaluated and why they pass/fail
2. **Debug policy conflicts**: Identify when multiple policies interact in unexpected ways
3. **Monitor performance**: Track guard execution time and identify bottlenecks
4. **Audit compliance**: Maintain detailed logs of policy enforcement decisions

### 1. Structured Logging and Tracing

Add structured logging throughout the ToolGuard runtime with configurable verbosity levels:

```python
class ToolGuardRuntime:
    def __init__(self, ..., debug_mode: bool = False, trace_handler: Optional[TraceHandler] = None):
        self.debug_mode = debug_mode
        self.trace_handler = trace_handler or DefaultTraceHandler()
```

### 2. Guard Execution Context

Capture detailed execution context for each guard evaluation:

```python
@dataclass
class GuardExecutionTrace:
    policy_id: str
    policy_name: str
    tool_name: str
    arguments: Dict[str, Any]
    start_time: float
    end_time: float
    result: Literal["ALLOWED", "BLOCKED"]
    violation_message: Optional[str]
    intermediate_checks: List[CheckResult]
```

### 3. Debug Dashboard/Visualization

Provide utilities to visualize guard execution:

- Timeline view of guard evaluations
- Policy dependency graph
- Violation statistics and patterns
- Performance profiling data

### 4. Testing Utilities

Enhance the debug script capabilities:

```python
class ToolGuardTester:
    async def run_test_suite(self, test_cases: List[TestCase]) -> TestReport
    async def profile_guards(self, tool_name: str) -> PerformanceReport
    async def validate_policy_coverage(self) -> CoverageReport
```

### 5. Runtime Metrics

Add metrics collection for monitoring in production:

- Guard evaluation count per tool
- Average/p95/p99 latency
- Violation rate by policy
- Cache hit rates for guard compilation

## Links or extra context

**Related Files:**
- `src/cuga/backend/cuga_graph/policy/tool_guard/tool_guard_runtime.py` - Core runtime implementation
- `src/cuga/backend/cuga_graph/policy/tool_guard/manager.py` - Guard generation and management
- `src/cuga/backend/cuga_graph/policy/tool_guard/tool_invoker.py` - Tool invocation bridge
- `src/cuga/sdk_core/debug_tool_guard_e2e.py` - E2E testing and debugging script

**Current Debug Capabilities:**
The existing `debug_tool_guard_e2e.py` provides basic test case execution but lacks:
- Detailed trace output for each guard evaluation
- Performance profiling
- Interactive debugging capabilities
- Export/reporting functionality

**Benefits:**
- Faster policy development and debugging
- Better production monitoring and alerting
- Improved policy testing and validation
- Enhanced compliance auditing capabilities
- Easier troubleshooting of policy conflicts

### LINKED ISSUE #266: [Feature] Integrate Tool Guard runtime policy enforcement
Parent epic: #75

## Summary

Integrate [Tool Guard](https://pypi.org/project/toolguard/) runtime enforcement into CUGA so tool-guide policies can validate tool calls at execution time, with SDK helpers to generate guard examples/code and persist guards to the policy filesystem.

## Context

Implementation is in progress on:
- PR #176 — core Tool Guard integration (runtime validation, policy persistence, `toolguard` dependency)
- PR #256 — SDK + registry/provider follow-on (stacked; merge after #176)

Follow-up observability work is tracked separately in #181.

**Merge order:** land #250 (agent-core refactor) first, then rebase and merge #176 → #256.

## Scope

- Runtime tool-call validation against tool-guide / tool-guard policies
- SDK methods to generate guard examples, enforcement code, and update guards
- Policy filesystem sync and admin-gated policy save where applicable
- Tests covering guard enforcement paths

## Out of scope

- Enhanced runtime observability → #181
- Broader policy v1 types (intent guard, playbook, approval) → #100

## Acceptance criteria

- [ ] #176 merged to `main` with CI green
- [ ] #256 merged after #176
- [ ] Tool-guide policy can enforce a guard at runtime without monkey-patching
- [ ] Documented SDK entry points for guard generation/update

offerakrabi:
## Feature

Closes #266

### Summary

This PR introduces **ToolGuard**, a runtime policy-enforcement layer for tool calls in Cuga. ToolGuard enables policy-driven control over tool execution by allowing ToolGuide policies to define guards that can allow or block tool calls before execution.

**Key additions:**
- **ToolGuardingToolProvider**: A provider-level decorator that wraps existing tool providers (DirectLangChain, Combined, Registry, etc.) to enforce policies at a single enforcement seam
- **ToolGuard Runtime**: Core enforcement engine that evaluates tool calls against policy guards using LLM-based decision making
- **ToolGuard Manager**: Manages policy loading, matching, and guard execution
- **Tool Invoker**: Handles safe tool invocation with proper error handling and result formatting
- **SDK Integration**: Extended Cuga SDK with ToolGuard support and policy management APIs

**Architecture:**
```
CugaAgent / DynamicAgentGraph
  └── ToolGuardingToolProvider
        └── DirectLangChainToolsProvider / CombinedToolProvider / custom provider
```

**Changes include:**
- 3,764 lines added across 22 files
- New ToolGuard module with comprehensive README documentation
- Unit tests for async tool provider construction, policy loading, and ToolGuard provider
- Integration with existing policy system and filesystem sync
- Support for violating/compliance examples and policy code in ToolGuide definitions

### Testing
- [x] Added unit tests for ToolGuard provider
- [x] Added unit tests for policy loading
- [x] Added unit tests for async tool provider construction

## Release Notes

* **New Features**
  * Added runtime tool-call enforcement driven by Tool Guard policies, including guarded tool wrapping and caching.
  * Extended Tool Guide policies with per-tool guard configuration and added SDK helpers to generate compliance examples and executable guard code.
* **Bug Fixes**
  * Improved synchronous invocation behavior when guarding is enabled.
  * Hardened policy frontmatter generation and updated policy UI payload to include tool guard fields.
* **Documentation**
  * Added a dedicated Tool Guard integration guide.
* **Tests**
  * Expanded Tool Guard provider/runtime unit tests and added a new test phase.
* **Chores**
  * Added `toolguard` to project dependencies.


sami-marreed:
@coderabbitai Can you validate how tool guard is applied for the SDK if it's using only the combined tool provider internally? What if it gives tools directly? Is that possible? Will it work?

sami-marreed:
@coderabbitai does this replace tool guide?

src/cuga/backend/cuga_graph/policy/tool_guard/tool_guard_runtime.py
```diff
@@ -0,0 +1,951 @@
...
+
+            # Validate that policy_code contains at least one async def guard_* function
+            guard_func_name = self._extract_guard_function_name(tool_guard.policy_code)
+            if not guard_func_name:
+                logger.error(
+                    f"Tool guard for '{tool_name}' in policy '{policy.name}' "
+                    f"has policy_code but no valid 'async def guard_*' function found. "
+                    f"Skipping registration to prevent marking tool as guarded without enforcement."
+                )
+                continue
```
sami-marreed, line 208: I see the existing comment frames this skip as intentional ("prevent marking tool as guarded without enforcement"), but the *consequence* is fail-open: skipping registration means `tool_name` never enters `tool_to_guards`, so `guard_tool_call` hits the `function_name not in self.tool_to_guards` branch and returns `None` (allow). So a guard whose generated `policy_code` lacks a valid `async def guard_*` ends up running the tool **unguarded** — the admin who attached it gets zero enforcement and only a log line.

This is inconsistent with the fail-closed posture elsewhere in `guard_tool_call` (runtime-unavailable → block; internal guard error → block). Suggest failing closed here too: track tools that have a *declared-but-invalid* guard and block them in `guard_tool_call`, or raise at policy-load time so the misconfiguration surfaces loudly instead of degrading to no enforcement.

src/cuga/backend/cuga_graph/nodes/cuga_lite/providers/toolguard.py
```diff
@@ -0,0 +1,405 @@
...
+                    arguments=all_kwargs,
+                )
+                if error:
+                    return {
+                        "error": f"Tool call blocked by policy: {error}",
+                        "blocked_by_policy": True,
+                        "policy_violation": True,
+                        "tool": tool_name,
+                        "app": app_name,
+                    }
```
sami-marreed, line 277: A blocked call returns this error-shaped dict into the sandbox `_locals` as the tool's normal return value. Nothing forces the agent to treat it as terminal — the LLM can ignore `blocked_by_policy` and retry with different args or just proceed. For an enforcement layer that's "discouragement," not a block.

Proposal/question: should we add a per-guard **policy action**, e.g. `block` (return error, agent may continue or try another path) vs `exit`/`terminate` (raise a dedicated terminal exception the executor surfaces as a hard stop)? For security guards the terminal action feels like the right default. At minimum we should have a test proving a block actually halts the workflow rather than just returning an error-shaped result. WDYT?

Recommendation from me: Check policy enactment with action block

src/cuga/backend/cuga_graph/policy/models.py
```diff
@@ -165,6 +165,26 @@ class IntentGuardResponse(BaseModel):
...
+    compliance_examples: List[str] = Field(
+        default_factory=list, description="Examples of compliant usage patterns"
+    )
+    policy_code: str = Field(
+        default="",
+        description=(
+            "Python code that validates tool usage compliance. "
+            "This code is executed in a sandboxed environment using the toolguard library. "
+            "Only trusted administrators with manage access should be allowed to modify policy code. "
+            "While sandboxed, policy code should still be reviewed for correctness and performance."
```
sami-marreed: This says policy_code "is executed in a sandboxed environment using the toolguard library," but the toolguard runtime does `exec(file_twin.content, module.__dict__)` in-process — no sandbox, no restricted builtins. So `policy_code` is arbitrary Python running with full process privileges.

Two parts:
1. Correct the "sandboxed" wording here (and the README security section) to state plainly that policy_code runs in-process with full privileges.
2. Open question — as we move tool execution into the sandbox, should `policy_code` execution move there too? The guard needs the tool args + runtime context, so the sandbox would have to expose those. By the trust model policy_code is admin-authored, so this is defense-in-depth rather than strictly required — but if we're sandboxing broadly anyway, running guards in the same sandbox is the consistent choice and shrinks blast radius. How do we want to scope this?

src/cuga/backend/cuga_graph/policy/tool_guard/tool_guard_runtime.py
```diff
@@ -0,0 +1,951 @@
...
+        except Exception as e:
+            logger.error(
+                f"Error executing umbrella guard for tool '{function_name}' in app '{app_name}': {e}",
+                exc_info=True,
+            )
+            # Fail closed: treat internal guard errors as a violation so a buggy
+            # or malformed guard cannot silently bypass policy enforcement.
+            return (
+                f"Internal guard error for '{function_name}': {e}. Tool call blocked as a safety precaution."
+            )
```
sami-marreed, line 867: These are the security-critical branches — internal guard error → block here, and runtime-unavailable-but-guard-exists → block just above (~860-868) — but every test uses `FakeRuntime` or `policy_storage=None`, so none of them are actually exercised. Please add tests asserting: (a) guard raises an exception → call blocked; (b) runtime unavailable while an applicable guard exists → blocked; (c) the invalid-`policy_code` registration case from the other thread. Right now the fail-closed guarantees are asserted only by code comments.

src/cuga/backend/cuga_graph/policy/tool_guard/tool_guard_runtime.py
```diff
@@ -0,0 +1,951 @@
...
+        )
+
+        try:
+            # Type-cast arguments to their proper types before validation
+            typed_arguments = self._type_cast_arguments(arguments, app_name, function_name)
+
+            args_obj = SimpleNamespace(**typed_arguments)
+            await runtime.guard_toolcall(
+                tool_name=function_name,
+                args=typed_arguments | {"args": args_obj},
```
sami-marreed, line 851: `args=typed_arguments | {"args": args_obj}` will silently overwrite a tool parameter legitimately named `args` with the `SimpleNamespace`. Edge case, but it corrupts the arguments the guard sees for any tool that actually has an `args` parameter. Suggest detecting/logging the collision, or namespacing the injected object (e.g. `_args`) to avoid clobbering real arguments.

src/cuga/backend/cuga_graph/policy/tool_guard/tool_guard_runtime.py
```diff
@@ -0,0 +1,951 @@
...
+
+            logger.info(f"✅ Runtime initialized for app '{app_name}'")
+            return runtime
+        except Exception as e:
+            logger.error(f"Failed to initialize runtime for app '{app_name}': {e}", exc_info=True)
+            # Cache None to avoid repeated failed attempts
+            self._runtimes_by_app[app_name] = None
+            return None
+
+    def _type_cast_arguments(
```
sami-marreed: `_type_cast_arguments` reconstructs the `*Args` class name from `function_name` (capitalize + join) and regex-scans the generated types file to recover types. This is brittle: tool names that don't map cleanly to the generated class name silently skip casting, and the guard then compares against raw strings (e.g. `revenue > 1000000` evaluated on a `str`). Prefer reading types from the tool's existing pydantic `args_schema` (already attached to the tool) instead of regex over source.

src/cuga/backend/cuga_graph/policy/tool_guard/tool_guard_runtime.py
```diff
@@ -0,0 +1,951 @@
...
+        if function_name not in self.tool_to_guards:
+            logger.debug(f"No guards registered for tool '{function_name}'")
+            return None
+
+        # Filter guards to only those applicable to this app
+        all_guards = self.tool_to_guards[function_name]
+        guards = [
+            guard
+            for guard in all_guards
+            if guard.target_apps is None or not guard.target_apps or app_name in guard.target_apps
```
sami-marreed, line 804: [ non critical ]

When a tool has registered guards but `target_apps` is set to a list that doesn't include the invoking app, this filter drops every guard and `guard_tool_call` returns `None` (allow) with only a `debug`-level log. For directly-provided tools the runtime app is `"runtime_tools"` (see `DirectLangChainToolsProvider(app_name="runtime_tools")` in sdk.py and `langchain.py`), so a policy whose `target_apps` omits `"runtime_tools"` silently runs the tool unguarded even though an admin clearly intended to guard it. Two asks:

1. Make this a `WARNING` (not `debug`) — a tool that has guards registered but all filtered out by app mismatch is almost always a misconfiguration that silently disables enforcement, and it should be loud.
2. Document (README + the `target_apps` field docstring) that directly-provided tools enforce under the app name `"runtime_tools"`, so any `target_apps` must include it. Note: the generate-time `app_name` auto-detection is fine — it's derived from the same provider index and raises on ambiguity — so this is specifically about the runtime `target_apps` filter, not guard-code generation.

sami-marreed (commented):
## ToolGuard review — summary

The provider-seam design is genuinely clean: `ToolGuardingToolProvider` is a transparent decorator, it's backward-compatible when disabled, and the runtime core is intentionally fail-closed (runtime-unavailable and internal-guard-error both block). Enforcement is also deterministic — the LLM is only used at build time, not per call. Nice work.

Below are the points worth resolving before merge. Two I'd treat as blocking; the rest are important or open questions.

**Critical**
- [ ] 1. Misconfigured guard (valid `policy_code` but no `async def guard_*`) is skipped → tool runs **unguarded** (fail-open). Should fail closed.
- [ ] 2. Blocked call returns an error-shaped dict the agent can ignore/retry — propose a per-guard `block` vs `exit/terminate` action. **Needs your decision.**

**Important**
- [ ] 3. `policy_code` docstring says "sandboxed" but it's `exec`'d in-process with full privileges — fix wording; open question on whether to move it into the sandbox.
- [ ] 4. Fail-closed branches (guard-raises, runtime-unavailable, invalid-policy_code) are untested.
- [ ] 5. `args` argument-name collision overwrites a real `args` tool param.
- [ ] 6. `_type_cast_arguments` regex-parses generated source — prefer the pydantic `args_schema`.
- [ ] 7. `target_apps` mismatch silently disables enforcement (allow + debug-only log); document `"runtime_tools"` as the enforcement app for directly-provided tools.

**Out of scope this round:** the broader full-API (non-Lite) execution-path coverage concern — noted but not blocking here.

---

**@offerakrabi — could you reply in a summary comment addressing each item above?** For each, please note whether it's **handled / won't-fix / deferred**, a one-line rationale, and the **commit SHA** that addresses it (or where you'd push back). The two `[ ]` Critical items and the two open questions (#2 block-action, #3 sandbox scope) are the ones I'd most like an explicit answer on before approving. Thanks!

offerakrabi:
@sami-marreed 
1. **handled** — Invalid `policy_code` now fails closed.  
   Added invalid-guard tracking so a declared guard without a valid `async def guard_*` blocks matching tool calls instead of silently allowing them.

2. **won’t-fix** — ToolGuard blocks the tool call, not the whole agent run.  
   This is the intended contract: the protected tool is not invoked, the agent receives a policy-violation result, and any retry will be blocked again.

3. **handled** — Corrected `policy_code` sandbox/security wording.  
   Updated docs to clarify that guard code currently runs in the ToolGuard/backend execution context with service-process privileges, not inside the agent code sandbox.

4. **handled** — Added fail-closed runtime tests.  
   Covered internal guard exceptions, runtime-unavailable-with-applicable-guard, and invalid declared `policy_code`.

5. **handled** — Prevented silent `args` argument clobbering.  
   Runtime now detects a real tool parameter named `args` and blocks with a clear internal collision error instead of overwriting it.

6. **handled** — Replaced brittle generated-source type casting.  
   Argument casting now uses the actual tool `args_schema` / Pydantic validation instead of reconstructing class names and regex-parsing generated type files.

7. **handled** — Made `target_apps` mismatches louder and documented `runtime_tools`.  
   Runtime now warns when all registered guards are filtered out by app mismatch, and docs/model text explain that direct LangChain/runtime tools use app name `"runtime_tools"`.

sami-marreed (approved):

Issue #357, opened 2026-06-23
[Bug]: TraceScopedLangfuseCallbackHandler override never executes (Python name mangling)


haroldship:
## Problem

`TraceScopedLangfuseCallbackHandler` in `src/cuga/backend/cuga_graph/utils/langfuse_tracing.py` (introduced in `77af4cfd`, part of the `fix/langfuse-nested-callback-propagation` follow-up work for #276/#288) was meant to attach orphan LLM calls (no tracked parent run) to the current eval trace id.

**This is currently live in `main`** — PR #277 stayed open and accumulated this follow-up work before being squash-merged as `b1373fc9`, so the bug shipped to `main`, not just an unreleased branch.

It subclassed `langfuse.langchain.CallbackHandler` and defined:

```python
class TraceScopedLangfuseCallbackHandler(base_cls):
    def __on_llm_action(self, serialized, run_id, prompts, parent_run_id=None, ...):
        ...
        super().__on_llm_action(...)
```

This is dead code. `__on_llm_action` is a double-underscore name, so Python mangles it to the *defining class's* name. The base class's `on_llm_start`/`on_chat_model_start` call `self.__on_llm_action(...)`, which is fixed at the base class's compile time to `self._LangchainCallbackHandler__on_llm_action(...)`. The subclass's method mangles to `_TraceScopedLangfuseCallbackHandler__on_llm_action`, a different attribute — so the override is never invoked. Worse, if it were ever invoked directly, its own `super().__on_llm_action(...)` call would mangle to a name the parent class doesn't have, raising `AttributeError`.

Net effect: the "attach orphan LLM runs to the eval trace" behavior promised by this class silently does nothing — it behaves identically to the unmodified base handler.

No existing test exercised `on_llm_start`/`on_chat_model_start` on this class, so it shipped unnoticed.

## Repro

```python
class Base:
    def public_caller(self):
        self.__action(1)
    def __action(self, x):
        print("Base.__action", x)

class Sub(Base):
    def __action(self, x):
        print("Sub.__action (override attempt)", x)
        super().__action(x)

Sub().public_caller()
# prints only "Base.__action 1" — the override never runs
```

## Fix

Override the public entry points (`on_llm_start`, `on_chat_model_start`) directly instead of the private dunder, swapping `_trace_context` around the `super()` call. Add regression tests using a fake base class that mirrors the real handler's private-dispatch shape, asserting the swap actually takes effect for orphan calls.

See #358.

PR #277, opened 2026-06-02
fix(tracing): propagate Langfuse callbacks to nested LLM calls

haroldship:
## Bug fix

Fixes #276

### Summary

Nested LLM paths (reflection, find_tools shortlister, NL auto-continue, output formatter, context summarization) did not inherit Langfuse `CallbackHandler` config from the parent graph invoke, so one logical `CugaAgent.invoke()` could produce multiple root traces in Langfuse. This change adds `langfuse_tracing.py` (contextvar + `get_langfuse_invoke_config()`), wires those call sites, and deduplicates agent-level vs per-invoke Langfuse handlers in `sdk._apply_callbacks`.

Related eval harness work: [cuga-eval#28](https://github.com/cuga-project/cuga-eval/issues/28).

### Testing

- [x] Verified fix locally; automated acceptance tests below pass (manual Langfuse UI check omitted — screenshots to follow on the issue).

### Prerequisites

```bash
source .venv/bin/activate
pip install -e ".[dev]"
```

### (a) Problem — `main`

```bash
git fetch origin
git checkout main
git pull origin main
```

Smoke check:

```bash
grep -n 'config={"callbacks": \[\]}' \
  src/cuga/backend/cuga_graph/nodes/cuga_lite/nl_auto_continue_classifier.py
```

Expected: one line (e.g. `121:            config={"callbacks": []},`) — NL auto-continue passes an empty callback list.

Run:

```bash
cat > /tmp/langfuse_before_check.py <<'PY'
import asyncio
from unittest.mock import AsyncMock, MagicMock, patch


async def main():
    from cuga.backend.cuga_graph.nodes.cuga_lite import nl_auto_continue_classifier as mod
    from cuga.config import settings

    llm = MagicMock()
    llm.ainvoke = AsyncMock(return_value=MagicMock(content='{"auto_continue": false}'))

    with patch.object(settings.advanced_features, "cuga_lite_nl_auto_continue", True, create=True):
        await mod.classify_nl_auto_continue(
            llm,
            assistant_visible="done with the task",
            reasoning_excerpt="finished",
        )

    _, kwargs = llm.ainvoke.call_args
    callbacks = (kwargs.get("config") or {}).get("callbacks") or []
    if callbacks:
        raise SystemExit("UNEXPECTED: callbacks present on main")
    print("OK (before): nested ainvoke has no Langfuse callbacks")


asyncio.run(main())
PY
python /tmp/langfuse_before_check.py
```

[...]

Issue #276, opened 2026-06-02
[Bug]: Langfuse: nested LLM calls create separate root traces (callback propagation)


haroldship:
## Problem

When `advanced_features.langfuse_tracing=true`, a single `CugaAgent.invoke()` run can produce **multiple root traces** in Langfuse instead of one tree. Generations from nested LLM calls appear as siblings or separate roots, not children of the main run.

This affects **standalone cuga-agent** (server, SDK, policy output formatting) and is amplified by eval harnesses (see [cuga-eval#28](https://github.com/cuga-project/cuga-eval/issues/28)).

### Root cause

LangGraph passes Langfuse `CallbackHandler` via `config["callbacks"]` for `call_model`, but nested paths call `ainvoke` without that config:

| Nested path | Module |
|-------------|--------|
| Reflection | `sandbox_node.py` |
| find_tools shortlister | `prompt_utils.py` |
| NL auto-continue | `nl_auto_continue_classifier.py` (on `main`: `config={"callbacks": []}`) |
| OutputFormatter | `enactment.py` |
| Context summarization | `context_management_utils.py` |

### Fix branch (ready for PR)

**Branch:** `fix/langfuse-nested-callback-propagation`  
**Commit:** `5591085a`

- `langfuse_tracing.py` — contextvar + config merge + `get_langfuse_invoke_config()`
- Wire all nested paths; SDK dedupes agent-level Langfuse when `langfuse_trace_id` is set
- Tests: `tests/unit/test_langfuse_tracing.py`

## Test evidence (no Langfuse keys)

**Before (`main`):** no regression module; NL auto-continue uses `config={"callbacks": []}`.

**After (fix branch):**
```bash
.venv/bin/python -m pytest tests/unit/test_langfuse_tracing.py -v
# 10 passed
```

## Manual verification (attach screenshots)

1. Langfuse trace **list** before: multiple root traces for one `invoke`
2. Langfuse trace **tree** after: one root, nested GENERATION nodes
3. (Optional) JSON: `parentObservationId` set on nested observations

**Repro:** `DYNACONF_ADVANCED_FEATURES__LANGFUSE_TRACING=true`, reflection + shortlist enabled, one short `CugaAgent.invoke()`.

## Acceptance criteria

- [ ] One trace per `invoke` when trace-scoped callbacks are used
- [ ] Nested paths propagate callbacks
- [ ] Unit tests pass in CI
- [ ] SDK-only `callbacks=[...]` on construction still works

[...]

PR #358, opened 2026-06-23
fix(tracing): make TraceScopedLangfuseCallbackHandler override reachable

haroldship:
## Summary

While preparing a follow-up PR for #276/#288 (already merged via #277), I found that `TraceScopedLangfuseCallbackHandler` in `langfuse_tracing.py` has been silently broken since it was introduced — it currently ships in `main`. CodeRabbit's review of this PR additionally surfaced two more real bugs in the same already-merged code (#359), fixed here too.

### 1. `TraceScopedLangfuseCallbackHandler` override never runs (#357)

It subclasses `langfuse.langchain.CallbackHandler` and was meant to attach orphan LLM calls (no tracked parent run) to the eval trace id, via a `__on_llm_action` override. `__on_llm_action` is a double-underscore name, so Python mangles it to the *defining class's* name — the base class's `on_llm_start`/`on_chat_model_start` call `self.__on_llm_action(...)`, fixed at the base class's compile time to `self._LangchainCallbackHandler__on_llm_action(...)`. The subclass's method mangles to a different attribute, so the override never runs; the class behaves identically to the unmodified base handler.

**Fix:** override the public `on_llm_start`/`on_chat_model_start` entry points directly, swapping `_trace_context` around the `super()` call.

### 2. Stale trace state persists across invocations (#359)

`sync_langfuse_callbacks_from_config` only updated `_langfuse_trace_id`/`_langfuse_primary_handler` when the new config had a truthy trace id, leaving old state in place for a later invocation in the same async context with no trace id (e.g. an eval-harness worker reusing one asyncio task) — nested LLM calls could reattach to the stale previous trace.

**Fix:** always set (or explicitly clear) the trace id and primary handler.

### 3. Callback managers not flattened before Langfuse dedup (#359)

`CugaAgent._apply_callbacks`'s `per_call_langfuse` check used `_as_list`, which wraps a non-list `callbacks` value (e.g. a LangChain callback manager) in a single-item list without flattening it — a Langfuse handler nested inside a manager was never detected, so the agent-level Langfuse handler wasn't dropped, duplicating root traces. Same bug class as #276.

**Fix:** use `collect_langfuse_callbacks_from_config`, which already flattens via `_flatten_callbacks`.

## Test plan

- [x] Added regression tests for all three fixes, each verified to fail before the fix and pass after
- [x] `pytest tests/unit/test_langfuse_tracing.py tests/unit/test_cuga_lite_bind_tools.py` — 45 passed
- [x] `ruff check` / `ruff format` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved trace context handling for nested AI operations to ensure accurate tracing across auto-continue classification, output formatting, tool selection, and context summarization
  * Enhanced callback synchronization and trace ID propagation
  * Fixed orphaned trace event attachment in complex scenarios

* **Tests**
  * Added comprehensive unit test coverage for Langfuse tracing and callback propagation in nested AI invocations

### DIFF
--- a/src/cuga/backend/cuga_graph/utils/langfuse_tracing.py
+++ b/src/cuga/backend/cuga_graph/utils/langfuse_tracing.py
@@ -168,9 +168,8 @@ def sync_langfuse_callbacks_from_config(config: Any = None) -> None:
         return
     _langgraph_run_config.set(config)
     configurable = config.get("configurable") or {}
-    trace_id = configurable.get("langfuse_trace_id")
-    if trace_id:
-        set_langfuse_trace_id(trace_id)
+    trace_id = configurable.get("langfuse_trace_id") or None
+    set_langfuse_trace_id(trace_id)
     callbacks = collect_langfuse_callbacks_from_config(config)
     if callbacks:
         _remember_primary_handler(callbacks)
@@ -179,6 +178,8 @@ def sync_langfuse_callbacks_from_config(config: Any = None) -> None:
         if handler:
             _langfuse_primary_handler.set(handler)
             callbacks = [handler]
+    else:
+        _langfuse_primary_handler.set(None)
     set_langfuse_callbacks(callbacks or None)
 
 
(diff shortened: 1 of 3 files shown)


haroldship:
@coderabbit resolve

sami-marreed (approved):

PR #361, opened 2026-06-24
feat(a2a): human-in-the-loop approval round-trip for the simple runner

laredo:
### Summary

Adds human-in-the-loop (HITL) approval support to the A2A **simple runner**, so a single CUGA agent exposed over A2A (no supervisor config) correctly handles approval interrupts and returns real answers. Builds on the merged A2A server (#323).

- **Fixes empty responses.** The simple runner now parses CUGA's real SSE wire format (`event:`/`data:`) instead of a fictional JSON envelope, so answers and errors are actually detected — previously every request fell through to `"Agent completed processing"`.
- **Answers reach the caller.** The router now populates the task's terminal `status.message` **and** an answer `artifact`. A2A clients read those, not `history`, so callers were getting empty results before.
- **HITL round-trip** (`settings.a2a.auto_approve`, default `false`):
  - `true` → interrupts are auto-confirmed with a real `ActionResponse(confirmed=true)` and the graph resumes to completion.
  - `false` → the interrupt is surfaced as a terminal `input_required` event carrying an actionable prompt + `action_id` (in message metadata). The caller's follow-up message on the same `contextId` resumes the parked graph.
  - Decision parsing: structured `{action_id, confirmed}` (DataPart or message metadata) preferred, plain-text `approve`/`deny` fallback, ambiguous replies re-ask. Tolerates callers (e.g. watsonx Orchestrate) that echo the whole conversation transcript back as one message.
- HITL interrupts map to a terminal `input_required` task state (previously they produced a spurious trailing `completed`).

Verified end-to-end against both a CUGA peer and watsonx Orchestrate (auto-approve and surface-then-approve flows).

### Testing
- [x] Tested locally; tests pass — `44 passed, 1 xfailed` (`tests/integration/a2a/`, `tests/unit/test_a2a_simple_runner_hitl.py`)

## Release Notes

* **New Features**
  * Added human-in-the-loop approval support to `message/send` and `message/stream`, including configurable auto-approval.
  * Introduced a lightweight “simple runner” for streaming responses directly from CUGA events (with interrupt/HITL handling).
  * When runs complete, the system can include richer task details and generate an answer artifact.
* **Bug Fixes**
  * Improved HITL handling so terminal HITL updates are treated as final.
* **Tests**
  * Added/updated integration and unit coverage for routing, streaming, approval round-trips, interrupt behavior, and error handling.


src/scripts/run_tests.sh
```diff
@@ -93,6 +93,9 @@ echo "Running ToolGuard provider tests..."
 run_pytest tests/unit/test_toolguard_provider.py \
     tests/unit/test_policy_tool_guide_loading.py \
     tests/unit/test_async_tool_provider_construction.py
+echo "Running A2A tests..."
+run_pytest tests/integration/a2a/ \
+    tests/unit/test_a2a_simple_runner_hitl.py
```
sami-marreed, line 98: Can you do wild card on tests? or you did choose only this one on purpose

sami-marreed (approved):

Issue #272, opened 2026-05-28
[Design] CugaLite struggles with weak/missing tool output schemas and state-observation tools


haroldship:
## Problem

CugaLite’s **CodeAct loop** assumes the model can plan over tool results using schemas and execution semantics shown in the system prompt. In practice, many registry/MCP tools—especially **state-observation** tools (browser/desktop snapshots, large unstructured blobs)—have **missing, vague, or mismatched output schemas** (`str`, free-form `dict`, `Any`, or no `responseSchema` at all).

The agent then:

- Mis-parses or over-shapes results in generated Python
- Chains multiple logical steps in one turn when intermediate observation should be visible (combined `extract_and_combine_codeblocks` behavior)
- Struggles to decide what to do next on observation-heavy tools (e.g. file reads, browser state snapshots)

This is a **reliability and evaluation** problem first—not “add native function calling” as the headline.

## Failure modes to evaluate (with examples)

| Category | Example tools / cases | What to measure |
|----------|----------------------|-----------------|
| **No output schema** | Tools registered without `outputSchema` / `_response_schemas` | Does the model hallucinate structure? Wrong indexing? |
| **Text / string payloads** | Large markdown or plain-text returns | Token blow-up, truncation, lost fields |
| **Dict / Any / loose object** | Nested dicts, dynamic keys | Wrong key paths, unsafe assumptions in code |
| **State observation** | Browser/CUA/desktop snapshot tools | One-shot vs multi-step; need isolated observation turn? |
| **Multi-block CodeAct** | Model emits 2+ fenced Python blocks | Combined execution hides intermediate tool output |

### Example evaluation tasks (starter set)

1. **Read + slice** — Read a file, then process the first N rows (separate blocks combined vs isolated execution).
2. **Discover → call** — Tool discovery, then a single app tool with a messy return shape.
3. **Observation loop** — State tool → decide action → state tool again (minimum 2–3 turns).
4. **Todo + tool mix** — Plan update in an isolated block, then an app tool with dict return.

Document per task: model, tool set size, pass/fail, failure mode tag, trace link.

## Candidate approaches (not committed—spike/eval order TBD)

1. **Tool guardrails** — Cap size, normalize shape, reject/trim unsafe observation payloads before prompt.
2. **Few-shot examples** — User/assistant/user turns showing correct handling per schema class.
[...]

PR #362, opened 2026-06-24
fix(272): probe weak/missing tool output schemas, isolate probe blocks (Stage 1+2)

haroldship:
## Summary

Implements a staged fix for #272 (CugaLite struggles with weak/missing tool output schemas), combining elements of candidate solutions #2, #3, and #5 from the original design deck (attached on the issue) into the smallest change set that closes the gap.

**Component A — weak-schema detection + per-tool probing directive**
Detects tools with no real declared output schema (both the genuinely-empty OpenAPI case and MCP's synthetic `{"success": {"type": "string"}}` placeholder) and replaces their rendered schema block with a directive telling the model to call the tool alone and `print()` the raw result before writing code that assumes its shape.

**Component C — post-probe session enrichment**
Captures the first observed shape of a weak-schema tool's result per session and surfaces it back into the prompt on every subsequent turn via the existing per-turn `prepare_system_content` hook, so the model stops re-probing a tool it has already seen.

**Component B — structural enforcement of the block split**
A+C alone are advisory: nothing stopped the runner from combining a probe call and a slicing call written in the same completion into one executed script, which is the literal mechanism behind the issue's failure trace. B adds a `CoreGraphAdapter.get_tools_needing_probing()` hook and truncates a multi-block response after the first block that references a still-unobserved weak-schema tool, discarding the rest so the model is forced to see the real result before writing dependent code. CugaSupervisor is unaffected (hook defaults to `frozenset()`).

Full design rationale: `docs/superpowers/specs/2026-06-23-weak-tool-schema-probing-design.md`
Implementation plans: `docs/superpowers/plans/2026-06-23-weak-tool-schema-probing-stage1.md`, `docs/superpowers/plans/2026-06-23-weak-tool-schema-probing-stage2.md`

## Staging

This PR is the union of two branches, kept separate during development so each component's contribution could be evaluated independently:

- `feat/272-tool-output-schema-probing` — Stage 1, Components A+C only
- `feat/272-stage2-block-isolation` (this branch, off Stage 1) — adds Component B

## Eval results (Vakra M3, 1 task, 5 runs each)

| Branch | Commit | Pass@1 | ExactMatch | Avg tokens | Avg LLM calls | Avg time |
|---|---|---|---|---|---|---|
| main | 8d3d3599 | 0.0% | 0.20 | 301,614 | 15.2 | 65.2s |
| feat/272-tool-output-schema-probing (A+C) | 2dc47738 | 0.0% | 0.80 | 205,719 | 11.2 | 25.2s |
| feat/272-stage2-block-isolation (A+B+C) | 4f623786 | 60.0% | 1.00 | 214,005 | 11.0 | 24.7s |

A+C alone already cuts tokens/calls/time roughly in half and raises ExactMatch substantially, but doesn't move Pass@1 — consistent with the design doc's prediction that advisory-only probing wouldn't close the gap on its own. Adding B (structural enforcement) is what converts that into actual task success.

## Test plan

- [x] Stage 1 unit tests (`prompt_utils`, `prepare_node`, `sandbox_node`, `graph_adapter`) — new + extended
- [x] Stage 2 unit tests (`code_extraction`, `graph_nodes`, `shared_nodes`, `graph_adapter`) — new + extended, including an end-to-end reproduction of the issue's literal original multi-block trace
- [x] Full regression: `cuga_lite` + `cuga_agent_core` + `cuga_supervisor` suites green (363 passed, 1 skipped, 3 pre-existing unrelated e2b-package-not-installed failures)
- [x] `ruff check` clean
- [ ] Broader eval run across the full Vakra(M3) task set (this PR's numbers are a single-task, 5-run sample)

### DIFF
--- a/docs/superpowers/plans/2026-06-23-weak-tool-schema-probing-stage1.md
+++ b/docs/superpowers/plans/2026-06-23-weak-tool-schema-probing-stage1.md
@@ -1,0 +1,637 @@
+# Weak Tool Output Schema Probing — Stage 1 (A+C) Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Give CugaLite a per-tool directive that tells the model to probe weak-schema tools in isolation, and surface the real observed output shape back into the prompt for the rest of the session once a probe happens.
+
+**Architecture:** Two independent, additive pieces wired through existing seams — no new abstractions, no shared-loop changes. (A) `PromptUtils.get_tool_docs` renders a probing directive instead of an empty/placeholder schema block when a tool has no real declared output schema. (C) `sandbox_node.py` captures the real shape of a weak-schema tool's first successful result into a per-session dict on the adapter; `AgentGraphAdapter.prepare_system_content` (already called before every `call_model` turn) appends it as a small note once populated.
+
+**Tech Stack:** Python 3.12, pytest, pytest-asyncio, LangGraph, LangChain `StructuredTool`.
+
+## Global Constraints
+
+- Spec: `docs/superpowers/specs/2026-06-23-weak-tool-schema-probing-design.md`. This plan covers Stage 1 (components A and C) only.
+- CugaLite only. Do **not** touch `shared_nodes.py`, `code_extraction.py`, or `graph_nodes.py` (`CoreGraphAdapter` ABC) — those belong to Stage 2 (component B), planned and implemented separately on its own branch.
+- Every new piece of state defaults to empty/falsy. Behavior for tools with a real declared output schema must be byte-identical to today — verify this with an explicit regression test in Task 1.
+- Match existing code style: plain `hasattr`/`getattr` duck-typing (no new Protocols/ABCs), `SimpleNamespace`-based test fixtures (matches `test_agent_graph_adapter.py` and `test_prepare_node_task_todos_reset.py`), no docstring bloat.
+- Run the full suite for any file you touch before committing it: `uv run pytest <test file> -v`.
+
+## File Structure
+
+- Modify: `src/cuga/backend/cuga_graph/nodes/cuga_lite/prompt_utils.py` — weak-schema detection + probing directive text.
+- Modify: `src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/prepare_node.py` — compute the weak-schema tool-name set once per session.
+- Modify: `src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/graph_adapter.py` — new per-session state + per-turn enrichment note.
+- Modify: `src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/sandbox_node.py` — capture the first observed shape per weak-schema tool.
+- Test: `src/cuga/backend/cuga_graph/nodes/cuga_lite/tests/test_prompt_utils_weak_schema.py` (new)
+- Test: `src/cuga/backend/cuga_graph/nodes/cuga_lite/tests/test_prepare_node_weak_schema_tools.py` (new)
+- Test: `src/cuga/backend/cuga_graph/nodes/cuga_lite/tests/test_sandbox_node_weak_schema_shapes.py` (new)
+- Test: `src/cuga/backend/cuga_graph/nodes/cuga_lite/tests/test_agent_graph_adapter.py` (extend existing `prepare_system_content` section)
+
+---
+
+### Task 1: Weak-schema detection + probing directive in `get_tool_docs`
+
+**Files:**
+- Modify: `src/cuga/backend/cuga_graph/nodes/cuga_lite/prompt_utils.py:16-17` (new module constant), `:149-150` (new method), `:163-174` (directive branch)
+- Test: `src/cuga/backend/cuga_graph/nodes/cuga_lite/tests/test_prompt_utils_weak_schema.py`
+
+**Interfaces:**
+- Produces: `PromptUtils.is_weak_schema_tool(tool: StructuredTool) -> bool` (static method) — used by Task 2.
+- Produces: module constant `_WEAK_SCHEMA_PROBE_DIRECTIVE: str` in `prompt_utils.py` (not exported elsewhere).
+
+- [ ] **Step 1: Write the failing tests**
+
+Create `src/cuga/backend/cuga_graph/nodes/cuga_lite/tests/test_prompt_utils_weak_schema.py`:
+
+```python
+"""Weak-schema detection and probing directive in PromptUtils.get_tool_docs (issue #272)."""
+
+from __future__ import annotations
+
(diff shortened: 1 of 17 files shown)


haroldship:
## Broader eval (Vakra M3, 10 tasks across 9 domains, 5 runs each) — and a finding

Ran this PR's branches at wider scope to validate the single-task numbers above. Results, with `pass@5`/`pass^5`/`maj@5`/`Cons` added:

| Branch | Commit | Pass@1 | pass@5 | pass^5 | maj@5 | Cons | Tokens | LLM | Time |
|---|---|---|---|---|---|---|---|---|---|
| main | 8d3d3599 | 16.0% | 30.0% | 10.0% | 10.0% | 1.00 | 3,730,539.6 | 157.0 | 758.1s |
| feat/272-tool-output-schema-probing (A+C) | 2dc47738 | 16.0% | 40.0% | 0.0% | 10.0% | 0.00 | 4,648,273.6 | 179.6 | 845.9s |
| feat/272-stage2-block-isolation (A+B+C) | 4f623786 | 10.0% | 20.0% | 0.0% | 10.0% | 0.00 | 5,048,238.4 | 190.2 | 939.2s |

This contradicts the single-task conclusion above ("Adding B is what converts that into actual task success") — at this scope, `main` actually has the best `pass^5`/`Cons` (it's the only branch with a task that passes all 5/5 runs), and both feature branches cost *more* tokens/time while scoring *worse* on Pass@1 than `main`.

### Root cause: Component B never actually engages

Traced this through the langfuse traces and the code, not just the aggregate numbers. In the `feat/272-stage2-block-isolation` trace for `beer_factory` task `a823e527d383-5ee3999c12b4` (run 1), the model wrote **three** fenced blocks in one turn — a probe call, code assuming the probe's result is a string, and a call to a hallucinated tool name — and **all three executed together** as one script (confirmed: the eventual `NameError` came from block 3, proving blocks 1–2 had already run). If block-isolation were active, that turn should have truncated after block 1.

Root cause: `prepare_node.py:600-603` computes `adapter._weak_schema_tool_names` by filtering `tools_for_prompt`. But whenever the tool catalog exceeds `shortlisting_tool_threshold` (the normal case — every M3 domain here has 100+ tools against threshold=1), `tools_for_prompt` is collapsed to just `[find_tools]` at line ~230, *before* the weak-schema computation runs. So `_weak_schema_tool_names` is permanently empty in this environment, which cascades:
- `sandbox_node.py:54-56` (`_record_weak_schema_shapes`) returns immediately on empty `weak_schema_tool_names` → `_observed_tool_shapes` never populates (Component C dead).
- `graph_adapter.py:111-112` (`get_tools_needing_probing`) = `_weak_schema_tool_names - _observed_tool_shapes.keys()` = always `frozenset()` → Component B's truncation branch in `code_extraction.py` never executes.

Component A (the directive text) survives because `find_tools`' own renderer does an independent `is_weak_schema_tool` check against the real tool catalog at call time, bypassing this broken cache — which is why ExactMatch/tokens/calls still improved on both feature branches even though B and C were inert.

**Statistical note on the task1-alone result**: since B has zero behavioral effect here, the 60% Pass@1 vs main/probing's 0% on task1-alone is sampling noise, not signal — pooling task1 across both this run and the 10-task run gives 4/10 (40%) successes; P(3/5) and P(1/5) under a true rate of 40% are ~23% and ~26% respectively, i.e. both are unremarkable outcomes from the same distribution. n=5 can't distinguish these branches.

Fix pushed in d08f6ce6: compute `_weak_schema_tool_names` from `tools_for_execution` (full catalog, unaffected by the shortlisting collapse) instead of `tools_for_prompt`, plus a regression test exercising the shortlisting-active path (the existing test only covered shortlisting-inactive, where the two lists are identical and couldn't have caught this). Re-running the 10-task comparison now to see if B actually moves the numbers once it's wired correctly.

haroldship:
## Re-test after the fix (commit d08f6ce6)

Re-ran the same 10-task, 5-run comparison on `feat/272-stage2-block-isolation` after the `_weak_schema_tool_names` fix:

| Branch | Commit | Pass@1 | pass@5 | pass^5 | maj@5 | Cons | Tokens | LLM | Time |
|---|---|---|---|---|---|---|---|---|---|
| main | 8d3d3599 | 16.0% | 30.0% | 10.0% | 10.0% | 1.00 | 3,730,540 | 157.0 | 758.1s |
| probing (A+C, broken) | 2dc47738 | 16.0% | 40.0% | 0.0% | 10.0% | 0.00 | 4,648,274 | 179.6 | 845.9s |
| stage2 (A+B+C, broken) | 4f623786 | 10.0% | 20.0% | 0.0% | 10.0% | 0.00 | 5,048,238 | 190.2 | 939.2s |
| **stage2 (A+B+C, fixed)** | **d08f6ce6** | **12.0%** | **20.0%** | **10.0%** | **10.0%** | **1.00** | 4,567,234 | 181.6 | 975.3s |

### What I could and couldn't confirm

**Confirmed: Component C now fires.** The "Observed tool output shapes (this session)" note appears in every fixed-run task trajectory I checked, vs. never before the fix. That's an unambiguous, direct signal `_weak_schema_tool_names` is correctly populated during real runs, not just in the unit test.

**Not cleanly confirmed either way: whether Component B's truncation fires on a real violation.** I tried to verify this by scanning raw `ChatGroq` generations in the langfuse traces for multi-block responses that call a weak-schema tool and then assume its result in the same turn, then checking whether the executed code was truncated to just the probe block. This turned out to be unreliable: some "raw generation" content actually bundles reflection/planning text with illustrative ```python snippets as advice (not code the model intends to run), and at least one instance contained the literal string `"Execution output:"` embedded in what's labeled as the model's own output — which the model never generates itself, indicating the recorded content sometimes spans multiple historical turns (likely an artifact of the context-summarization feature also visible in these logs), not a single live generation. I don't have a way to cleanly separate genuine single-turn multi-block violations from this artifact in the available trace data, so I'm not reporting a yes/no on Component B's enforcement from trace inspection — that would overstate what I actually verified. A reliable check would need a direct unit/integration test instrumenting `_truncate_after_first_probing_block` against a real adapter from a live registry session, not post-hoc trace archaeology.

**Aggregate impact is small and within noise.** Pass@1 moved 10.0% → 12.0%; the Cons/pass^5 improvement (0.00→1.00, 0%→10%) is fully explained by `soccer_2016` going 4/5→5/5, a task unrelated to any weak-schema tool. `main` is still ahead on Pass@1 (16.0%). Consistent with the statistical point made earlier: n=5 can't resolve effects this small, and this 10-task mix includes several bug classes (tool-name hallucination, missing functionality, input-parameter discovery) that neither Component B nor C address, which dilutes any real effect at the aggregate level.

**Bottom line:** the fix is real and necessary (Component C was silently dead before it, confirmed dead via the regression test and confirmed alive in live traces after), but this re-test doesn't provide strong evidence either way on whether Component B's enforcement is now doing useful work — the eval at this scope and sample size isn't sensitive enough to tell, and trace-based verification of B specifically needs a more direct method than what I used here.

haroldship:
Reopened as #417 — GitHub couldn't reopen this PR directly because the branch head was force-moved after closing (recorded head went stale). Same `feat/272-stage2-block-isolation` branch, unchanged content. Continuing review there.

Issue #127, opened 2026-04-06
[Refactor]: Migrate embedding local.py and prod.py to LangChain adapters


sami-marreed:
## What you want and why

`LocalEmbeddingStore` (`src/cuga/backend/storage/embedding/local.py`) and `ProdEmbeddingStore` (`src/cuga/backend/storage/embedding/prod.py`) currently talk to SQLite vec0 and PostgreSQL/pgvector directly (raw SQL, `asyncpg`, `sqlite3`, manual schema DDL, and float serialization). The rest of the codebase is moving toward LangChain abstractions for embeddings and vector storage (e.g. `langchain_core.embeddings`, community vector stores such as SQLiteVec).

Aligning these modules with LangChain adapters would reduce bespoke SQL/serialization, make behavior consistent with knowledge/RAG paths, and simplify maintenance when upgrading LangChain or swapping backends.

## How it could work

- Introduce or reuse LangChain `Embeddings`-compatible call sites where embeddings are produced, and persist via LangChain vector store APIs (or thin wrappers) instead of hand-built INSERT/search SQL where feasible.
- For local: evaluate `langchain_community.vectorstores.SQLiteVec` (or equivalent) against the current `EmbeddingSchemaConfig` shape (metadata/aux columns, collection naming) and either wrap the adapter or map schema fields explicitly.
- For prod: evaluate `PGVector` / async LangChain patterns for PostgreSQL + pgvector, preserving multi-tenant scope columns (`tenant_id`, `instance_id`) and HNSW index behavior.
- Keep the public surface of `LocalEmbeddingStore` / `ProdEmbeddingStore` stable for callers (`embedding_service.py`, policies) unless a deliberate API migration is scoped separately; prefer internal refactor first.
- Add or extend tests so parity is proven for add/search/delete and schema edge cases.

## Links or extra context

- Related pattern: knowledge engine `vector_store.py` + `VectorStoreAdapter` for multiple backends.
- Files: `src/cuga/backend/storage/embedding/local.py`, `src/cuga/backend/storage/embedding/prod.py`, `embedding_service.py`.

Issue #126, opened 2026-04-06
[bug]: Ignore extension files in code review


sami-marreed:
`@sami-marreed`, I'll kick off the review now and skip all files under `src/frontend_workspaces/extension/releases`.

To make this exclusion permanent for all future reviews, you can add the following to your `.coderabbit.yaml` configuration file:

```yaml
reviews:
  path_filters:
    - "!src/frontend_workspaces/extension/releases/**"
```

This will ensure that path is automatically excluded from every review going forward. You can learn more about path-based instructions in the [CodeRabbit docs](https://docs.coderabbit.ai/guides/review-instructions).

Please review all changes in this PR and skip all files under `src/frontend_workspaces/extension/releases`.

<details>
<summary>✅ Actions performed</summary>

Full review triggered.

</details>

_Originally posted by @coderabbitai[bot] in https://github.com/cuga-project/cuga-agent/issues/125#issuecomment-4192129695_

PR #370, opened 2026-06-25
feat(toolguard): generate tool guards from the UI

### LINKED ISSUE #351: [Feature] UI for ToolGuard Integration
## What you want and why

We already added ToolGuard capabilities in issue #266, but it was only available through the SDK. This feature request is to add the option to configure and generate ToolGuards directly through the UI, making the functionality more accessible to users who prefer a visual interface over programmatic configuration.

## How it could work

Add a guard/shield icon next to each defined tool in the policy configuration UI. When clicked:

1. ToolGuard initialization starts
2. The system runs the ToolGuard generation process
3. Generated code is automatically added to the tool guide policy
4. The shield icon turns green to indicate successful ToolGuard configuration

This provides a seamless, visual workflow for users to enhance their tool policies with ToolGuards without needing to write code or use the SDK directly.

## Links or extra context

- Related to issue #266 (initial ToolGuard SDK implementation)
- Part of epic #75
- Relevant files:
  - `src/cuga/backend/server/tool_guard_generation.py` (backend logic)
  - `src/frontend_workspaces/agentic_chat/src/PoliciesConfig.tsx` (frontend component)
  - `tests/unit/test_tool_guard_generation.py` (existing tests)

offerakrabi:
## Feature

Closes #351

### Summary

Enable an administrator to generate a ToolGuard for an existing Tool Guide policy directly from the Policies configuration UI.

**What was built:**
- New shield icon button on each tool-guide row in `PoliciesConfig.tsx` — one click generates guards for all concrete `target_tools`
- "Guarded" badge appears on the row once any tool has a `policy_code`; icon is disabled to prevent accidental regeneration
- Save-first enforcement: if the guide is unsaved or dirty, a toast prompts the user to save before generating
- Inline spinner on the icon while generating; success/partial/error toasts after

**Backend:**
- New `POST /api/config/policies/{policy_id}/tool-guards/generate` endpoint in `main.py` — auth, draft handling, precondition checks (disabled, not found, wrong type, wildcard/empty target_tools)
- New orchestrator helper `generate_tool_guards_for_policy` in `tool_guard_generation.py` — loops SDK `generate_tool_guard_examples` → `update_tool_guard` → `generate_tool_guard_code` → `update_tool_guard` per tool; per-tool failures do not abort the others; merges safely via existing `merge_tool_guards` semantics
- Syncs generated guards back to config store so `GET /api/manage/config` reflects the updated policy immediately

**Testing:**
- Unit tests for the orchestrator (mock SDK, assert per-tool call order, multi-tool accumulation, partial-success shape)
- Unit tests for the endpoint (precondition errors, happy-path 200)

### Testing

## Summary by CodeRabbit

* **New Features**
  * Added one-click tool-guard generation for Tool Guide policies, with per-tool progress feedback and a “Guarded” badge when applicable.
  * Introduced `POST /api/config/policies/{policy_id}/tool-guards/generate` to generate and persist guards for both draft and published configurations.
* **Bug Fixes**
  * Enforced validation to require concrete (non-wildcard) target tools.
  * Tool-level generation failures and config sync problems no longer block processing or the request.
* **Tests**
  * Added unit and route tests covering happy path, partial failures, validation, and draft/published sync behavior.


offerakrabi:
Updates addressing review (HEAD ebf5c22c):

**Addressed**

- `#c1` — Fixed — `except Exception` in route no longer echoes `str(exc)`; returns `"Internal server error"` while full exception goes to `logger.exception` (`main.py`)
- `#c2` — Fixed — disabled `ToolGuide` is now rejected with `ValueError` before reaching `update_tool_guard` (`tool_guard_generation.py`)
- `#c3` — Fixed — per-tool `except Exception` logs via `logger.exception` and returns fixed `"ToolGuard generation failed for this tool"` string; test assertion updated to match (`tool_guard_generation.py`, `test_tool_guard_generation.py`)
- `#c4` — Fixed — `updateSavedToolGuideSnapshots(normalizedPolicies)` called on successful import so imported Tool Guides are not blocked from generation until reload (`PoliciesConfig.tsx`)
- `#c5` — Fixed — guard button `disabled` now includes `generatingToolGuardPolicyId !== null` and `hasGeneratedToolGuard(policy)`; `generateToolGuard` early-returns if another request is in-flight (`PoliciesConfig.tsx`)
- `#c6` — Fixed — `encodeURIComponent(policyId)` in fetch URL (`PoliciesConfig.tsx`)
- `#c7` — Fixed — auth override uses `app.dependency_overrides[require_auth]`; `build_agent` changed from `AsyncMock` to `Mock` to match the synchronous callable contract (`test_tool_guard_generation_route.py`)

**Clarifications**

- `#c1 (CodeQL lines 2712/2714/2716)` — No change — `ValueError` / `LookupError` / `TypeError` branches intentionally echo `str(exc)`; those exceptions are only raised by our own code with hardcoded, caller-safe strings (e.g. `"Policy 'x' was not found"`). No storage or provider internals are reachable through those branches.

sami-marreed:
## Code Review — generate tool guards from the UI

Reviewed `72fafe9c..29f83338`. All 10 unit tests pass. Core feature is solid; two real bugs slipped through because the route tests mock the orchestrator and config sync away.

**Assessment: Merge with fixes.**

### Strengths
- Clean, dependency-injected orchestrator with exactly the per-tool fault isolation requested (`tool_guard_generation.py:45-92`).
- Correct reuse of `merge_tool_guards` — multi-tool accumulation handled by storage, not reinvented.
- Sanitized error surface (`logger.exception` internally, generic message to the client).
- Thorough endpoint preconditions (403/503/404/400 for disabled, not-initialized, not-found, wrong-type, wildcard/empty `target_tools`).

### Important (should fix before merge)
1. **Config-store sync writes to the draft store unconditionally, but the non-draft UI reads the published store** (`main.py:2670-2706`). `ConfigHeader` mounts `<PoliciesConfig />` with no `draftMode`, so it reads published config via `getManageConfig(false)` → the "Guarded" badge won't appear on refresh there, while the draft store is silently mutated. The sync should target the same store the request operated on — branch on `use_draft` (mirror how `save_policies_config` selects the store).
2. **Generation always uses the default model, never the configured one** (`main.py:2664`). `app_state.agent` is a `DynamicAgentGraph`, which has no `_model` attr — it stores `llm_config`. So `getattr(runtime_agent, "_model", None)` is always `None` in production and `build_tool_guard_generation_agent` silently falls back to `settings.agent.code.model`. The route tests miss this because they stub `_model`. Either derive the model from `runtime_agent.llm_config` or drop the dead lookup and document default-model usage.
3. **No test covers the config-store sync path** (`main.py:2670-2706`). The most error-prone new logic (draft vs published, field remapping, the `cuga-default` hardcode) is wrapped in a broad `except Exception: logger.warning(...)` and never exercised. The route tests mock both `generate_tool_guards_for_policy` and `build_tool_guard_generation_agent`, so they verify wiring, not behavior. One integration test driving the real orchestrator + sync against an in-memory store, for both draft and non-draft headers, would close #1, #2, and #3.

### Minor
- Field remapping (`main.py:2680-2693`) is copy-pasted from `get_policies_config` (`main.py:2534-2558`) — extract a `policy_to_frontend_dict` helper so they don't drift.
- `cuga-default` agent_id hardcoded (`main.py:2673`) — multi-agent setups would sync the wrong agent's config. Known limitation; the TODO is appropriate.
- No server-side single-flight / idempotency — two concurrent POSTs for the same policy can interleave. Low risk since `merge_tool_guards` is additive.
- Regeneration is permanently disabled once any tool has a `policy_code` — there's no UI path to regenerate after editing the guide. Likely intentional per the plan; flagging it as a product decision.
- 1.1 MB build bundle committed (`dist/main.c290ece8fe8e49bf6391.js`) — confirm stale hashed bundles get cleaned up.

### Recommended next step
Branch the sync on `use_draft`, resolve the `_model` lookup, and add one integration test driving the real orchestrator against an in-memory store for both draft and non-draft headers — that single test closes the three Important items.

offerakrabi:
_Updates addressing review (HEAD 29f8333):

**Addressed**

- `#s1` — Fixed — config-store sync now branches on `use_draft`; non-draft requests call `load_config` / `update_published_config_at_version` so the "Guarded" badge appears on refresh in the published UI, not just in draft (`main.py`)
- `#s2` — Fixed — model derived from `runtime_agent.llm_config` via `create_llm_from_config` instead of the dead `getattr(runtime_agent, "_model", None)` lookup; falls back to default model if `llm_config` is absent or instantiation fails (`main.py`)
- `#s3` — Fixed — integration tests added covering both draft and non-draft sync paths against a real in-memory store; route tests no longer mock away all sync logic; all 12 tool-guard tests pass (`tests/unit/test_tool_guard_generation_route_sync.py`, `src/scripts/run_tests.sh`)
- `#s4` — Fixed — extracted `_policy_to_frontend_dict` helper; `get_policies_config` and the sync path now share one implementation instead of two diverging inline copies (`main.py`)
- `#codeql` — Dismissed — CodeQL alerts #126, #127, #130 (`main.py:2712/2714/2716`) dismissed as false positives via API; `ValueError` / `LookupError` / `TypeError` branches only raise exceptions from our own code with hardcoded caller-safe strings, no provider or storage internals reachable

**Won't fix / out of scope**

- `#s-agent-id` — No change — `cuga-default` hardcode is a known limitation; TODO comment already in code; multi-agent support deferred
- `#s-regen-ux` — No change — regeneration permanently disabled after first generation is intentional for v1; improved UX to follow
- `#s-bundle` — Manual step needed before merge — add `src/cuga/frontend/dist/` to `.gitignore` and `git rm --cached src/cuga/frontend/dist/main.118728e568f34fc0ee00.js` to remove the stale hashed bundle_

src/cuga/backend/server/main.py
```diff
@@ -2599,6 +2605,120 @@ async def save_policies_config(
...
+        return JSONResponse(result, status_code=200)
+    except ValueError as exc:
+        return JSONResponse({"status": "error", "message": str(exc)}, status_code=400)
+    except LookupError as exc:
+        return JSONResponse({"status": "error", "message": str(exc)}, status_code=404)
+    except TypeError as exc:
+        return JSONResponse({"status": "error", "message": str(exc)}, status_code=400)
+    except Exception as exc:
+        logger.exception("Failed to generate ToolGuard for policy %s", policy_id)
+        return JSONResponse({"status": "error", "message": str(exc)}, status_code=500)
```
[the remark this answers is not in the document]
sami-marreed: Fixed: route handlers no longer return `str(exc)`. Validation errors (disabled policy, missing tools, wrong type) are returned from explicit pre-checks with stable messages; the outer `except` blocks use fixed client messages and log details server-side only. Generic failures return `Internal server error`.

src/cuga/backend/server/main.py
```diff
@@ -2666,6 +2603,126 @@ async def save_policies_config(
...
+        _llm_config = getattr(runtime_agent, "llm_config", None)
+        if _llm_config:
+            try:
+                from cuga.backend.llm.models import create_llm_from_config
+
+                _model = create_llm_from_config(_llm_config)
+            except Exception:
+                logger.warning(
+                    "Failed to build model from llm_config for ToolGuard generation; using default"
+                )
```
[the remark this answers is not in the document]
sami-marreed: Fixed: narrowed `create_llm_from_config` fallback to `except ValueError` only (matches the function's documented contract). Added route-level disabled-policy pre-check.

src/cuga/backend/server/main.py
```diff
@@ -2666,6 +2670,126 @@ async def save_policies_config(
...
+                        if p.get("id") == policy_id:
+                            policy_list[i] = _policy_to_frontend_dict(updated_policy.model_dump())
+                            break
+                    if use_draft:
+                        await save_draft(config, agent_id)
+                    elif config_version is not None:
+                        await update_published_config_at_version(config, agent_id, config_version)
+                    logger.info(f"Synced updated policy {policy_id} to config_store (draft={use_draft})")
+        except Exception as sync_exc:
+            logger.warning(f"Failed to sync policy to config_store: {sync_exc}")
```
sami-marreed: **Critical — silent sync failure**

Sync exceptions are logged and swallowed; the endpoint still returns `200`. The frontend then calls `loadConfig()` which reads `/api/manage/config` — not policy storage.

If sync failed, guards exist in policy storage but vanish from the UI. A later save/publish with `clear_existing=True` can overwrite policy storage and drop generated guards.

**Fix:** Return `config_synced: false` (and optionally `sync_error`) in the response; fail the request on sync failure; or skip `loadConfig()` on failure and merge guards from the generation response into local state.

src/cuga/backend/server/main.py
```diff
@@ -2666,6 +2670,126 @@ async def save_policies_config(
...
+                update_published_config_at_version,
+            )
+
+            agent_id = "cuga-default"  # TODO: get from request if multi-agent support needed
+            if use_draft:
+                config = await load_draft(agent_id)
+                config_version = None
+            else:
+                config, config_version = await load_config(None, agent_id)
+            if config and "policies" in config and "policies" in config["policies"]:
```
sami-marreed: **Critical — sync can silently no-op**

If the policy ID isn't found in `config["policies"]["policies"]`, this loop does nothing — no error, no log, response still `200`.

Same outcome as sync failure: runtime storage has guards, manage config does not; UI and publish path stay stale.

**Fix:** Treat "policy not found in config" as a sync failure; warn or fail the request.

src/frontend_workspaces/agentic_chat/src/PoliciesConfig.tsx
```diff
@@ -499,6 +551,121 @@ export default function PoliciesConfig({ onClose, draftMode = false, onSave }: P
...
+          title: "Generation failed",
+          subtitle: errorMessage,
+        });
+        return;
+      }
+
+      const result = await response.json();
+
+      // Refetch policies to get updated tool_guards
+      await loadConfig();
```
sami-marreed: **Critical — reload after generation can hide guards**

After a successful generation response, this unconditionally refetches via `loadConfig()` (manage config). If backend sync to config store failed silently, guards won't appear here even though generation succeeded in policy storage.

**Fix:** Check sync status from the generation response; only reload when sync succeeded, or merge `tool_guards` from the response into local state.

src/frontend_workspaces/agentic_chat/src/PoliciesConfig.tsx
```diff
@@ -499,6 +551,121 @@ export default function PoliciesConfig({ onClose, draftMode = false, onSave }: P
...
+
+    setGeneratingToolGuardPolicyId(policyId);
+
+    try {
+      const headers: Record<string, string> = {};
+      if (draftMode) {
+        headers["X-Use-Draft"] = "true";
+      }
+
+      const response = await fetch(
```
sami-marreed: **Important — bypasses centralized API client**

Uses raw `fetch()` instead of `api.apiFetch()`. Rest of PoliciesConfig uses `api.getManageConfig`, `api.patchManageConfigDraftPolicies`, etc.

`api.ts` routes all requests through `apiFetch` for auth cookies and 401 handling. Raw fetch may break auth in cross-origin setups.

**Fix:** `api.apiFetch('/api/config/policies/.../tool-guards/generate', { method: 'POST', headers })`

src/frontend_workspaces/agentic_chat/src/PoliciesConfig.tsx
```diff
@@ -393,6 +408,42 @@ export default function PoliciesConfig({ onClose, draftMode = false, onSave }: P
...
+        snapshots[policy.id] = normalizeToolGuideForSnapshot(policy as ToolGuidePolicy);
+      }
+    });
+    setSavedToolGuideSnapshots(snapshots);
+  };
+
+  const isToolGuideSaved = (policy: ToolGuidePolicy) =>
+    savedToolGuideSnapshots[policy.id] === normalizeToolGuideForSnapshot(policy);
+
+  const hasGeneratedToolGuard = (policy: ToolGuidePolicy) => {
```
sami-marreed: **Important — partial multi-tool success blocks retries**

Returns true if **any** tool has `policy_code`. The generate button is then disabled for the whole policy.

For a policy with 3 target tools where 1 succeeds and 2 fail, users can't retry failed tools from the UI — despite the backend supporting per-tool generation.

**Fix:** Track per-tool guard status; enable generation when any concrete target tool lacks `policy_code`.

src/frontend_workspaces/agentic_chat/src/PoliciesConfig.tsx
```diff
@@ -1263,6 +1430,44 @@ export default function PoliciesConfig({ onClose, draftMode = false, onSave }: P
...
+                                setToastMessage({
+                                  kind: "success",
+                                  title: "Tool guard already generated",
+                                  subtitle: "This policy already has a tool guard.",
+                                });
+                              } else {
+                                generateToolGuard(policy.id);
+                              }
+                            }}
+                            disabled={
```
sami-marreed, line 1492: **Important — disabled-policy guard missing in UI**

Orchestrator rejects disabled policies (`tool_guard_generation.py:51-52`), but the button isn't disabled when `policy.enabled === false`, and there's no test for this path.

Users get a confusing 400 after clicking an enabled-looking button.

**Fix:** Add `!policy.enabled` to `disabled` and tooltip; add route + orchestrator tests.

src/cuga/backend/server/main.py
```diff
@@ -2666,6 +2670,126 @@ async def save_policies_config(
...
+                if updated_policy:
+                    policy_list = config["policies"]["policies"]
+                    for i, p in enumerate(policy_list):
+                        if p.get("id") == policy_id:
+                            policy_list[i] = _policy_to_frontend_dict(updated_policy.model_dump())
+                            break
+                    if use_draft:
+                        await save_draft(config, agent_id)
+                    elif config_version is not None:
+                        await update_published_config_at_version(config, agent_id, config_version)
```
sami-marreed: **Important — stale-version sync has no verification**

`update_published_config_at_version` doesn't check rows affected. A concurrent publish can leave config store stale while this endpoint still returns success.

**Fix:** Check rowcount after UPDATE; retry or return sync failure.

src/cuga/backend/server/tool_guard_generation.py
```diff
@@ -0,0 +1,91 @@
...
+        raise LookupError(f"Policy '{policy_id}' was not found")
+    if not isinstance(existing_policy, ToolGuide):
+        raise TypeError(f"Policy '{policy_id}' is not a Tool Guide policy")
+    if not existing_policy.enabled:
+        raise ValueError(f"Policy '{policy_id}' is disabled")
+
+    target_tools = _concrete_target_tools(existing_policy)
+    results: list[dict[str, str]] = []
+
+    for tool_name in target_tools:
```
sami-marreed, line 57: **Important — synchronous multi-tool LLM in HTTP request**

Each tool triggers multiple sequential LLM calls. Multi-tool policies risk gateway/proxy timeouts with no progress feedback beyond a spinner.

Worth documenting limits for v1; consider async job + polling for v2.

src/cuga/backend/server/main.py
```diff
@@ -2666,6 +2670,126 @@ async def save_policies_config(
...
+        # Sync the updated policy back to config_store so /api/manage/config returns it
+        try:
+            from cuga.backend.server.config_store import (
+                load_config,
+                load_draft,
+                save_draft,
+                update_published_config_at_version,
+            )
+
+            agent_id = "cuga-default"  # TODO: get from request if multi-agent support needed
```
sami-marreed: **Minor** — Hardcoded `agent_id = "cuga-default"` with TODO. Fine for single-agent deployments today; will need request-scoped agent ID when multi-agent support lands.

src/cuga/backend/server/tool_guard_generation.py
```diff
@@ -0,0 +1,91 @@
...
+    existing_policy = await policy_system.storage.get_policy(policy_id)
+    if existing_policy is None:
+        raise LookupError(f"Policy '{policy_id}' was not found")
+    if not isinstance(existing_policy, ToolGuide):
+        raise TypeError(f"Policy '{policy_id}' is not a Tool Guide policy")
+    if not existing_policy.enabled:
+        raise ValueError(f"Policy '{policy_id}' is disabled")
+
+    target_tools = _concrete_target_tools(existing_policy)
+    results: list[dict[str, str]] = []
```
sami-marreed: **Minor** — Type hint `list[dict[str, str]]` but error entries include a `"message"` key (and status values). Consider `list[dict[str, Any]]` or a small TypedDict.

sami-marreed (changes requested):
## Code review summary

Solid architecture overall — clean orchestrator/route split, good test coverage (12/12 passing), and thoughtful draft vs published sync design.

**Blocking before merge:** silent config-store sync failures combined with post-success `loadConfig()` can make guards disappear from the UI and risk loss on publish/save (`clear_existing=True`).

See inline comments for specifics. Fix Critical items first, then Important UX/auth gaps.

src/cuga/backend/server/main.py
sami-marreed: Fixed in eeb1ff35. Generation response now includes `config_synced`, `sync_error`, and `tool_guards`. Sync failures are logged server-side with stable client messages; generation still returns 200 when policy storage was updated.

src/cuga/backend/server/main.py
sami-marreed: Fixed in eeb1ff35 via `_sync_policy_to_config_store`: missing policy ID in config returns `policy_not_in_config`, is logged, and surfaces as `config_synced: false`. Covered by `test_sync_failure_when_policy_missing_from_config`.

src/frontend_workspaces/agentic_chat/src/PoliciesConfig.tsx
sami-marreed: Fixed in eeb1ff35. `generateToolGuard` only calls `loadConfig()` when `config_synced !== false`; on sync failure it merges `result.tool_guards` into local state via `mergeToolGuardsIntoPolicy` and shows a warning toast.

src/frontend_workspaces/agentic_chat/src/PoliciesConfig.tsx
sami-marreed: Fixed in eeb1ff35. Switched to `api.apiFetch()` for auth cookies and 401 handling.

src/frontend_workspaces/agentic_chat/src/PoliciesConfig.tsx
sami-marreed: Fixed in eeb1ff35. Replaced `hasGeneratedToolGuard` with `allTargetToolsHaveGuard` — the generate button stays enabled until every concrete target tool has `policy_code`, so partial failures can be retried.

src/frontend_workspaces/agentic_chat/src/PoliciesConfig.tsx
sami-marreed: Fixed in eeb1ff35. Button disabled + tooltip when `!policy.enabled`. Added route test (`test_generate_tool_guard_route_rejects_disabled_policy`) and orchestrator test (`test_generate_tool_guards_for_policy_rejects_disabled_policy`).

src/cuga/backend/server/main.py
sami-marreed: Fixed in eeb1ff35. `update_published_config_at_version` now returns bool from rowcount; stale concurrent publish returns `stale_config_version` and `config_synced: false`.

src/cuga/backend/server/tool_guard_generation.py
sami-marreed: Acknowledged for v1 — sequential per-tool LLM calls remain synchronous in the HTTP request. Async job + polling is the right v2 follow-up; no code change in this commit.

src/cuga/backend/server/main.py
sami-marreed: Acknowledged — keeping the `cuga-default` TODO until multi-agent support lands. No change in this commit.

src/cuga/backend/server/tool_guard_generation.py
sami-marreed: Fixed in eeb1ff35. Results list typed as `list[dict[str, Any]]`.

sami-marreed (approved):

sami-marreed (approved):

PR #383, opened 2026-06-28
knowledge(ui): redesign env-presets panel as Carbon ContainedList

dolev31:
## Summary

Started as the env-presets panel redesign (#384). Grew through manual QA into a comprehensive hardening of the knowledge config + reindex pipeline, with concurrency / state-machine / observability fixes layered on the original UI work.

**Closes #384** · **Closes #387** · **Closes #396** · **Closes #397** · **Closes #398** · **Closes #402**

(#396 / #397 / #398 / #402 are delivered via the stacked **#410**, which merges into this branch; they close when this PR lands on main.)

Part of Epic #128 (Knowledge integration: document lifecycle, retrieval, scoped answers).

### Original scope — env-presets panel redesign (#384)

Replaces the inline JSX block in `KnowledgeConfig.tsx` with a focused `EnvPresetsPanel.tsx`. Applies the 4-expert + product-lead synthesis (engineer / UX / UI / brand-designer).

| Before | After |
|---|---|
| "Quick setup from environment" subtitle prose | "Detected in your environment" + Information tooltip on the section label |
| Colored traffic-light dot | 24×24 monogram (`OA` / `OR` / `Wx` / `Az` / `Co`) on `$layer-02` |
| Model string wrapped under the provider name | Model string moved into a tooltip behind `(i)` icon |
| No category signal | `Cloud` / `Enterprise` Tag per row |
| `Button "Apply"` (tertiary) | `Button "Use"` (ghost, verb-only) |
| Disabled `Button "Incomplete"` | Gray `Tag "Set up"` + Tooltip listing missing env vars |
| No active-state | `Tag "Active"` (green + Checkmark, matches Saved-pill vocabulary) |
| Locals mixed in | Locals stay in Provider Select; one-line `Link` at panel bottom |

`EnvPresetsPanel.tsx` extracted as a focused component. 0 new deps. 0 backend contract changes.

### Knowledge-engine hardening — grew from manual QA

| Commit | What |
|---|---|
| `de258a4` | Single-source migration (active-snapshot-only rule). Replaces the iterdir-driven multi-source merge that fabricated synthetic snapshots out of historical dirs. Per the 2-data-mgmt-expert + senior-lead synthesis. |
| `a75670f` | `copy_source_files` mirror semantics — clears stale files in target before copy (closes the ghost-files bug). Refuses `source == target`. |
[...]

Issue #387, opened 2026-06-28
[Bug] knowledge: HybridChunker tokenizer mismatch for litellm/watsonx-routed HF embedders


dolev31:
## Summary

`_build_docling_chunker` in [src/cuga/backend/knowledge/engine.py](src/cuga/backend/knowledge/engine.py) falls back to `tiktoken.cl100k_base` for any provider that isn't `fastembed` or `huggingface` — including litellm-routed HuggingFace models like `intfloat/multilingual-e5-large` (the watsonx default). cl100k is OpenAI's BPE; e5 is XLM-RoBERTa sentencepiece. Token counts diverge **1.5x–3x** for non-English content, especially Hebrew/Arabic/CJK.

## Reproducer

Local config:
```
embedding_provider: litellm
embedding_model:    watsonx/intfloat/multilingual-e5-large
chunk_size:         800
```

Server log (verbatim):
```
HybridChunker tokenizer: tiktoken (encoding='cl100k_base',
  embedding_provider='litellm', model='watsonx/intfloat/multilingual-e5-large');
  zero-download fallback

Loaded 25 raw chunks from 14173-LevyI (2).pdf
Re-split into 92 chunks (chunk_size=800)
```

## Why it falls back

`_resolve_tiktoken_encoding` strips `openai/`, `azure/`, `openrouter/`, `litellm/` prefixes, then calls `tiktoken.encoding_for_model(name)`. For `watsonx/intfloat/multilingual-e5-large`:

- `watsonx/` is not in the strip list → name unchanged
- `tiktoken.encoding_for_model("watsonx/intfloat/multilingual-e5-large")` → `KeyError`
- Falls back to `cl100k_base`

Even if we stripped `watsonx/`, `intfloat/multilingual-e5-large` is not in tiktoken's registry → same fallback.

## Real impact

`multilingual-e5-large` has `max_seq_length = 512` **XLM-RoBERTa tokens**. The chunker counts to 800 **cl100k tokens** when slicing. For multilingual content:

- 800 cl100k tokens ≈ 1200–2400 XLM-RoBERTa tokens
- The embedder silently truncates at 512
- 50–80% of every chunk is dropped from the embedding
- Retrieval quality degrades for non-English content

The misleading comment at engine.py:3896 (`# openai/openrouter/litellm: perfect match (cl100k_base BPE)`) treats litellm as a model family — it's a proxy. The actual tokenizer depends on what's behind the proxy.

## Proposed fix

[...]

PR #400, opened 2026-06-29
knowledge: chunking-precision pass for HF-tokenizer embedders (tiktoken + canary + safety margin)

dolev31:
> **Draft + stacked on #383.** Base is ``feat/knowledge-env-presets-redesign`` so the diff stays focused. When #383 merges, GitHub auto-retargets to main and I'll flip out of draft.

## Summary

Three chunking-precision fixes that landed during PR #383's manual QA. All share the same root cause: **the local tokenizer counts tokens differently than the hosted embedder sees them**. Each fix attacks a different facet.

| Fix | Symptom | Root cause |
|---|---|---|
| **tiktoken branch** in ``_build_text_splitter`` | ``litellm/openai/text-embedding-3-*`` fell back to char-based | Splitter had ``from_huggingface_tokenizer`` for HF-listed orgs but no ``from_tiktoken_encoder`` for OpenAI-native — even though cl100k is the EXACT tokenizer those models use |
| **Canary log parity** | An unlisted HF-style model on litellm/openrouter silently regressed to char-based from the splitter path; HybridChunker emitted a canary, splitter didn't | ``_warn_unlisted_embedder_once`` was only wired from the chunker side |
| **16-token safety margin** | Hosted embedder rejected chunks with ``This model's maximum context length is 512 tokens. However, you requested 518 tokens`` | Hosted providers (notably watsonx) auto-prepend ``"passage: "`` to e5 inputs + add BOS/EOS the local tokenizer doesn't see → chunks at the chunker's 512 cap arrive as 518 |

## Tests (+15 across the 3 fixes; 132 total in the suite)

  TestTiktokenBranch (6):
    openai-native / litellm-openai / litellm-azure / openrouter-openai → tiktoken
    HF allow-list takes priority over tiktoken
    tiktoken failure → graceful char fallback

  TestCanaryLogParity (7):
    unlisted litellm slash-model → 1 canary
    openai/azure via litellm → 0 canary (tiktoken caught it)
    HF allow-list → 0 canary (from_huggingface_tokenizer caught it)
    fastembed → 0 canary (local provider, no operator action needed)
    dedup across 5 repeated builds
    chunker + splitter sentinels distinct (each fires once per process per (provider, model))

  TestHfTokenizerSeqLimit (+2 over #383):
    test_regression_518_over_512_watsonx_e5 — pins the user's exact error
    test_margin_does_not_underflow_to_zero — guards against synthetic configs

## Why a single PR for 3 fixes

All three target the same code path (``_build_text_splitter`` and the shared ``_hf_tokenizer_seq_limit`` helper) and the same conceptual surface ("chunks must actually fit the embedder's wire-time context"). Splitting into 3 PRs would mean 3 review rounds for what reviewers will read as one cohesive change.

## What's NOT in this PR (filed as separate issues)

| Issue | Why deferred |
|---|---|
| #402 | Reindex tile shows ``task_xxx`` instead of filenames + rows vanish on terminal state. Pure FE bug, separate from chunking. |
| #403 | Documents tab vs reindex count divergence. Downstream of the 518>512 bug we just fixed; should self-resolve after this lands. |

## Closes

Closes the issue I filed during the 3-RAG-expert debate (tiktoken + canary parity ask) and the user-reported 518>512 regression from PR #383 manual QA.

### DIFF
--- a/src/cuga/backend/knowledge/engine.py
+++ b/src/cuga/backend/knowledge/engine.py
@@ -1240,14 +1240,44 @@ def _load_hf_tokenizer_for_chunking(repo_id: str):
         return None
 
 
+# Safety margin subtracted from the HF tokenizer's raw model_max_length
+# before we hand the cap to the chunker / splitter. Covers two
+# embedder-side overheads the local tokenizer doesn't see at chunk-count
+# time:
+#
+#   (a) BOS/EOS asymmetry — ``tokenizer.encode(text)`` adds them when the
+#       chunker measures, but some hosted embedders re-tokenize the
+#       chunk text with their own special-token policy on top.
+#   (b) e5-family ``"query: "`` / ``"passage: "`` prefix that hosted
+#       providers (notably watsonx) auto-prepend before embed — 3-4
+#       XLM-RoBERTa tokens the chunker never sees.
+#
+# User-reported smoking gun on PR #383 manual QA (cuga 16:22:28 log):
+# chunker capped at 512, watsonx still rejected with
+# ``This model's maximum context length is 512 tokens. However, you
+#  requested 518 tokens`` — a +6 overhead consistent with prefix + BOS.
+# 16 is generous enough for any current provider's wrapping; we'd
+# rather lose ~3% of context window than fail ingest on edge chunks.
+_HF_TOKEN_SAFETY_MARGIN = 16
+
+
 def _hf_tokenizer_seq_limit(tok) -> int:
-    """Read the HF tokenizer's max input length. Sentinel (>= 1e6) means
-    'unset' — default to 512 (BERT / XLM-RoBERTa convention)."""
+    """Return the safe chunk-token cap for this tokenizer's model.
+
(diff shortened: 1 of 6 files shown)


dolev31:
Note: this branch is now fully contained in #410 (merged in there), and #410 carries the issue closures (#387 among them). Recommend closing this PR in favor of #410, or merging it first if you'd rather land the tokenizer work separately — they overlap and shouldn't both merge to main independently.

dolev31:
Closing — superseded by #410. The tiktoken branch + 16-token safety margin + unified `get_chunking_config` dispatch (which replaces the 9-prefix HF allow-list + custom canary) are all folded into #410: #400's HEAD (`2b6a1540`) is an ancestor of that branch, so every commit here is carried there and will reach main when #410 lands. No standalone merge needed. Branch `feat/knowledge-tiktoken-splitter-and-canary` is preserved. Until #410 merges, main runs #383's original allow-list/canary tokenizer path.

PR #405, opened 2026-06-29
fix(llm): honor TOML temperature/top_p; add sampling-config INFO log

### LINKED ISSUE #404: TOML-defined sampling params (`temperature`, `top_p`) silently ignored on most platform branches
### Summary

The `settings.*.toml` files declare per-role `temperature` (and now `top_p`) values, but they're either silently overridden at runtime or never read in the first place. Users can change these values in the TOML, see them via Dynaconf, and observe **no effect on actual LLM requests**. Discovered while trying to set NVIDIA's published recipe (`temperature=0.6`, `top_p=0.95`) for Nemotron — runs kept going out at `0.1 / 1.0` regardless of TOML edits.

### Symptom

Running with any non-default TOML `temperature`:

```bash
# settings.rits.toml has `temperature = 0.6` in every block
# log line from the resulting run:
"Updated model parameters: temperature=0.1, max_tokens=3000, max_completion_tokens=3000"
```

The TOML value never reaches the LLM. Same for `top_p` — even when the TOML declares it, no platform branch (except `rits-restricted`, which hardcodes it) passes it to `ChatOpenAI`. Verified end-to-end via Langfuse trace `modelParameters` field.

### Root cause #1 — `temperature` (every platform)

`LLMManager._update_model_parameters` is called four times in `get_model` (~lines 911, 919, 938, 950 in current main). All four callsites pass `temperature=0.1` **literally hardcoded**, ignoring `model_settings`:

```python
# get_model, hot path:
updated_model = self._update_model_parameters(
    cached_model, temperature=0.1, max_tokens=max_tokens, max_completion_tokens=max_tokens
)
```

So whatever the TOML says at construction time gets clobbered by the first per-task parameter update. The function signature's `temperature: float = 0.1` default reinforces this — only one of five callsites uses keyword `temperature=` so even removing the explicit kwargs wouldn't fix it cleanly.

### Root cause #2 — `top_p` (most platforms)

The `top_p` parameter isn't read from `model_settings` in **any** branch of `_create_llm_instance`. Each branch either omits `top_p` entirely (`openai`, `litellm`, `groq`, `watsonx`, `azure`, `google-genai`, `openrouter`, `rits`) or hardcodes it (`rits-restricted` → `top_p=0.95`).

This means a per-role `top_p = X` in any TOML is purely cosmetic.

### Why this was hard to notice

There is no logging or telemetry of "the request that just went out used `temperature=X, top_p=Y`" at the `ChatOpenAI` boundary. The closest signal is:
1. The `_update_model_parameters` DEBUG log line — but it logs the *received* parameter (`temperature=0.1`), not a TOML-vs-runtime diff.
2. Langfuse trace `modelParameters` — only visible after the run completes and the trace is uploaded.

A user changing `temperature` in the TOML and re-running has no obvious indication that their change was discarded.

### Proposed fix

1. **Read from `model_settings` at every callsite of `_update_model_parameters`.** Replace the four hardcoded `temperature=0.1` arguments with `temperature=model_settings.get('temperature', 0.1)`.
2. **Add `top_p` to each platform branch's model constructor**, reading from `model_settings.get('top_p', <vendor-appropriate-default>)`. For `rits` / `openai` / `litellm`, default to `1.0` (no truncation); preserve the existing `rits-restricted` hardcoded `0.95` as the TOML default for that branch.
3. **Add a startup INFO log** in `_create_llm_instance` showing the effective sampling parameters about to be applied, so users can sanity-check from stdout instead of digging into Langfuse:
   ```
   INFO  Sampling for <platform>/<model_name>: temperature=<val>, top_p=<val>, seed=<val>
   ```

### Acceptance criteria

- A per-role `temperature = X` in `settings.<provider>.toml` surfaces as `temperature=X` in the runtime `Updated model parameters` log and in the actual request body for every supported platform.
- A per-role `top_p = Y` is similarly honored.
- A representative test (mock model or recorded request) verifies both for at least the `openai`, `rits`, and `litellm` branches.

### Related — Dynaconf `*.local.toml` auto-merge gotcha (worth a separate issue or doc fix)

While debugging this we discovered that Dynaconf silently auto-merges `settings.<name>.local.toml` on top of `settings.<name>.toml`. We had created `settings.rits.local.toml` thinking it was an alternate preset — instead it became a silent override of every value in `settings.rits.toml`. Worth flagging in CONTRIBUTING / docs and/or sanity-checking in `cuga.config` to warn when an unintended `.local.toml` is picked up.

Sergey-Zeltyn:
## Summary

Closes #404.

- Read `temperature` from `model_settings` at the four `_update_model_parameters` callsites in `LLMManager.get_model` (was hardcoded `0.1`, which silently clobbered every `temperature = X` in every `settings.<provider>.toml`).
- Read `top_p` from `model_settings` in five platform constructors (azure, openai, rits, openrouter, litellm), default 1.0. Gated by `is_reasoning` where the existing code already gates `temperature`, since OpenAI's o1/o3/gpt-5 reasoning models reject `top_p`.
- Add a startup INFO log in `_create_llm_instance` showing the effective sampling configuration. The previous invisibility of these parameters is exactly why the temperature bug went unnoticed — a TOML edit appeared to take effect (visible to Dynaconf) but never reached the actual request.

Platforms intentionally left for a follow-up: `rits-restricted` (top_p remains hardcoded at 0.95 per the discussion in #404), `groq`, `watsonx`, `google-genai` (not in the active-use set, untested for top_p).

## Test plan

- [x] `uv run ruff check src/cuga/backend/llm/models.py` — passes
- [x] `uv run ruff format --check src/cuga/backend/llm/models.py` — passes
- [x] Manual e2e on AppWorld via cuga-eval with both gemma and nemotron RITS endpoints; the new INFO log line surfaces the effective `temperature`/`top_p`/`max_tokens` at construction time, and runtime `Updated model parameters` log shows the TOML value (was `0.1` before this PR regardless of TOML).
- [ ] Add a small unit test that mocks `model_settings` with `temperature = 0.7, top_p = 0.9` and asserts both reach the constructed `ChatOpenAI` instance — recommend adding before merge.

## Out of scope

- Adding `top_p` to `groq`/`watsonx`/`google-genai` constructors (those use non-OpenAI SDKs and need separate verification — happy to follow up in a separate PR).
- Making `rits-restricted`'s hardcoded `top_p=0.95` configurable (intentional preservation per #404).

## Summary by CodeRabbit

* **New Features**
  * Enhanced request tuning across multiple AI model providers, with improved support for `top_p` in non-reasoning configurations.

* **Bug Fixes**
  * Applied `temperature` from model settings more consistently when loading or reusing models.
  * Ensured `max_completion_tokens` aligns with `max_tokens` for mock and cached model flows.

* **Style**
  * Added clearer logging of the final, effective sampling and token limits before creating a model instance.

### DIFF
--- a/src/cuga/backend/llm/models.py
+++ b/src/cuga/backend/llm/models.py
@@ -625,6 +625,11 @@ class LLMManager:
         api_version = self._get_api_version(model_settings, platform)
         base_url = self._get_base_url(model_settings, platform)
         http_timeout = self._get_http_timeout(model_settings)
+        logger.info(
+            f"Sampling config for {platform}/{model_name}: "
+            f"temperature={temperature}, top_p={model_settings.get('top_p', 1.0)}, "
+            f"max_tokens={max_tokens}"
+        )
         if platform == "azure":
             api_version = str(model_settings.get('api_version'))
             is_reasoning = self._is_reasoning_model(model_name)
@@ -644,6 +649,7 @@ class LLMManager:
                     timeout=http_timeout,
                     azure_deployment=model_name + "-" + api_version,
                     temperature=temperature,
+                    top_p=model_settings.get('top_p', 1.0),
                     max_tokens=max_tokens,
                 )
         elif platform == "openai":
@@ -657,6 +663,7 @@ class LLMManager:
 
             if not is_reasoning:
                 openai_params["temperature"] = temperature
+                openai_params["top_p"] = model_settings.get('top_p', 1.0)
             else:
                 logger.debug(f"Skipping temperature for reasoning model: {model_name}")
 
@@ -743,6 +750,7 @@ class LLMManager:
                 max_tokens=max_tokens,
                 model=model_name,
                 temperature=temperature,
+                top_p=model_settings.get('top_p', 1.0),
                 seed=42,
             )
         elif platform == "rits-restricted":
@@ -794,6 +802,7 @@ class LLMManager:
 
             if not is_reasoning:
                 openrouter_params["temperature"] = temperature
+                openrouter_params["top_p"] = model_settings.get('top_p', 1.0)
             else:
                 logger.debug(f"Skipping temperature for reasoning model: {model_name}")
 
@@ -826,6 +835,7 @@ class LLMManager:
             }
             if not is_reasoning:
                 litellm_params["temperature"] = temperature
+                litellm_params["top_p"] = model_settings.get('top_p', 1.0)
             else:
                 logger.debug(f"Skipping temperature for reasoning model (litellm): {model_name}")
             # Tell litellm to use the OpenAI-compatible code path without parsing
@@ -881,7 +891,10 @@ class LLMManager:
         if is_mock_llm_enabled():
             mock = clone_load_test_mock_chat_model()
             return self._update_model_parameters(
-                mock, temperature=0.1, max_tokens=max_tokens, max_completion_tokens=max_tokens
+                mock,
+                temperature=model_settings.get('temperature', 0.1),
+                max_tokens=max_tokens,
+                max_completion_tokens=max_tokens,
             )
 
         # Check if pre-instantiated model is available
@@ -889,7 +902,9 @@ class LLMManager:
             logger.debug(f"Using pre-instantiated model: {type(self._pre_instantiated_model).__name__}")
             # Update parameters for the task
             updated_model = self._update_model_parameters(
-                self._pre_instantiated_model, temperature=0.1, max_tokens=max_tokens
+                self._pre_instantiated_model,
+                temperature=model_settings.get('temperature', 0.1),
+                max_tokens=max_tokens,
             )
             return updated_model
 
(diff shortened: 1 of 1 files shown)


sami-marreed:
is 0.1 good default for top p?

Sergey-Zeltyn:
> is 0.1 good default for top p?

@sami-marreed 
1. It is 1.0, not 0.1
2. **No behavior change for existing users**. Before this PR the platform constructors didn't pass top_p at all, which causes the request body to omit the field, which causes the server (OpenAI, Azure, vLLM all behave this way) to apply its own default — also 1.0. So a user who hasn't added top_p to their TOML sees identical sampling behavior before vs after.
3. **No opinionated framework-level choice.** Per-model published recommendations vary (e.g. NVIDIA Nemotron 0.95, Gemma 0.95, OpenAI publishes nothing). Picking any non-1.0 default would mean cuga is silently truncating tokens for users on models where the vendor doesn't recommend it. 1.0 keeps that decision in the user's TOML, which is the same surface as temperature.

sami-marreed:
Did you test this on litellm internally and e.g. on external providers?

sami-marreed (approved):

PR #410, opened 2026-06-30
fix(knowledge): collection-management & reindex state-machine hardening + tile/chip UX (closes #402 #398 #397 #396 #387)

### LINKED ISSUE #387: [Bug] knowledge: HybridChunker tokenizer mismatch for litellm/watsonx-routed HF embedders

### LINKED ISSUE #396: [Bug] knowledge: hold _agent_draft_lock across _migrate_and_reindex_for_agent (engine config drift)
## Background

PR #383 introduced ``_agent_draft_lock`` to serialize all draft PATCHes for a given agent (closes engine/DB desync + cross-section + same-section LMW races). The lock wraps ``patch_draft_knowledge`` from read through save.

But ``_migrate_and_reindex_for_agent`` (called by ``POST /api/manage/knowledge/reindex_for_config``) does NOT hold the same lock.

## The race

  - User clicks Re-index → ``_migrate_and_reindex_for_agent`` reads ``live_engine._config.vector_config_hash()`` at engine.py-level
  - In parallel, another PATCH /draft/knowledge fires and mutates the engine config mid-migration
  - The migration computes ``target_collection`` against config-at-time-T0 but the engine moves to config-at-time-T1
  - Vectors land in the wrong-hash collection name; ``app_state.knowledge_config_hash`` gets promoted to a hash that doesn't match the engine's current config

## Severity

Medium — the FE pattern is one PATCH at a time per user, and the autosave abort suppresses overlap. But a power user with two tabs open OR a programmatic SDK consumer can trigger this.

## Proposed fix

Hold ``_agent_draft_lock(agent_id)`` across ``_migrate_and_reindex_for_agent``. Symmetric with how ``patch_draft_knowledge`` holds it across apply+save.

Alternative (lighter): snapshot ``vector_config_hash`` locally at the top of the migration helper, never re-read engine state mid-flight.

## Tests

  - Concurrent PATCH /draft/knowledge + POST /reindex_for_config: assert the target_collection used by the migration matches the engine's config at the moment migration STARTED, not whatever the engine is at when migration finishes.

### LINKED ISSUE #397: [Bug] frontend: draftSaveStatus survives modal close / agent switch / publish (state leaks)
## Background

The ``draftSaveStatus`` state machine ({idle, saving, saving-slow, saved, failed}) is tied to the per-PATCH lifecycle in ManagePage.tsx. PR #383 added the 25s/90s safety net and the Retry button on failure.

## The bug (audit Medium finding)

``draftSaveStatus`` never resets to ``idle`` on:

  - Modal close (Knowledge config dialog closes with a Retry-able failure still in state)
  - Agent switch (state for agent A bleeds into agent B's view)
  - Publish success (the saved-snapshot is now live; the draft pill should clear)

User-visible: a stale Retry button on a different agent's panel, or a stuck Failed pill after publish.

## Proposed fix

  - ``setDraftSaveStatus({ kind: "idle" })`` in the Knowledge modal's onClose
  - Same on the agent-switch handler
  - Same on a successful Publish callback

Plus a unit test (or a Playwright smoke) that confirms the pill clears.

## Severity

Medium — no data loss, but UX confusion and a stale "Retry" button can mis-PATCH a different agent.

### LINKED ISSUE #398: [Bug] knowledge: ReindexBusyError should map to its own FE code (reindex_busy), not collapse to reindex_failed
## Background

PR #383 added three FE error codes for triggered:false responses from ``POST /reindex_for_config``:

  - active_snapshot_missing
  - copy_failed
  - reindex_failed

## The gap

``engine.reindex(target)`` raises ``ReindexBusyError`` when uploads are in progress (existing engine machinery). The migration helper's try/except at manage_routes.py:462-468 collapses this to ``reindex_failed`` — the user sees "Re-index ran but didn't embed anything" when the real cause is "wait, an upload is in progress."

There's already a precedent for surfacing this specifically in ``routes.py:780-784`` (the user-facing /api/knowledge/reindex endpoint returns 409 with a clear message).

## Proposed fix

  - Catch ``ReindexBusyError`` separately in the migration helper.
  - Return ``{triggered: false, error: "reindex_busy"}``.
  - Frontend ERR dict: add ``reindex_busy: "Uploads in progress. Try again in a moment."``

## Severity

Medium — the failure message misleads users on a recoverable condition (just wait).

### LINKED ISSUE #402: [Bug] knowledge UX: reindex tile shows task_xxx instead of filenames + rows vanish on terminal state
## Symptom (user-reported on PR #383 QA)

> 1. The files in the Documents section showed 3
> 2. When reindexing it showed 5 documents
> 3. It didn't show in the reindex phase the documents being reindexed by name, it showed task_... and it disappeared after a while (the documents list)

### (a) Document name not shown in the reindex tile

The tile lists task rows by ``task_xxx`` instead of the file's display name. Backend ``/api/knowledge/tasks?scope=agent`` does return the ``filename`` in the file_tasks payload — see ``_ingest_inner`` at engine.py:2191 (``{filename: {filename: ..., stage: "parsed", ...}}``). The FE either isn't reading ``file_tasks[filename].filename`` or is falling back to ``task_id`` when the join fails.

### (b) Rows vanish before reindex completes

After a worker reaches terminal state, the corresponding row disappears from the tile. Either the FE filters out terminal tasks too eagerly, or the backend's polling-window query excludes completed tasks too soon (we should keep them visible until the user acknowledges or the modal closes).

## Reproduction

  - Click Re-index with a multi-doc collection
  - Watch the reindex tile rows
  - Expected: rows show document filenames, persist through completion
  - Actual: rows show ``task_xxx``, vanish on terminal state

## Likely fix locations (didn't dig deeply yet)

  - ``ManagePage.tsx`` reindex tile renderer: ensure it reads ``file_tasks[Object.keys(file_tasks)[0]].filename`` (or similar), with ``task_id`` as a last-resort label only.
  - Same renderer: stop filtering ``status in ("completed", "failed", "cancelled")`` from the visible set during an active reindex window.

## Related

Tracks alongside the documents-tab-vs-reindex-count discrepancy in the sibling issue (file separately) — both stem from the FE's task model being slightly out of sync with the backend's lifecycle.

dolev31:
Stacked on **#383** (env-presets + reindex hardening). Base is `feat/knowledge-env-presets-redesign` so the diff stays focused; GitHub will auto-retarget to `main` when #383 merges, and full CI runs then.

What started as three FE UX bugs grew (through extensive manual QA and review) into an end-to-end hardening of the knowledge **collection-management & reindex state machine** — backend + UI. No schema changes.

Closes #402
Closes #398
Closes #397
Closes #396
Closes #387

## Highlights

**One strict promotion path.** Re-index and Publish now both promote a collection to "active" only through the same deferred flip — after every worker reaches a terminal state with `n_failed == 0`, atomically under `_agent_draft_lock`, re-checking the engine hash, and persisting the result so it survives restart. Previously Publish flipped the pointer the instant `engine.reindex()` returned `"started"` (workers still running in the background), activating an empty/partial collection with no success check — the exact silent-data-loss the reindex path was hardened against.

**The invariant now enforced end-to-end:** the active pointer only ever moves to a collection that is fully built with the current embedder and verified; on any failure the old collection stays active and intact; the UI reflects exactly that truth.

### Tokenizer / chunking precision (#387 — merged from #400)
- HF-tokenizer dispatch keyed on the embedder (e5/bge), with a 16-token safety margin so watsonx/e5 stops hitting `518 > 512`.
- tiktoken-first for OpenAI/Azure routes; HF Hub as source of truth (no curated allow-list).
- `safe_max_tokens` + `recommended_chunk_tokens`; cap the chunker/splitter at `safe_max` (incl. the tiktoken re-split path).
- Strict token-measured re-split at the chunk→embedder boundary; benign transformers `N > 512` measurement warning silenced at its fire-once flag.

### Reindex tile + chip UX (#402, #398, #397)
- Tile shows real filenames from frame 0 (no `task_xxx`); `cancelled` counted as terminal so a superseded worker can't hang the tile.
- `ReindexBusyError` → its own `reindex_busy` FE code (warning toast, not a generic failure).
- `draftSaveStatus` resets correctly on publish / agent-switch; the `reindex_in_progress` 409 no longer shows "Couldn't save — Retry" — it keeps "saving" and self-heals via a bounded retry.
- `/tasks?scope=agent` spans the agent's in-flight collection so the tile shows live per-document progress during a deferred-flip reindex.

### Collection-management & reindex correctness (#396 + audit)
- **Publish** unified onto the strict deferred flip + an in-flight 409 guard (it previously bypassed `apply_knowledge_config`'s guard, so a publish mid-reindex superseded workers and double-wrote the pointer).
- **Durable flip:** a successful flip persists the vector-affecting fields + hash together (self-consistent) to draft + published, so it survives restart instead of silently reverting.
- **SOURCE stays guarded** for the whole migration (closes the upload-lost-on-flip and back-to-back-reindex windows); **`delete_document`** rejects during reindex (→409) so a deleted doc can't be resurrected; **supersede re-checked inside the insert lock**.
- **Adopt-existing-collection:** applying a config whose embedder maps to an already-built collection adopts it as active immediately (no reindex) — fixes "imported a config → no documents" and keeps engine-config ↔ active-pointer consistent.
- **Deploy fix:** the server serves `src/cuga/frontend/dist/` (tracked) while `pnpm build` writes the workspace dir — `scripts/build_frontend.sh` builds (production) + syncs, ending the recurring stale-bundle traces.

## Tests / verification
- A thorough multi-pass review drove the correctness work; all surfaced holes were fixed and re-verified.
- Knowledge unit suite green except 5 pre-existing unrelated failures (3 CLI-publish, 2 publish-sync — confirmed red on the base before this branch). New tests added for the guards, publish-defer, durability, cancelled-tally, tiktoken cap, agent-scope task span, and adopt-existing-collection.

## Known follow-ups (tracked, not data-loss)
- GC of superseded/loser agent collections (unbounded disk growth).
- Restart-during-reindex reconciliation (fire-and-forget flip lost on restart — old collection stays active).
- `knowledge_state` cross-machine import consumption (today reconnect only works on the same host, where the collection exists on disk).

## Note on #400
This branch already contains #400 (merged in), so #400 is effectively folded here. Recommend closing #400 in favor of this PR, or merging #400 first — they overlap.

## Summary by CodeRabbit

* **New Features**
  * Added clearer knowledge reindex progress with filenames, richer task status, and terminal-state handling.
  * Health and manage screens now surface embedder availability and model details.
  * Draft knowledge updates can now adopt an already-built collection and show the available document count.

* **Bug Fixes**
  * Fixed reindex and publish conflicts so non-vector changes can still save during reindex, while vector changes are correctly blocked.
  * Improved autosave retry behavior when reindexing is temporarily busy.
  * Prevented stale “saved”/“reindexing” UI states and better handled failed or cancelled tasks.


dolev31:
@sami-marreed could you please review this when you get a chance? 🙏

It hardens the knowledge **collection-management & reindex state machine** end-to-end (backend + UI) and closes #402 #398 #397 #396 #387. Strict single promotion path (Re-index + Publish both go through the deferred flip; promote only on all-workers-success, durable across restart), plus the tile/chip UX fixes and the tokenizer-precision work (folded from #400).

Stacked on #383 (merge that first; this auto-retargets to main + runs full CI then). PR body has the full breakdown and the tracked follow-ups.

dolev31:
@sami-marreed this is ready for review 🙏

**Retargeted to `main`.** Now that #383 is merged, I brought #410 up to date by merging `main` in and retargeting the base from #383's branch to `main`. The PR diff is now just #410's own work — **17 files** (knowledge backend, the two FE sources, the build script, the rebuilt served bundle, and the knowledge tests).

**Also:** closed #400 — its tiktoken/canary + safety-margin precision pass is folded into this branch (its commits are ancestors here), so it lands with #410 rather than as a separate PR.

**Conflict resolution** kept #410's refactor (unified `get_chunking_tokenizer` dispatch + tiktoken branch + 16-token safety margin, replacing the old allow-list/canary; the reindex state machine; the live progress tile; the embedder-unavailable alert) while preserving every `main`-only fix (secret-redaction list recursion, `from None` on the sanitized 500s, `isIndexConfigEquivalent` empty-model rule, `onDismissDraftSave` wiring).

**Tests:** 553 knowledge/manage unit tests pass. Heads-up — 5 `test_manage_publish_sync` / `test_knowledge_cli_publish` failures show red, but they pre-exist on `main` (something added an "Agent name is required" gate to `POST /api/manage/config` without updating those tests); they're unrelated to this PR.

dolev31:
Follow-up: fixed those 5 red tests (`4ee4122f`) — they were test-only. The 2 `test_manage_publish_sync` cases just omitted `config.agent.name` (so the legitimate publish gate 400'd); the 3 `test_knowledge_cli_publish` cases patched `_cuga_server_base_url`/`httpx` on `cuga.cli.main` instead of `cuga.cli.knowledge_cmds` where those names live. Endpoint gate unchanged. #410's suite is green now.

src/cuga/backend/server/manage_routes.py
```diff
@@ -2336,14 +2547,50 @@ async def save_manage_config_publish(request: Request, agent_id: Optional[str] =
...
                     _vec_hash,
                 )
+            elif _reindex_status == "started":
+                # CRITICAL (#1): engine.reindex returns 'started' immediately with
+                # workers in the background. Promoting the pointer now would
+                # activate an empty/still-filling collection and skip the strict
+                # n_failed==0 check — the exact silent-data-loss the Re-index path
+                # was hardened against. DEFER to the SAME strict flip: it promotes
+                # the in-memory pointer AND persists _vector_config_hash durably
+                # ONLY after all workers terminate successfully and the engine
```
sami-marreed: #c1 — Deferring the pointer flip here is the right call, but the migration block above still calls `copy_source_files` + `reindex(_new_collection)` without flagging `_old_collection` in `_reindex_in_progress`. The UI reindex path flags SOURCE for the whole deferred-flip window and only releases in `_flip_then_release_source`'s `finally`. During async publish reindex, uploads/deletes to the still-active old collection remain allowed — the same window this PR closes on the reindex side. Could we reuse the same SOURCE-flag + wrapper here?

tests/unit/test_manage_publish_sync.py
```diff
@@ -46,11 +46,12 @@ def test_publish_syncs_draft_with_published_knowledge_flags(monkeypatch):
...
+                "agent": {"name": "test-agent"},
                 "knowledge": {
                     "enabled": True,
                     "agent_level_enabled": True,
                     "session_level_enabled": False,
-                }
+                },
             }
         },
     )
```
sami-marreed, line 61: #c2 — This file covers draft/published flag sync, but I don't see coverage for publish when vector config changes and `reindex()` returns `status: "started"`. Would be great to add a test asserting: pointer is not promoted immediately, strict deferred flip is spawned with `task_ids`, and (once #c1 lands) `_old_collection` stays in `_reindex_in_progress` during the window. The reindex guard tests set a high bar — a publish equivalent would close the loop.

src/cuga/backend/server/manage_routes.py
```diff
@@ -1957,6 +2117,29 @@ async def save_manage_config_publish(request: Request, agent_id: Optional[str] =
     if app_state is None:
```
sami-marreed: #c3 — Branch looks ~3 commits behind current `main` (#430 RITS, #431 WorkspacePanel, #428 HITL). Merge should be clean, but a rebase + CI rerun on latest `main` would avoid confusion — especially since the GitHub diff vs the old base still shows RITS/README changes that won't actually land on merge.

src/cuga/backend/server/manage_routes.py
```diff
@@ -470,11 +534,33 @@ async def _deferred_reindex_complete_and_flip(
...
+    # 518>512, 4/5 succeeded, pointer flipped — search for that file's content
+    # returned nothing because its vectors never made it into the new
+    # collection while the FE banner said the embedder switch succeeded.
+    #
+    # The strict choice: stay on the old collection (where ALL files have
+    # working vectors), surface the failure visibly via task status, let
+    # the user retry or revert. The user has the old collection intact
+    # for retrieval the whole time; correctness is preserved while they
+    # fix the failing file or pick a different embedder.
+    if n_failed > 0:
```
sami-marreed: #c4 — The strict `n_failed == 0` gate here is correct, but the docstring on `_deferred_reindex_complete_and_flip` (a few lines above) still says the pointer moves "ONLY IF at least one task completed successfully". One-line docstring update would save the next reader from second-guessing the strict flip.

src/cuga/backend/server/manage_routes.py
```diff
@@ -1957,6 +2117,29 @@ async def save_manage_config_publish(request: Request, agent_id: Optional[str] =
...
+            raise HTTPException(
+                status_code=409,
+                detail={
+                    "error": "reindex_in_progress",
+                    "collections": _busy_pub,
+                    "message": "A re-index is running for this agent. Wait for it to finish before publishing.",
+                },
+            )
+
     data = await request.json()
```
sami-marreed: #c5 — `request.json()` is read again later in this handler (around line 2170). Starlette caches the body so runtime is fine, but reading once into a variable would make the flow easier to trace.

src/cuga/backend/knowledge/routes.py
```diff
@@ -740,6 +746,18 @@ async def delete_document(
...
     except DocumentNotFoundError:
         raise HTTPException(status_code=404, detail="document not found")
+    except ReindexInProgressError:
+        # Delete is rejected while this collection is being reindexed —
+        # otherwise the doc would be re-embedded into the in-flight target and
+        # resurrect after the pointer flips. 409 mirrors the upload guard shape.
+        raise HTTPException(
+            status_code=409,
+            detail={
+                "error": "reindex_in_progress",
```
sami-marreed, line 756: #c6 — Nice structured 409 here (`error: "reindex_in_progress"`). Out of scope for #398, but flagging for later: `POST /knowledge/reindex` still returns a plain string 409 on `ReindexBusyError` while the manage route returns `error: "reindex_busy"`. Not blocking, but unifying the shape would let the FE share one handler eventually.

src/cuga/backend/server/manage_routes.py
```diff
@@ -470,11 +534,33 @@ async def _deferred_reindex_complete_and_flip(
...
+    # collection while the FE banner said the embedder switch succeeded.
+    #
+    # The strict choice: stay on the old collection (where ALL files have
+    # working vectors), surface the failure visibly via task status, let
+    # the user retry or revert. The user has the old collection intact
+    # for retrieval the whole time; correctness is preserved while they
+    # fix the failing file or pick a different embedder.
+    if n_failed > 0:
+        logger.warning(
+            f"Deferred flip for {target}: {n_completed}/{len(relevant)} succeeded, "
```
sami-marreed: #c7 — Not blocking — strict flip correctly refuses to promote on partial failure. Just noting that failed reindexes can leave `kb_agent_*_<new_hash>` collections on disk with no GC. The PR body already tracks this; a one-liner in ops docs when GC lands would help operators.

src/cuga/backend/server/manage_routes.py
```diff
@@ -541,47 +632,85 @@ async def _migrate_and_reindex_for_agent(agent_id: str, live_engine: Any, live_s
...
-        # Busy flag on source so concurrent uploads hit ReindexBusyError;
+        # Flag SOURCE for the WHOLE Path A lifetime — NOT just the copy window.
+        # The active pointer still resolves to SOURCE until the deferred flip,
+        # so concurrent uploads/deletes to the active collection must be rejected
+        # until then; releasing the flag after copy (the old behavior) left a
+        # window where an upload lands in SOURCE only and is lost on flip (#7),
+        # and where a second Re-index slips the 409 guard (#9). Released in the
+        # flip wrapper's finally below (or here on copy failure / no-flip).
         # per-collection locks serialize simultaneous Re-index clicks.
         live_engine._reindex_in_progress.add(source)
```
sami-marreed: #c8 — This is the pattern I want everywhere: SOURCE stays flagged for the whole Path A lifetime, not just the copy window, and release is tied to the deferred flip's `finally`. This closes the upload-lost-on-flip and back-to-back-reindex windows cleanly.

scripts/build_frontend.sh
```diff
@@ -0,0 +1,46 @@
...
+#
+# WHY THIS EXISTS — webpack writes the built bundle to
+#   src/frontend_workspaces/frontend/dist/
+# but the FastAPI server serves
+#   src/cuga/frontend/dist/   (PACKAGE_ROOT/frontend/dist; see
+#   src/cuga/backend/server/main.py -> FRONTEND_DIST_DIR)
+# which is a SEPARATE, git-tracked copy. A bare `pnpm build` leaves the
+# served copy stale, so frontend changes silently never reach
+# `cuga start` — every "I rebuilt but still see the old behavior" trace
+# is this gap. Run THIS instead of a bare build whenever FE source changes.
```
sami-marreed: #c9 — This explains a recurring "I rebuilt but still see old behavior" class of bugs. Glad the workspace-dist vs served-dist gap is documented inline — should save future QA cycles.

sami-marreed (changes requested):
Thanks for the thorough write-up — the unified reindex/publish state machine and the regression tests give me real confidence in the core path.

I checked the five linked issues against the three-dot diff vs current `main` (not just the stale PR base). All five look addressed, and the actual merge delta is clean — no unrelated RITS/README churn will land. Scope did grow beyond the original tile UX bugs, but it's documented and stays in the knowledge domain.

Two items block merge for me: publish-path SOURCE guarding (same pattern you already nailed on the UI reindex path) and a rebase onto latest `main` (#430, #431, #428) with CI rerun. Everything else is nits or follow-up.

**Tagged items**

- **#c1** — Publish migration doesn't flag the old collection busy during deferred flip
- **#c2** — Missing test for publish deferred-flip path
- **#c3** — Branch is behind `main`; rebase before merge
- **#c4** — Stale docstring on strict deferred flip
- **#c5** — Duplicate `request.json()` in publish handler
- **#c6** — `/knowledge/reindex` 409 shape differs from manage route
- **#c7** — Orphan target collections after strict-flip refusal (follow-up)
- **#c8** — SOURCE busy-flag pattern on the UI reindex path is exactly right
- **#c9** — `build_frontend.sh` closes a recurring deploy gap

---

**When you respond, please tie your answers to each tag** so we can track what changed:

- **#c1** — Publish migration doesn't flag the old collection busy during deferred flip
- **#c2** — Missing test for publish deferred-flip path
- **#c3** — Branch is behind `main`; rebase before merge
- **#c4** — Stale docstring on strict deferred flip
- **#c5** — Duplicate `request.json()` in publish handler
- **#c6** — `/knowledge/reindex` 409 shape differs from manage route
- **#c7** — Orphan target collections after strict-flip refusal (follow-up)
- **#c8** — SOURCE busy-flag pattern on the UI reindex path is exactly right
- **#c9** — `build_frontend.sh` closes a recurring deploy gap

Reply in the inline threads where you can, or summarize in your next commit message / PR comment using the same tags (`#c1`, `#c2`, …). That makes re-review much faster.

src/cuga/backend/server/manage_routes.py
sami-marreed: #c1 — Deferring the pointer flip here is the right move.

One gap though: the migration code above runs `copy_source_files` + `reindex` on the new collection without marking the **old** collection as busy.

On the UI reindex path you already solve this — SOURCE stays in `_reindex_in_progress` until the flip finishes. During a publish reindex, someone can still upload or delete docs on the old (still-active) collection. That's the same race this PR closes elsewhere.

Could we reuse the same SOURCE-flag + release wrapper here?

tests/unit/test_manage_publish_sync.py
sami-marreed, line 61: #c2 — These tests cover draft/published flag sync nicely.

What's missing: a case where publish changes vector config and `reindex()` comes back with `status: "started"`.

Would be great to check that:
- the pointer doesn't flip immediately
- the deferred flip gets spawned with `task_ids`
- once #c1 is in, the old collection stays busy during the window

The reindex guard tests are thorough — a publish version would round this out.

src/cuga/backend/server/manage_routes.py
sami-marreed: #c3 — Small housekeeping: the branch is about 3 commits behind `main` (#430, #431, #428).

Merge should be fine, but a rebase + fresh CI run would help — the diff against the old base still shows some RITS/README noise that won't actually land once you're on current `main`.

src/cuga/backend/server/manage_routes.py
sami-marreed: #c4 — The strict `n_failed == 0` check here is correct.

Tiny docstring drift: `_deferred_reindex_complete_and_flip` still says the pointer moves when "at least one task completed successfully." The code now requires **all** tasks to succeed.

A one-line docstring tweak would save the next person some head-scratching.

src/cuga/backend/server/manage_routes.py
sami-marreed: #c5 — Minor cleanup: `request.json()` gets called again further down in this handler (~line 2170).

Starlette caches it so nothing breaks at runtime — just reads a bit cleaner if you parse once into a variable.

src/cuga/backend/knowledge/routes.py
sami-marreed, line 756: #c6 — Love the structured 409 here (`error: "reindex_in_progress"`).

Not blocking and out of #398 scope, but worth a follow-up: `POST /knowledge/reindex` still returns a plain string on `ReindexBusyError`, while the manage route returns `error: "reindex_busy"`. Unifying those would let the frontend share one error handler.

src/cuga/backend/server/manage_routes.py
sami-marreed: #c7 — Not blocking — refusing to promote on partial failure is the right call.

Just noting for later: a failed reindex can leave orphan `kb_agent_*_<hash>` collections on disk. You already track GC in the PR body; a line in ops docs when that lands would help.

src/cuga/backend/server/manage_routes.py
sami-marreed: #c8 — This is exactly the pattern we want.

SOURCE stays flagged for the whole migration — not just during the file copy — and only gets released when the deferred flip finishes. That closes the upload-lost-on-flip and double-reindex windows cleanly. Nice.

scripts/build_frontend.sh
sami-marreed: #c9 — This script addresses a real pain point.

Webpack writes to one dist folder, the server reads from another — so "I rebuilt but nothing changed" keeps coming up. Documenting that gap here should save future QA cycles.

sami-marreed (changes requested):
Thanks for putting this together — the reindex/publish state machine work is solid, and the test coverage on the UI reindex path gave me real confidence.

I checked all five linked issues against what's actually landing on `main`. They all look covered, and nothing unrelated sneaks in on merge.

Two things I'd like fixed before we merge: the publish path should guard the old collection the same way reindex does, and a quick rebase onto latest `main` would be good. Everything else is small stuff or can wait.

**Tagged items**

- **#c1** — Publish should flag the old collection as busy during reindex
- **#c2** — Add a test for the publish + deferred-flip path
- **#c3** — Rebase onto latest `main` before merge
- **#c4** — Update the deferred-flip docstring to match strict mode
- **#c5** — Read `request.json()` once in the publish handler
- **#c6** — Align 409 error shape on `/knowledge/reindex` (follow-up)
- **#c7** — Orphan collections after failed reindex (follow-up)
- **#c8** — The SOURCE busy-flag pattern on reindex is exactly right
- **#c9** — `build_frontend.sh` fixes a real deploy footgun

---

**When you respond, please tie your answers to each tag** so we can track what changed:

- **#c1** — Publish should flag the old collection as busy during reindex
- **#c2** — Add a test for the publish + deferred-flip path
- **#c3** — Rebase onto latest `main` before merge
- **#c4** — Update the deferred-flip docstring to match strict mode
- **#c5** — Read `request.json()` once in the publish handler
- **#c6** — Align 409 error shape on `/knowledge/reindex` (follow-up)
- **#c7** — Orphan collections after failed reindex (follow-up)
- **#c8** — The SOURCE busy-flag pattern on reindex is exactly right
- **#c9** — `build_frontend.sh` fixes a real deploy footgun

Reply in the inline threads where you can, or summarize using the same tags (`#c1`, `#c2`, …). That makes re-review much faster.

sami-marreed:
@dolev31 please resolve conflicts

dolev31:
@sami-marreed all your comments are addressed and the branch is rebased onto current main — ready for another look 🙏

**Rebase (#c3):** while this was open, #429 split `manage_routes.py` into the `manage_routes/` package. I re-homed #410's reindex/publish state machine onto the new submodules (`knowledge_reindex.py`, `config_routes.py`, `knowledge_routes.py`) and merged main — **CONFLICTING → MERGEABLE, CI green.**

**Your comments**
- **#c1** ✅ Publish now keeps the OLD active collection in `_reindex_in_progress` for the whole migration+embed window (mirrors the Re-index path), released only in the deferred-flip wrapper's `finally` — uploads/deletes can't be lost at flip.
- **#c2** ✅ Added `test_publish_defers_flip_when_reindex_started`: asserts the pointer isn't promoted immediately, the strict flip is spawned with `task_ids`, and the old collection stays busy during the window.
- **#c3** ✅ Rebased onto main (above).
- **#c4** ✅ Docstring fixed — the flip promotes only if **all** tasks succeed (strict `n_failed==0`).
- **#c5** ✅ Moot in the #429 refactor — the publish handler reads `request.json()` once.
- **#c6** ⏭️ Left as the follow-up you flagged (unify `POST /knowledge/reindex`'s plain-string 409 with the structured shape) — out of #398 scope.
- **#c7** ⏭️ Orphan-collection GC stays a tracked follow-up; ops-docs line when GC lands.
- **#c8 / #c9** 🙏 The SOURCE-flag-for-the-whole-window pattern + the build_frontend.sh doc are preserved through the port.

**CodeRabbit (all Major addressed):** terminalize timed-out reindex tasks (real self-heal); redact the *actual* configured key (not just known shapes); fallback splitter uses the token cap; delete/reindex TOCTOU (snapshot+flag under the collection lock + in-lock delete re-check); reindex-vs-publish serialized under `agent_draft_lock`; missing-task-record → refuse promotion; adopt path now asserts the durable persist; source-string durability tests replaced with behavioral ones.

**Verification:** 564 knowledge/manage unit tests pass, full CI green, ruff clean, FE rebuilt. I also ran an adversarial re-review of the ~1000-line re-home (lock lifecycle, flag leak/double-free, deadlock, silent revert-to-main) — **0 correctness/security findings**. Only cosmetic follow-ups remain.

dolev31:
@sami-marreed friendly re-review nudge 🙏 Two commits landed since my last update, both green on CI (`099abe9b`):

1. **CodeRabbit's re-review addressed** — 10 fixes (most notably: publish no longer persists the new `_vector_config_hash` while an async reindex is still "started" — the strict deferred flip owns promotion after all workers succeed; publish/reindex now re-check the in-flight guard *inside* `agent_draft_lock`; fail-closed secret stripping; embedder-probe off the event loop; rebuild the embedder client on a key/base-url-only fix). Declined only the RUF022 `__all__`-sort nitpick (this repo doesn't enforce it).
2. **A small cleanup** — single release point for the publish old-collection busy flag.

564 knowledge/manage unit tests pass, full CI green, and I ran an adversarial re-review of the re-home (0 correctness/security findings). Whenever you have a moment 🙏

scripts/build_frontend.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
```
sami-marreed: @dolev131 REMOVE THIS

dolev31: Hi @sami-marreed — happy to remove it, just want to be sure I catch your intent, since your other note (#c9) flagged this same script as genuinely useful ("addresses a real pain point… glad the workspace-dist vs served-dist gap is documented inline").

Do you mean:
- (a) drop `scripts/build_frontend.sh` from the repo entirely (I'd move the dist-gap explanation into a short README/CONTRIBUTING note so the knowledge isn't lost), or
- (b) keep the script but take it out of *this* PR's scope (it's unrelated to the reindex work), or
- (c) something specific on line 1?

It's also committed on #468, so I'll keep both branches consistent with whatever you prefer. Thanks! 🙏

dolev31:
Thanks for the thorough review, @sami-marreed! 🙏 The code items were addressed in commits *after* your pass (which is why the threads now show as outdated). Quick map:

- **#c1** (old collection not flagged busy during publish reindex) — ✅ `manage_routes/config_routes.py` now adds `_old_collection` to `_reindex_in_progress` for the whole migration + embed window and releases it in the deferred flip's `finally` (`_pub_flip_then_release`) plus an outer safety `finally`. Mirrors the Re-index path exactly, as you suggested.
- **#c2** (missing publish + reindex test) — ✅ added `test_publish_defers_flip_when_reindex_started` in `test_manage_publish_sync.py`: asserts the pointer is **not** promoted immediately, the deferred flip is spawned with `task_ids`, and the old collection stays busy during the window.
- **#c4** (docstring drift) — ✅ `deferred_reindex_complete_and_flip` docstring now states the strict "promotes … ONLY IF ALL tasks completed successfully (no failures)" gate.
- **#c5** (double `request.json()`) — ✅ read once into `data`.
- **#c3** (rebase) — ✅ merged latest `main`; the branch is now 0 behind and merges cleanly, so the earlier RITS/README diff noise is gone. Fresh CI is running.
- **#c6 / #c7** (unify the 409 error shape / orphan-collection GC) — noted as good follow-ups, tracked for a later PR (out of this one's scope).
- **build_frontend.sh** — replied in the thread with a quick clarifying question.

Re-requesting your review — thanks again!

scripts/build_frontend.sh
sami-marreed: remove entirely @coderabbitai validate please we already have build.sh

sami-marreed:
will approve after final removal of file i commented on a thread

scripts/build_frontend.sh
dolev31: Done ✅ — removed `scripts/build_frontend.sh` from this PR (f16e10c1) and from #468 (ed3e8b8b), since `src/frontend_workspaces/frontend/build.sh` already covers it (rm dist → prod build → copy into `cuga/frontend/dist`). Also repointed the two doc references. Thanks for catching the duplication! 🙏

dolev31:
Hi @sami-marreed 👋

Done — `scripts/build_frontend.sh` is removed (`f16e10c1` here, and `ed3e8b8b` on #468), since `src/frontend_workspaces/frontend/build.sh` already covers it. That was the last outstanding ask, so **#410 is ready to merge on your approval** — CI is just re-running on the deletion push (it was fully green beforehand, and this is a pure file removal).

Separately, whenever you get a window: **#468 (inline citations & sources) is an important one for us** — it's the customer-facing citations feature, it's green, and it's waiting on your first review. Would really appreciate you prioritizing that one too. 🙏

Thanks a lot!

sami-marreed (approved):

PR #414, opened 2026-06-30
fix(policy): json_schema for conflict-resolution structured output (gpt-oss/Watsonx)

segevshlomovIBM:
## Bug fix

Fixes #411

### Summary
`PolicyAgent._resolve_nl_trigger_conflicts` called `with_structured_output(PolicyConflictResolution)` without `method=`, defaulting to function/tool-calling. On `openai/gpt-oss-120b` (Watsonx) the model often emits no tool call → returns `None` → `AttributeError` on `.matched_policy_index` → NL-trigger policies silently never match. Fix: pass `method="json_schema"` (enforced server-side by Watsonx; matches the existing OutputFormatter in `enactment.py`). The rest of the graph already uses `response_format=json_schema` via `BaseAgent.get_chain`, so it is unaffected.

### Testing
- [x] Verified fix locally; tests pass

**Checks performed — live on `openai/gpt-oss-120b` (IBM Watsonx):**
- Policy: the 2 failing conflict-resolution tests now pass; all 7 originally-failing policy tests pass in isolation after the fix.
- Full task-decomposition graph: `accurate_test` 3/3, `balanced_test` 3/3.
- AB / stability suite: 10/10 ×3 (100%).
- Mock unit tier (`run_tests.sh unit_tests`): ~535 passed / 0 failed.
- `ruff check` clean on the changed file.

### Related (follow-ups, not fixed here)
- #412 — policy e2e tests flaky with `RuntimeError: Event loop is closed` when run as a suite (test-harness async teardown; each passes in isolation).
- #413 — no live test coverage for the browser sub-graph & feature-flagged paths on gpt-oss/Watsonx.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of conflict resolution for natural-language triggers, reducing cases where responses could be misparsed or fail during processing.

Issue #411, opened 2026-06-30
[Bug]: Policy NL-trigger conflict resolution returns None on gpt-oss-120b (Watsonx) — policies never match


segevshlomovIBM:
### What happened

On `openai/gpt-oss-120b` (IBM Watsonx), policy **NL-trigger conflict resolution silently fails** — natural-language-triggered policies never match.

`PolicyAgent._resolve_nl_trigger_conflicts` (`src/cuga/backend/cuga_graph/policy/agent.py:624`) calls `self.llm.with_structured_output(PolicyConflictResolution)` **without** a `method=` argument, so it defaults to function/tool-calling. `gpt-oss-120b` frequently returns **no tool call**, so the call returns `None`; the next line (`result.matched_policy_index`, agent.py:668) raises `AttributeError`, which is caught and falls back to "no match".

**Expected:** the matching policy is selected.
**Actual:** `policy_matched = False` for every NL-trigger policy.

One-line fix (matches the existing `OutputFormatter` at `enactment.py:924`; Watsonx enforces json_schema server-side):

```python
# policy/agent.py:624
structured_llm = self.llm.with_structured_output(PolicyConflictResolution, method="json_schema")
```

The rest of the agent graph is unaffected, since it routes through `BaseAgent.get_chain` (`response_format=json_schema`, which Watsonx enforces).

### How to reproduce

```bash
# Env: AGENT_SETTING_CONFIG=settings.watsonx.toml, MODEL_NAME=openai/gpt-oss-120b,
#      plus valid WATSONX_URL / WATSONX_APIKEY / WATSONX_PROJECT_ID
uv run pytest src/cuga/backend/cuga_graph/policy/tests/test_e2e_healthcare_family_claims.py::test_healthcare_family_claims_e2e_with_tools
```

Fails on `assert metadata.get("policy_matched")`; logs show the `AttributeError` below. Passes after the one-line fix.

### Environment

- OS: macOS (reproducible anywhere targeting Watsonx)
- CUGA: 0.2.28 (commit 6efa7f4b)
- Python: 3.12.7
- langchain-ibm: 1.0.7
- Model: `openai/gpt-oss-120b` on IBM Watsonx (`settings.watsonx.toml`)

### Logs, screenshots, or config

```
cuga.backend.cuga_graph.policy.agent:_resolve_nl_trigger_conflicts:668 -
❌ Error in LLM conflict resolution: 'NoneType' object has no attribute 'matched_policy_index'
[...]

PR #415, opened 2026-06-30
test(policy): process-isolate policy e2e tests to fix 'Event loop is closed' flakiness

### LINKED ISSUE #412: [Bug]: Policy e2e tests flaky — "RuntimeError: Event loop is closed" when run as a suite
### What happened

Running the full policy e2e suite, several live `test_e2e_*` tests fail with `RuntimeError: Event loop is closed` (during the async `call_model` task). The **set of failing tests varies run-to-run**, and **each failing test passes when run in isolation**. This is a test-harness async-teardown / isolation issue (async LLM calls outliving the per-test event loop), not a product bug.

Observed: full dir → ~6 nondeterministic failures; the same tests run individually → pass (e.g. one run failed `nl_trigger_conflict` + 5 `output_formatter`/`playbook`/`tool_enrichment` tests; a prior run failed a different subset).

### How to reproduce

```bash
# Env: AGENT_SETTING_CONFIG=settings.watsonx.toml, MODEL_NAME=openai/gpt-oss-120b
uv run pytest src/cuga/backend/cuga_graph/policy/tests/        # ~6 "Event loop is closed" failures (set varies)

# the same tests pass alone:
uv run pytest src/cuga/backend/cuga_graph/policy/tests/test_e2e_output_formatter.py::test_e2e_output_formatter_sdk_integration
```

### Environment

- OS: macOS
- CUGA: 0.2.28 (commit 6efa7f4b)
- Python: 3.12.7, pytest-asyncio
- Model: `openai/gpt-oss-120b` on IBM Watsonx (likely model-independent — this is async-teardown, not a model issue)

```
RuntimeError: Event loop is closed
  During task with name 'call_model' ...
```

Suggested fix: function-scoped event loops for these tests, or await/cancel pending async tasks in teardown.

### Checklist

segevshlomovIBM:
## Bug fix

Fixes #412

### Summary
Policy e2e tests flaked with `RuntimeError: Event loop is closed` when run as one pytest session: `LLMManager` caches async clients (e.g. `ChatWatsonx`'s httpx client) bound to the event loop active at creation, and pytest's function-scoped loops reuse a cached client on a later test's loop. Each test passes in isolation; the failing set varied run-to-run.

Fix: run the policy tests **process-per-test** in `src/scripts/run_tests.sh` via a new `run_pytest_isolated` helper — one `pytest` subprocess per test (fresh interpreter / loop / singletons), with a single retry to absorb transient live-LLM nondeterminism.

### Testing

- **Before** (one session): 6 failed (all `Event loop is closed`) / 51 passed; failing set varied run-to-run.
- **After** (process-isolated): full policy dir 56/57 passed; the 1 was a transient accuracy flake (`healthcare_family_claims`) that passes on re-run — now absorbed by the retry. The 6 event-loop offenders (`output_formatter`×3, `playbook_guidance`, `tool_enrichment`, `nl_trigger_conflict`) all pass.

Note: a fully-green policy run also requires the #411 fix (PR #414).

## Summary by CodeRabbit

* **Tests**
  * Improved reliability of the test runner by discovering test nodes first and then executing each test in an isolated pytest subprocess.
  * Added an automatic single retry for any individual test that fails; the overall run fails only if a test fails twice.
  * Treats “no tests collected” during discovery as a successful run.
  * Updated policy-related test execution to use the isolated per-test-node flow for more consistent results.

### DIFF
--- a/src/scripts/run_tests.sh
+++ b/src/scripts/run_tests.sh
@@ -75,8 +75,24 @@ run_pytest_with_e2b() {
 # transient live-LLM nondeterminism. See issue #412.
 run_pytest_isolated() {
     echo "Running $* with process-per-test isolation + retry (issue #412)..."
-    local nodes
-    nodes=$(uv run pytest "$@" --collect-only -q -p no:cacheprovider 2>/dev/null | grep '::')
+    local nodes collect_output collect_ec
+    if collect_output=$(uv run pytest "$@" --collect-only -q -p no:cacheprovider 2>&1); then
+        collect_ec=0
+    else
+        collect_ec=$?
+    fi
+    # pytest exit code 5 = no tests collected (not a failure); any other non-zero is a
+    # collection/import error and must NOT be silently treated as "no tests".
+    if [ "$collect_ec" -eq 5 ]; then
+        echo "No tests collected for $* (treating as success)"
+        return 0
+    fi
+    if [ "$collect_ec" -ne 0 ]; then
+        echo "$collect_output"
+        echo "❌ Test collection failed for $* (exit $collect_ec)! Exiting..."
+        exit 1
+    fi
+    nodes=$(printf '%s\n' "$collect_output" | grep '::' || true)
     if [ -z "$nodes" ]; then
         echo "No tests collected for $* (treating as success)"
         return 0
--- a/src/scripts/run_tests.sh
+++ b/src/scripts/run_tests.sh
@@ -67,6 +67,39 @@ run_pytest_with_e2b() {
     fi
 }
 
+# Helper: run each collected test in its OWN pytest subprocess (process-per-test isolation),
+# retrying a failure once. Live async e2e tests (e.g. policy/tests) cache LLM/HTTP clients bound
+# to the event loop active at creation; pytest's function-scoped loops mean a cached client gets
+# reused on a later test's loop -> "RuntimeError: Event loop is closed" (nondeterministic, passes
+# in isolation). A fresh interpreter per test avoids the cross-loop reuse; the single retry absorbs
+# transient live-LLM nondeterminism. See issue #412.
+run_pytest_isolated() {
+    echo "Running $* with process-per-test isolation + retry (issue #412)..."
+    local nodes
+    nodes=$(uv run pytest "$@" --collect-only -q -p no:cacheprovider 2>/dev/null | grep '::')
+    if [ -z "$nodes" ]; then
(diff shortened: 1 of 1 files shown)


segevshlomovIBM:
@coderabbitai this was addressed in e1de9b29 (the review above was generated against 4544f6b, before the fix). `run_pytest_isolated` now captures pytest's `--collect-only` exit code: only exit `5` (no tests collected) is treated as success; any other non-zero (collection/import error) prints the output and exits 1, before grepping node IDs. Verified: happy path → exit 0 / 57 nodes; a missing target → exit 4 → correctly fails. Please re-review.

haroldship (approved):
lgtm

sami-marreed:
### ⚠️ CI runtime concern

The process-per-test isolation works, but it has a real cost: the latest `Tests / Unit Tests` run took **~35m**. The policy dir has **57 test nodes**, and each now spawns its own `uv run pytest` subprocess, so the heavy import/startup tree (`cuga` + langchain + watsonx) is paid ~58× instead of once. The live-LLM time is unchanged — this is pure subprocess/startup overhead.

This is acceptable as a short-term fix to unblock #412, but **35m is too slow to keep long-term.**

**Future direction:** move everything back to a single `pytest` command (one session) once we land the proper in-process fix — e.g. an autouse fixture that resets `LLMManager`'s cached event-loop-bound async clients between tests, instead of relying on a fresh interpreter per test. That removes the cross-loop reuse at the source and lets us drop process isolation entirely.

Tracking this as a follow-up to #412.

sami-marreed (changes requested):
Let's close this

haroldship:
> **35m is too slow to keep long-term.** [...] move everything back to a single `pytest` command (one session) once we land the proper in-process fix — e.g. an autouse fixture that resets `LLMManager`'s cached event-loop-bound async clients between tests [...] Tracking this as a follow-up to #412.

Agreed on all points. The process-per-test isolation is intentionally a short-term unblock for #412's suite-wide async-teardown flakiness, and the ~58× import/startup cost is the price of a fresh interpreter per node — real, and not something to keep permanently.

The in-process fix (autouse fixture that resets `LLMManager`'s cached loop-bound async clients between tests, removing the cross-loop reuse at the source) is the right long-term direction; once that lands we drop `run_pytest_isolated` and go back to a single `pytest` session. Tracked as a follow-up to #412 so the runtime regression doesn't get lost.

Issue #312, opened 2026-06-04
[Bug / Unexpected Behavior]: CodeAgent invokes undefined / non-registered tool functions


Sergey-Zeltyn:
### What happened

The CodeAgent (CUGA-lite) sometimes emits Python code that calls function names which are not present in the active tool registry — i.e., not in the Shortlister's `Current Available Tools` list and not returned by any prior `find_tools `call. The code fails at execution with `NameError`, costing a turn and often leading to a repeat hallucination on retry.

Examples for **AppWorld** traces: `phone_resolve_contact_by_name`, `phone_resolve_contact`, `supervisor_show_addresses_addresses_get`, `supervisor_show_payment_cards_payment_cards_get`, `gmail_send_email_with_attachment`, `phone_get_contacts_by_relationship_relationship_get`

Model: **GPT 4.1**

Possible approaches:
- Prompt improvement
- Runtime AST guard

### How to reproduce

- Run AppWorld `compare` with CUGA-lite  
- Run comparison analytics and check summary or inspect failing turns for NameError and compare the called identifier against the turn's tool registry.

### Environment

Mac

### Checklist

- [x] I searched existing issues and this doesn’t look like a duplicate

haroldship:
PR #416 lands a mitigation for this failure mode in the local executor: when executed code raises a `NameError` on a name that is *called like a function*, the error is augmented with a `[tool-name correction]` hint — the closest real tool names (difflib, cutoff 0.6 so cross-app junk is filtered), a caution that lookalike names can perform different actions (delete vs get), and a directive not to retry invented names; when nothing similar is loaded, the hint covers both possibilities (a lost agent-written helper → re-include its definition; a guessed tool → re-query `find_tools`).

Following review feedback on the PR, the call detection is AST-based rather than a text search: only real `Call` nodes count (names inside string literals or comments never trigger it), and the hint is suppressed entirely when the code itself defines the name (`def` / class / assignment / import) — a NameError on the agent's own helper or variable keeps its bare traceback, since the right fix there is repairing the code, not hunting for a tool.

[...]

PR #389, opened 2026-06-28
fix: replace loguru %s/%d format strings with f-strings (closes #388)

Maqbool61:
## What
Sweeps all `logger.X("%s", arg)` calls and replaces them with f-strings, fixing logs that were printing literal `%s` instead of values.

## Changes
- 22 files changed, 88+ lines fixed
- All `%s`/`%d` logger format strings replaced with f-strings
- Dict `.get()` calls inside f-strings use single quotes to avoid quote conflicts

## Verification
```bash
# No %s/%d logger calls remain
grep -rn 'logger\.(info|warning|error|debug|critical)([^)]*"%[sd]' src/cuga/
# → no output ✅

# All files compile clean
find src/cuga -name '*.py' | xargs python3 -m py_compile
# → no errors ✅
```

Closes #388

## Summary by CodeRabbit

* **Chores**
  * Standardized many log messages across backend, CLI, server, knowledge, and demo tools to use clearer f-string formatting.
  * Improved error, warning, and info logs in startup, reload, snapshot, policy, and search flows.
  * Added one extra startup log for the MCP server address.
  * No user-facing behavior, control flow, or APIs were changed.

PR #416, opened 2026-06-30
fix(cuga-lite): don't stop mid-task after stating a plan; warn on suspicious tool arguments; correct made-up tool names

haroldship:
## What this PR fixes

Three robustness fixes for CugaLite, described in plain words with real examples. Independent of issue #272.

### 1. The agent could stop mid-task right after stating a plan

**Problem.** Sometimes the model replies with a short plan instead of code — and the run treats that plan as the final answer.

**Example (observed).** The agent replies:

> "We need to search student_loan app."

…and stops. That sentence is returned to the user as the answer; the search it announced never happens.

**Fix.** Planning statements like this are now recognized (a deterministic check for the observed patterns, plus example-guided hardening of the existing auto-continue LLM classifier), and the agent keeps working instead of stopping.

The check is careful about the opposite mistake: a plan that also asks the user for something —

> "Ok I will fetch the information, but first I require your ID"

— is **not** auto-continued, because the user has to reply. (Second-person guard, added after review.)

The broader redesign of this decision (autonomous vs interactive, deferral handling) is tracked on #445.

### 2. Made-up tool names caused silent retry loops

**Problem.** Models sometimes invent plausible-looking tool names instead of using the exact names `find_tools` returned. The resulting `NameError` carried no guidance, so the agent kept inventing variants until the step limit.

**Example (observed, gpt-oss).** The agent calls `mondial_geo_get_countries_countries_get(...)` — a name it constructed. Before: bare `NameError`, repeated re-fabrication, task dies at the step limit. Now the error says:

> [tool-name correction] 'mondial_geo_get_countries_countries_get' is NOT an available tool — … Did you mean one of: mondial_geo_get_top_country_by_gdp_and_agriculture, …? Similarly named tools can do very different things (e.g. delete vs get) …

and the agent recovers in one step.

**Precision (after review).** The check is AST-based: a name inside a string like `var = " agent_1("` can't trigger it, and a helper function the agent wrote itself keeps its plain `NameError` (the right fix there is repairing the helper, not hunting for a tool). When no similar tool exists, the message covers both possibilities instead of assuming a tool was meant.

Tool-name hallucination itself is tracked in #312.

### 3. Suspicious tool arguments were invisible

**Problem.** The sandbox calls tools directly, bypassing schema validation — so a malformed argument reaches the registry with no earlier signal.

**Example.** The agent passes `amount="5"` (a string) where the tool wants a number. Now the log shows:

> [arg-warning] venmo_send_money: 'amount' looks malformed — stringized number; value='5'. Forwarded unchanged (no coercion; see arg_warning.py).

**Fix.** Detection-only: the warning logs the suspect shape and always forwards the argument unchanged. Mining of past eval runs showed these shapes don't currently cause agent-visible failures, so auto-fixing was deliberately rejected. Examples and evidence: #452.

### Removed during review: require-tool-call-before-final guard

An earlier commit added an optional guard that refused a final answer until at least one code block had executed (aimed at "confident answer with zero tool calls"). Review found it enforces the wrong thing: `what can you do?`, skills-without-tools, and follow-up questions would loop to the step limit, and any trivial code block (`x = 1`) satisfied it. It has been **removed from this PR**; the underlying problem — the agent answering from general LLM knowledge instead of tools — is now tracked comprehensively in #451, including a configurable answer-source policy.

## Testing

- `test_nl_auto_continue_classifier.py`, `test_shared_call_model.py`, `test_code_executor.py`, `test_arg_warning.py` — all pass.
- Full `cuga_lite` + `cuga_agent_core` suites: 414 passed; the only failures are 5 pre-existing ones unrelated to this PR (3 e2b-package-not-installed, 2 `workspace_fs`), verified identical with this PR's changes stashed.
- `ruff check` and `ruff format` clean.
- AppWorld test-easy rerun: pending (after this review round).

## Summary by CodeRabbit

* **New Features**
  * Added smarter guidance when tool or code execution hits a missing-name error, including hints for likely matches or how to re-check available tools.
  * Added diagnostic warnings for suspicious tool arguments, helping surface likely mismatched inputs without changing how calls run.
  * Improved natural-language auto-continue detection for short planning-style messages, reducing unnecessary model calls.

* **Bug Fixes**
  * Improved error messages for tool-call and execution issues.
  * Added safeguards to avoid false correction hints in non-tool, non-call cases.


sami-marreed:
requires testing on appworld again and lets get help of @Sergey-Zeltyn on groq

haroldship:
> requires testing on appworld again and lets get help of @Sergey-Zeltyn on groq

test-easy results on litellm gpt-oss-120b:

Evaluation Report
=================


Summary
-------

  Pass@1             28/43 (65.1%)
  Total Tokens       9,222,914
  Avg Tokens/Task    214,486.4
  Total LLM Calls    470
  Avg LLM Calls/Task 10.9
  Total Duration     7119.7s
  Avg Duration/Task  165.6s


Per-Task Results
----------------

  Task       R      Tokens     Cost    LLM       Cache   Duration  Steps
  ──────────────────────────────────────────────────────────────────────
  e775c78_1  ✓     134,240  $0.0000      8           0      95.9s     13
  07bb666_1  ✗     219,363  $0.0000     12           0     149.4s     25
  9aae7da_1  ✓      44,881  $0.0000      3           0      26.1s     13
  245cb43_1  ✓     152,499  $0.0000      9           0     135.1s     20
  4d12842_1  ✓     223,018  $0.0000     12           0     204.6s     18
  81be677_1  ✓      46,496  $0.0000      3           0      54.9s     13
  365e0a3_1  ✓     109,608  $0.0000      8           0     157.8s     13
  ba46d91_1  ✗      78,580  $0.0000      7           0      95.5s     13
  eb5ad85_1  ✓     115,750  $0.0000      7           0      80.8s     13
  dbc0276_1  ✗     383,922  $0.0000     18           0     284.3s     20
  5e27cd7_1  ✓     324,745  $0.0000     19           0     264.5s     33
  6588a51_1  ✗     199,099  $0.0000     10           0     143.4s     15
  5238afc_1  ✗     255,649  $0.0000     13           0     266.5s     18
  9bf2c8a_1  ✓     265,688  $0.0000     13           0     213.5s     23
  7d26579_1  ✓     208,271  $0.0000     10           0     124.7s     27
  0d22252_1  ✓      92,696  $0.0000      7           0      95.0s     18
  3650990_1  ✗     100,629  $0.0000      8           0     162.4s     15
  2e9b91e_1  ✓     195,148  $0.0000     11           0     244.9s     25
  baeb104_1  ✗   2,311,692  $0.0000     52           0    1268.2s    120
  6a5e690_1  ✓      13,716  $0.0000      2           0       9.7s     15
  f691597_1  ✓           0  $0.0000      0           0       0.0s     13
  16be9ce_1  ✗      46,825  $0.0000      3           0      36.6s     15
  277d81d_1  ✓           0  $0.0000      0           0       0.0s     20
  98d2608_1  ✓      82,319  $0.0000      9           0      98.0s     40
  fd1f8fa_1  ✓           0  $0.0000      0           0       0.0s     23
  29a7b7e_1  ✗     180,673  $0.0000     13           0     186.0s     35
  21abae1_1  ✓       6,967  $0.0000      1           0      13.7s     13
  5a83b05_1  ✓      76,764  $0.0000      9           0     113.9s     25
  cef9191_1  ✓     204,491  $0.0000      9           0      77.2s     13
  afc4005_1  ✓     220,969  $0.0000     15           0     149.4s     30
  425a494_1  ✗      63,769  $0.0000      4           0      39.7s     13
  a30375d_1  ✗   1,386,987  $0.0000     51           0     699.2s     90
  09b0ee6_1  ✓     144,499  $0.0000      8           0      80.3s     13
  7847649_1  ✓      76,522  $0.0000     17           0     134.9s     28
  552869a_1  ✓      55,175  $0.0000      5           0      86.5s     13
  024c982_1  ✓     301,030  $0.0000     20           0     277.6s     35
  13547f5_1  ✓      76,122  $0.0000      7           0     131.3s     13
  1150ed6_1  ✗      53,924  $0.0000      3           0      39.4s     13
  31dc501_1  ✗     115,179  $0.0000     11           0     127.2s     23
  59fae45_1  ✓     151,453  $0.0000      8           0     239.0s     18
  166f4ff_1  ✗     219,645  $0.0000     13           0     160.6s     28
  dac78d9_1  ✓      51,163  $0.0000      5           0      54.8s     13
  f3f60f0_1  ✗     232,748  $0.0000     27           0     297.1s     45


Difficulty Breakdown
--------------------

Diff  Tasks    Pass@1      Tokens    Tok/Task  LLM Calls   LLM/Task   Duration   Dur/Task
─────────────────────────────────────────────────────────────────────────────────────────
1        43     65.1%   9,222,914   214,486.4        470       10.9    7119.7s     165.6s


Test-Set Breakdown (AppWorld)
-----------------------------

Test Set   Tasks    Pass@1      Tokens    Tok/Task  LLM Calls   LLM/Task   Duration   Dur/Task
──────────────────────────────────────────────────────────────────────────────────────────────
challenge     24     66.7%   5,604,834   233,534.8        244       10.2    4211.6s     175.5s
normal        19     63.2%   3,618,080   190,425.3        226       11.9    2908.0s     153.1s

src/cuga/backend/cuga_graph/nodes/cuga_agent_core/graph/shared_nodes.py
```diff
@@ -241,6 +251,31 @@ async def call_model(state: Any, config: RunnableConfig = None) -> Command:
                 },
             )
 
+        # Model wants to finalize. Guard: refuse a final answer produced without
+        # any tool/code execution (e.g. a confident hallucination); inject a
+        # directive and continue instead. Bounded by the step limit.
+        if require_tool_call and code_exec_count == 0:
+            logger.warning(
+                "%s: require_tool_call_before_final — final answer with 0 tool calls; injecting directive and continuing",
```
sami-marreed: **Important — loguru `%s` regression.** Loguru formats with `{}`, not printf `%s`. This logs the literal `%s: ...` and silently drops `adapter.sender_name`. This is exactly the defect class that #389 just removed across 22 files (merged 2 commits before this PR's base). Fix with an f-string:

```python
logger.warning(
    f"{adapter.sender_name}: require_tool_call_before_final — final answer with 0 tool calls; injecting directive and continuing"
)
```

src/cuga/backend/cuga_graph/nodes/cuga_agent_core/graph/shared_nodes.py
```diff
@@ -214,8 +214,18 @@ async def call_model(state: Any, config: RunnableConfig = None) -> Command:
...
+        # inside the adapter metadata dict so no graph state schema change is
+        # needed (works for lite / supervisor / core).
+        require_tool_call = bool(getattr(settings.advanced_features, "require_tool_call_before_final", False))
+        code_exec_count = int((adapter.get_metadata(state) or {}).get("code_exec_count", 0) or 0)
+
         # ── Route: code → execute node; text → END or auto-continue ────────
         if code:
+            if require_tool_call:
+                # Record that at least one tool/code block has run this task.
+                meta_value["code_exec_count"] = code_exec_count + 1
```
sami-marreed: **Important — counts emitted code blocks, not successful tool calls.** `code_exec_count` increments whenever the model emits *any* code block, regardless of whether it executed successfully or actually called a tool. A model can emit `x = 1` (or a block that NameErrors) and thereby satisfy `require_tool_call_before_final`, then still hallucinate a final answer — weaker than the name implies. Default-off mitigates severity. Either increment only after the execute node reports a non-error tool invocation, or tighten the comment/flag wording to "at least one code block executed."

sami-marreed (commented):
Code review — overall: merge with fixes. Three independent, well-scoped fixes; defaults are backward-compatible and non-intrusive as claimed; all four new test files pass (78/78). One trivial must-fix plus two should-fix items, left inline.

**Strengths**
- Genuinely independent fixes on disjoint code paths, each individually gated.
- Arg-warning is provably non-mutating (forwards kwargs verbatim; tests assert "NOT coerced").
- Tests exercise real behavior (real code through `eval_with_tools_async`, real pydantic models, `assert_not_called()`), not mocks.
- Conservative classifier: negation/question guards, 400-char cap, anchored match, bounded by step limit.

**Minor (not inlined)**
- PR body says 76 tests; actually 78 passing.
- `meta_value` aliases the state metadata dict, so `meta_value["code_exec_count"] = ...` mutates graph state before `Command(update=...)` is returned. Works under LangGraph but a `dict(meta_value)` copy is cleaner.
- `require_tool_call_before_final` loops to the step limit on legitimate no-tool tasks (e.g. a greeting), returning a step-limit error instead of an answer. Opt-in, but worth noting in the flag's help text.

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/prepare_node.py
```diff
@@ -348,6 +358,11 @@ async def prepare_tools_and_apps(state: Any, config: Optional[RunnableConfig] =
                 tool_func = getattr(tool, '_run', None)
 
             if tool_func:
+                tool_func = make_arg_warning_callable(
+                    tool_func,
+                    getattr(tool, "args_schema", None),
+                    enable=_warn_args,
+                )
```
sami-marreed, line 363: **Important — default-on warning silently moves sync tools onto the event loop.** For a sync-only tool (no `.coroutine`, has `.func`/`._run`) with an `args_schema`, the old path was sync func → `make_tool_awaitable` → `run_in_executor` (worker thread). `make_arg_warning_callable` returns an `async def wrapper` that calls the sync `tool_func` inline; `make_tool_awaitable` then sees a coroutine function and awaits it directly — so the sync tool now runs **in the event loop** instead of a worker thread, and a blocking sync tool could stall the loop. Scope is narrow (only sync-only tools with a schema; `.coroutine` tools are unaffected) but it's a silent, default-on behavior change. Fix: dispatch sync callables via `run_in_executor` inside the wrapper, or only wrap when `inspect.iscoroutinefunction(tool_func)`.

haroldship:
@sami-marreed @Sergey-Zeltyn Re-ran appworld easy with increased timeout. Below results for gpt-oss-120b and gpt-4.1

GPT-4.1 is a bit below what Sergey had before (84.7%)

```text
Evaluation Comparison Report
============================

1 run(s) per configuration.


Summary
-------

Configuration                Runs     Pass@1     pass@1     pass^1      maj@1   Cons      Tokens    LLM     Time
────────────────────────────────────────────────────────────────────────────────────────────────────────────────
cuga (GPT-OSS-120B)             1      79.1%      79.1%      79.1%      79.1%   1.00  3,090,341.0  171.0   586.6s
cuga (GPT-4.1)                  1      81.4%      81.4%      81.4%      81.4%   1.00  7,190,672.0  399.0  2090.8s


Cost Summary
------------

Configuration                    Tokens    Avg/Task     LLM   Avg/Task      Time   Avg/Task
───────────────────────────────────────────────────────────────────────────────────────────
cuga (GPT-OSS-120B)          3,090,341.0    71,868.4   171.0        4.0    586.6s      13.6s
cuga (GPT-4.1)               7,190,672.0   167,224.9   399.0        9.3   2090.8s      48.6s
```

Sergey-Zeltyn:
There are several changes in PR, here I comment on one of them

Deterministic `looks_like_planning_text()` short-circuit before the LLM classifier in **nl_auto_continue_classifier.py:160.** For AppWorld this is a genuine behavioral delta on an already-active path — previously every no-code NL turn was classified by the LLM; now text matching the planning regex is forced to auto-continue without the LLM call.

**False positives = wasted steps.** Any genuine final answer that matches the regex (starts with we/I/let's/let me + optional discourse marker + a present-tense action/modal verb, ≤400 chars, no ?, no negation) now loops instead of finalizing. The design mitigates this well — past-tense finals ("I found…", "I determined…") don't match because the verb list is present-tense/infinitive with \b boundaries, and questions/negations are excluded. But a correct final phrased as "I'll return the list: …" would match and burn a step. With cuga_lite_max_steps=70 it can't loop forever, but repeated over-fire near the cap could convert a would-be pass into a step-limit error.

**My verdict when I consider AppWorld ground truth answers:** it should not affect appworld negatively

Sergey-Zeltyn:
Doing some gpt-oss log analysis to find possible concerns

**Concern (misdirection from the unknown-tool correction):** The new `_unknown_tool_correction` in `local_executor.py `format_error fires on every `NameError`, but its message assumes the missing name is a fabricated **tool**. When the missing name is actually an ordinary **variable**, the injected guidance ("not an available tool… call find_tools… do not retry") is wrong and could steer the model away from the real fix (define/compute the variable) — and it fires even on tasks that currently pass.

**Example (real, bundle b6d1104, score 1.0 — a passing task):**

Agent runs a block referencing **formatted_total,** never defined → `NameError: name 'formatted_total' is not defined.`
**Current behavior:** bare NameError → agent recovers next step by writing the actual computation → task passes.
**With the correction:** error also carries "[tool-name correction] 'formatted_total' is NOT an available tool… Call tools by the EXACT name returned by find_tools… Do not retry" (closest matches in my sim: format, total, find_tools) — misleading noise on a variable, injected into a currently-passing trajectory.

**Rare** (1 exposed passing task in 87; the other 84 NameErrors were genuinely fabricated tool names where the correction helps), but it's the one place the change can actively hurt.

Sergey-Zeltyn:
Second (and last) concern 

**loose cutoff=0.4 in difflib.get_close_matches(missing, available_tools, n=5, cutoff=0.4) (_unknown_tool_correction, local_executor.py):**

The app-prefix structure usually keeps suggestions in the right app and often exact (e.g. `amazon_get_returns_returns_get → amazon_show_returns_returns_get`). But the 0.4 cutoff is loose enough to surface two hazards:

- Same-app but semantically wrong — including destructive tools. phone_get_sms_messages_sms_get` (a read) → top `suggestion phone_delete_sms_message_sms_delete` (a delete). The correct read tool appears too, but offering a destructive op first is the sharpest risk if the model follows the head of the list.

- Cross-app junk when the right tool isn't loaded. `simple_note_list_notes_get` → `spotify_create_playlist_post` (wrong app entirely). Likely partly a proxy artifact of my reconstructed namespace, and on an already-failing task — but it shows 0.4 lets unrelated names through when no good match is present.

haroldship:
@Sergey-Zeltyn addressed in cc2d8a8a:

**Variable NameError misdirection** — the correction now fires only when the missing name is *called* like a function in the executed code (`format_error` receives the code and checks for `name(...)`). A plain variable reference like your `formatted_total` example keeps the bare NameError, so the passing trajectory you found is no longer disturbed.

**Loose cutoff=0.4** — raised to 0.6. Verified against your examples: `simple_note_list_notes_get → spotify_create_playlist_post` scores 0.481 and is now filtered (falls back to the find_tools re-query hint), while legitimate corrections like `amazon_get_returns… → amazon_show_returns…` (0.885) survive. The destructive-lookalike case (`phone_get_sms… → phone_delete_sms…`, 0.831) clears any reasonable cutoff, so the suggestion list now carries an explicit caution: "Similarly named tools can do very different things (e.g. delete vs get) — pick only a suggestion whose action matches your intent."

Regression tests added for all three behaviors (variable NameError stays bare, cross-app junk filtered, caution present) — 82 tests pass.

**Planning-text short-circuit** — no change; per your analysis the present-tense verb list and question/negation guards keep AppWorld finals out of the match, and the step cap bounds any residual over-fire.

src/cuga/backend/cuga_graph/nodes/cuga_agent_core/graph/shared_nodes.py
haroldship: Fixed in 28276814 — f-string, matching the #389 sweep.

src/cuga/backend/cuga_graph/nodes/cuga_agent_core/graph/shared_nodes.py
haroldship: Took the wording option in 28276814: the code comment and the settings.toml help now state it counts emitted/executed code blocks (a `x = 1` block or an erroring block still counts) — a lightweight grounding proxy, not a confirmed-tool-call guarantee. Counting only non-error tool invocations would need the execute node to report back into the adapter metadata; given the flag is default-off I kept the cheap proxy and made the docs honest about it.

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/prepare_node.py
haroldship: Fixed in 28276814 — `make_arg_warning_callable` now preserves the sync/async nature of the wrapped tool: async tools get an async wrapper, sync tools a sync wrapper, so `make_tool_awaitable` still sees a non-coroutine for sync-only tools and dispatches them via `run_in_executor` as before. Regression test covers the sync path.

haroldship:
@sami-marreed swept your review — inline threads answered individually; the non-inlined minors:

- **test count**: PR body updated to 82 (78 at review time + 4 regression tests from the Sergey-review fixes in cc2d8a8a).
- **`meta_value` aliasing**: fixed in 28276814 — the metadata dict is copied before the `code_exec_count` bump, so only the returned `Command` payload is mutated, not live graph state.
- **`require_tool_call_before_final` on tool-free tasks**: the settings.toml help text now carries the caveat verbatim (loops to the step limit and returns a step-limit error instead of an answer; flag stays default-off).

sami-marreed:
@coderabbitai can you label this pr readiness ?

src/cuga/backend/cuga_graph/nodes/cuga_agent_core/graph/shared_nodes.py
```diff
@@ -241,6 +256,31 @@ async def call_model(state: Any, config: RunnableConfig = None) -> Command:
                 },
             )
 
+        # Model wants to finalize. Guard: refuse a final answer produced without
+        # any tool/code execution (e.g. a confident hallucination); inject a
+        # directive and continue instead. Bounded by the step limit.
+        if require_tool_call and code_exec_count == 0:
```
sami-marreed: Edge cases when require tool call is true:

1. Saying to the agent `what can you do?` `-->` no tools `-->` Error
2. Will affect working with skills and no tools
3. Follow-up questions on previous threads

Problem: Cuga returns `intermediate/hallucinated` answers without doing actual computations based on tools
examples:

1. Let me perform the second phase. ( intermediate )
2. Cuga answers based on general LLM knowledge instead of tool computation

Open questions:
1. Does the flag of autonomous mode relate to this flag require_tool_call?

Decision:

1. Let's remove it and open issue on agent answering from general knowledge instead of tools, use might want to configure that when to answer in general/llm vs based on tools

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/prepare_node.py
```diff
@@ -334,6 +335,15 @@ async def prepare_tools_and_apps(state: Any, config: Optional[RunnableConfig] =
...
         # rewriting is no longer needed here — filesystem tools come from
         # the consolidated runtime class below, not from MCP.
+        # Pre-flight arg WARNING: the sandbox calls tool.coroutine directly,
+        # bypassing the StructuredTool's args_schema, so nothing flags malformed
+        # kwargs before the registry. This detector logs (never mutates) suspect
+        # shapes — the dict-as-string bug and friends. The former coercion layer
+        # (Wave-1 Change #4) is WONTFIX: a 200-task M3 mining pass found the
+        # failure does not occur in the current corpus. See arg_warning.py.
+        _af = getattr(settings, "advanced_features", None)
+        _warn_args = bool(getattr(_af, "cuga_lite_warn_suspect_args", True))
```
sami-marreed, line 343: Provide issue with example input and output. delete M3 logs etc

src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/local/local_executor.py
```diff
@@ -103,4 +122,57 @@ def format_error(self, error: Exception) -> str:
...
+            error_msg += correction
         return error_msg
+
+    @staticmethod
+    def _unknown_tool_correction(
+        error: Exception,
+        available_tools: Optional[list[str]],
+        code: Optional[str] = None,
+    ) -> str:
+        """Build a correction hint when the agent calls a non-existent tool.
```
sami-marreed, line 141: Add issue for tool name hallucination, we might already have one

src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/local/local_executor.py
```diff
@@ -103,4 +122,57 @@ def format_error(self, error: Exception) -> str:
...
+        steer it away from the real fix.
+        """
+        if not isinstance(error, NameError) or not available_tools:
+            return ""
+
+        missing = getattr(error, "name", None)
+        if not missing:
+            return ""
+
+        if code is not None and not re.search(rf"\b{re.escape(missing)}\s*\(", code):
```
sami-marreed: This can match many phrases like `var = " agent_1"`

sami-marreed: sometimes it can also catch a helper function that agent generated with current code

async def helper1(arg1):
  pass

res = await helper1(arg1)

src/cuga/backend/cuga_graph/nodes/cuga_lite/nl_auto_continue_classifier.py
```diff
@@ -104,6 +157,9 @@ async def classify_nl_auto_continue(
         return False
     visible = normalize_assistant_text(assistant_visible)
     reasoning = normalize_assistant_text(reasoning_excerpt)
+    if looks_like_planning_text(visible):
+        logger.info("NL auto-continue: planning-text fast-path matched; auto-continuing")
+        return True
     combined = build_combined_content_and_reasoning(visible, reasoning)
```
sami-marreed, line 181: Make sure you have issue created on looks like planning text and what's the best approach to solve it

src/cuga/backend/cuga_graph/nodes/cuga_lite/nl_auto_continue_classifier.py
```diff
@@ -29,6 +29,59 @@
...
+    r"^(?:(?:ok(?:ay)?|now|first(?:ly)?|next|then|so|alright|well)[\s,]+)*"
+    r"(?:we|i|let'?s|let\s+me)\b"
+    r"(?:(?!\.).)*?\b"
+    r"(?:need\s+to|have\s+to|should|must|will|'ll|going\s+to|gonna|"
+    r"start\s+by|begin\s+by|"
+    r"search|discover|find|look\s+up|fetch|call|query|inspect|"
+    r"explore|examine|check|investigate|figure\s+out|determine|"
+    r"retrieve|gather|list|enumerate)\b",
+    re.IGNORECASE,
+)
```
sami-marreed, line 64: @coderabbitai how this will react to `Ok i will fetch the infromation, but first i require your ID`

sami-marreed: Will this enter endless loop because of this?

haroldship:
## AppWorld results

| Litellm/gpt4.1 | Number of tasks | 120 sec Average@1 | Average duration sec | Average Input Tokens | Average output tokens | Average cached token |
|---|---|---|---|---|---|---|
| Test easy | 43 | 0.814 | 55.1 | 137132 | 5497 | 80714 |
| Test medium | 40 | 0.575 | 105 | 376932 | 11096 | 218349 |

haroldship:
## Vakra results

| Litellm/gpt-oss-120b | Number of tasks | 120 sec Average@1 | Average duration sec | Average Input Tokens | Average output tokens | Average cached token |
|---|---|---|---|---|---|---|
| Capability 2 | 100 | 0.410 | 88.4 | 228854 | 7771 | 0 |
| Capability 3 | 100 | 0.150 | 102.7 | 218459 | 10998 | 0 |
| Total | 200 | 0.280 | 95.5 | 223657 | 9385 | 0 |

Notes:
- Bundle `20260709_115720_compare_gpt-oss-120b_no-policies` (M3, cuga v0.2.20, no policies, 10 domains × 10 tasks per capability, 1 run).
- Token splits come from the Langfuse traces; output tokens include gpt-oss reasoning tokens.
- The endpoint reported `input_cache_read = 0` on every generation, so average cached tokens is 0.
- 2 capability-2 tasks (1.25M and 1.21M total tokens) are missing Langfuse traces; their input/output split was imputed from the capability-2 traced input fraction (96.7%) applied to their recorded totals.

sami-marreed (approved):

sami-marreed:
will merge after ido's

PR #417, opened 2026-06-30
fix(272): CugaLite executes code that guesses at undeclared tool output shapes

haroldship:
> **Note:** This reopens the work from #362, which GitHub could not reopen directly — its branch head was force-moved after the PR was closed, leaving #362's recorded head commit stale. Branch content is unchanged; this is a fresh PR from the same `feat/272-stage2-block-isolation` branch.

---

## The problem (#272)

CugaLite writes Python code that calls tools. To write correct code it needs to know what each tool *returns* — the output schema. Some tools don't declare one, and for MCP tools with no declared schema the registry was **making one up**: it injected a placeholder claiming the tool returns a string.

So the model confidently wrote code like this, all in one script:

```python
res = await file_readfile('x')   # actually returns a list
res_2 = res[0:15]                # written assuming it's a string — wrong
```

By the time the wrong assumption blew up, the whole script had already run. The model was guessing the shape and executing the guess in one shot. That's the literal failure trace in #272.

## The fix, in three parts

**Part A — be honest in the prompt.** Detect tools with no real declared schema: an empty schema, or the synthetic MCP placeholder — which the MCP manager now tags with a `_synthetic_placeholder` marker so it can't be confused with a tool that *genuinely* returns a bare string. Instead of rendering the fake schema, the prompt tells the model: *call this tool alone in its own block and `print()` the raw result; write follow-up code only after you've seen the real shape.*

**Part B — enforce it, don't just ask.** Part A is only a request; the model can still write "probe + code that uses the probe" in one response. So during code extraction, if the response contains a call to a still-unknown-schema tool, **everything after that call is cut off**. Only the probe runs; the model sees the real result and writes the rest next turn with actual knowledge instead of a guess. This lives behind a new adapter hook, `CoreGraphAdapter.get_tools_needing_probing()`, which defaults to `frozenset()` — CugaSupervisor and any adapter that hasn't opted in behave exactly as before.

**Part C — don't probe twice.** The first time an unknown-schema tool returns a result, the sandbox records a short description of its shape (`"list of 3 items"`, `"dict with keys ['id', 'name']"`). That description is injected into the system prompt on every later turn ("Observed tool output shapes — use this, no need to probe again"), and the tool drops out of Part B's enforcement set. The one-call-at-a-time restriction only applies until the shape is known.

## Implementation notes (the fiddly bits)

- **Code extraction has messy fallback paths** (model forgot the closing ``` fence, or emitted raw unfenced Python). There the code is one blob with no block boundaries, so Part B truncates at the *line* level instead — growing the cut until the kept prefix parses, so a multi-line probing call is kept whole and the cut never produces syntactically broken Python.
- **`prepare_node` builds the unknown-schema set from `tools_for_execution`, not `tools_for_prompt`.** When `find_tools` shortlisting is active (the common case once an app has more than a handful of tools), `tools_for_prompt` collapses to just the `find_tools` meta-tool — using it would have left the set permanently empty and silently disabled Part B and Part C for every tool actually reachable through `find_tools`.
- **The sandbox now enables tool-call tracking internally** whenever a weak-schema tool's shape hasn't been observed yet, even if the caller didn't opt into `track_tool_calls` — without leaking the tracked calls into the caller's `tool_calls` state. Without this, Part C never recorded anything in production (the Stage 1 A+C branch was dead on arrival; there's a regression test that reproduces this).
- **Weak-schema detection trusts the `_synthetic_placeholder` marker, not schema shape.** A genuine string-returning tool (OpenAPI text body, or an MCP tool that actually declares `outputSchema: {"type": "string"}`) produces a `success` schema byte-identical to the placeholder; matching on shape would have suppressed real schemas. The marker key is a plain literal on both sides to avoid a graph→registry import dependency.
- **Two policy-approval tests were loosened** because the agent's first approved code block is now often a structure probe: the "last AI message mentions digital_sales" assertion moved to the code actually pending approval, and the "modified code differs from original" assertion was dropped since both requests' first blocks are legitimately identical probes.

## Staging

This PR is the union of two branches, kept separate during development so each component's contribution could be evaluated independently:

- `feat/272-tool-output-schema-probing` — Stage 1, Parts A+C only
- `feat/272-stage2-block-isolation` (this branch, off Stage 1) — adds Part B

## Eval results (Vakra M3, 1 task, 5 runs each)

| Branch | Commit | Pass@1 | ExactMatch | Avg tokens | Avg LLM calls | Avg time |
|---|---|---|---|---|---|---|
| main | 8d3d3599 | 0.0% | 0.20 | 301,614 | 15.2 | 65.2s |
| feat/272-tool-output-schema-probing (A+C) | 2dc47738 | 0.0% | 0.80 | 205,719 | 11.2 | 25.2s |
| feat/272-stage2-block-isolation (A+B+C) | 4f623786 | 60.0% | 1.00 | 214,005 | 11.0 | 24.7s |

A+C alone (asking nicely) already cuts tokens/calls/time roughly in half and raises ExactMatch substantially, but doesn't move Pass@1. Adding B (forcing the split) is what converts that into actual task success — the enforcement is the part that works.

## Test plan

- [x] Stage 1 unit tests (`prompt_utils`, `prepare_node`, `sandbox_node`, `graph_adapter`) — new + extended
- [x] Stage 2 unit tests (`code_extraction`, `graph_nodes`, `shared_nodes`, `graph_adapter`) — new + extended, including an end-to-end reproduction of the issue's literal original multi-block trace
- [x] Full regression: `cuga_lite` + `cuga_agent_core` + `cuga_supervisor` suites green (363 passed, 1 skipped, 3 pre-existing unrelated e2b-package-not-installed failures)
- [x] `ruff check` clean
- [ ] Broader eval run across the full Vakra(M3) task set (this PR's numbers are a single-task, 5-run sample)

## Summary by CodeRabbit

* **New Features**
  * Added probing-aware code extraction: when probing is needed, generated code stops after the first relevant tool invocation instead of including extra fenced blocks.
  * Enhanced weak-schema handling by adding an “Observed tool output shapes (this session)” section to the system prompt and tracking observed output shapes during sandbox runs.
* **Bug Fixes**
  * Improved truncation for partially fenced/malformed responses and ensured consistent probing behavior across response fields.
  * Made observed-shape tracking work even when tool-call details aren’t returned externally.


docs/superpowers/plans/2026-06-23-weak-tool-schema-probing-stage1.md
```diff
@@ -0,0 +1,637 @@
+# Weak Tool Output Schema Probing — Stage 1 (A+C) Implementation Plan
```
sami-marreed: Remove

docs/superpowers/specs/2026-06-23-weak-tool-schema-probing-design.md
```diff
@@ -0,0 +1,222 @@
+---
```
sami-marreed: remove

sami-marreed: remove

src/cuga/backend/cuga_graph/nodes/cuga_agent_core/execution/code_extraction.py
```diff
@@ -38,21 +51,54 @@ def extract_and_combine_codeblocks(text: str) -> str:
...
+    call. Recovery paths produce one contiguous block (no fence boundaries to
+    split on), so block-level truncation is a no-op here — fall back to line
+    granularity to keep the probe from running with later dependent code."""
+    if not tools_needing_probing:
+        return code
+    pattern = _probing_call_pattern(tools_needing_probing)
+    lines = code.split("\n")
+    for i, line in enumerate(lines):
+        if pattern.search(line):
+            return "\n".join(lines[: i + 1])
```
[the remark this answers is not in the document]
haroldship: > Line-based cutoff can return invalid Python for multiline probes.

Fixed in 056cccf3. `_truncate_after_first_probing_line` no longer stops at the first matching **line** — it now grows the cut to the end of the first probing **statement** using the same incremental `compile()` boundary check the recovery path already uses. So a multiline probe like `res = await file_readfile(\n    "x"\n)` is kept whole (valid, parseable Python) and executes, while later dependent code is still dropped.

Added two regression tests (raw Python + unclosed-fence, both with a multiline probing call) that assert the kept code compiles; the full `test_code_extraction.py` suite passes 20/20.

docs/superpowers/plans/2026-06-23-weak-tool-schema-probing-stage1.md
haroldship: Removed in 056cccf3.

docs/superpowers/specs/2026-06-23-weak-tool-schema-probing-design.md
haroldship: Removed in 056cccf3.

haroldship: Removed in 056cccf3.

src/cuga/backend/cuga_graph/nodes/cuga_agent_core/execution/code_extraction.py
```diff
@@ -20,12 +20,22 @@
...
+    instead of running a blind guess in the same execution.
+    """
     code_blocks = re.findall(BACKTICK_PATTERN, text, re.DOTALL)
 
     if code_blocks:
-        return "\n\n".join(block.strip() for block in code_blocks)
+        blocks = [block.strip() for block in code_blocks]
+        if tools_needing_probing:
+            blocks = _truncate_after_first_probing_block(blocks, tools_needing_probing)
+        return "\n\n".join(blocks)
```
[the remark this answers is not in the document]
haroldship: Addressed in 2a60ad6 — the recovery/raw-code paths (unclosed fence, raw `print(...)`) now apply the same first-probing-block truncation as the fenced path, so a probe can't run in the same turn as later dependent code. Follow-up 056cccf3 hardens that path further so a multiline probing statement is kept whole rather than cut mid-line (see the multiline thread above).

src/cuga/backend/cuga_graph/nodes/cuga_lite/prompt_utils.py
```diff
@@ -148,6 +155,24 @@ def get_tool_params_str(tool: StructuredTool) -> str:
...
+        when they declare no ``outputSchema`` (see mcp_manager.py).
+        """
+        response_schemas = {}
+        if hasattr(tool, 'func') and hasattr(tool.func, '_response_schemas'):
+            response_schemas = tool.func._response_schemas
+
+        if not response_schemas or not isinstance(response_schemas, dict):
+            return True
+
+        return response_schemas.get('success') == {'type': 'string'}
```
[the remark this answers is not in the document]
haroldship: Addressed in commits 8287bd2–70a84f9 — weak-schema detection now keys off an MCP-specific placeholder signal rather than the bare `{"type": "string"}` shape, so genuine string-returning tools keep their real schema and are no longer forced into the probing directive.

haroldship:
Filed #427 for the pre-existing test-harness gap surfaced here: the tool-approval e2e helper (`resume_graph_with_response`) never delivers the HITL response, so the approval tests don't actually exercise post-approval execution. Tracked separately from this PR — the probing change only altered the regenerated code's content, it didn't create the gap.

haroldship:
*Scope & intent*

> 1. Are we treating this as the full fix for #272, or only Stage 2 on top of Stage 1?

This is the full implementation as far as I'm concerned. It's a full feature with 3 parts. This PR contains both stage 1 and stage 2. See details in this issue [comment](https://github.com/cuga-project/cuga-agent/issues/272#issuecomment-4785417258)

> 2. Why is the branch named "stage2" if A, B, and C all ship together?

It has to do with how the 3 parts of the fix were implemented and tested. We had a design discussion, and came up with several ideas on how to solve the issue. In the end, I tried 3 fixes derived from that. At first, I combined A and C in stage 1. I branched off that branch and added B in stage 2. All 3 are necessary, so stage 2 is the branch I want to merge.

> 3. Is CugaSupervisor deliberately left out forever, or just for this PR?

CugaSupervisor was never considered. If it's desired, I suggest a second issue be created and I'm happy to work on it. However, this was driven by Vakra benchmark which does not use CugaSupervisor. Also, [the base adapter returns an empty tool set](https://github.com/cuga-project/cuga-agent/blob/87e17c36/src/cuga/backend/cuga_graph/nodes/cuga_agent_core/graph/graph_nodes.py#L142-L146) and the truncation helpers early-return on empty, so Supervisor's path is unchanged and so it is unaffected by this PR.

> 4. What does "done" look like for #272 once this merges?

For #272, there is support for tools with no output schema. The support has 3 parts that complement each other. This support works well enough to reduce time, LLM calls, tokens, and produce correct answers when tools have no output schema.


*How it behaves*

> 1. If the model puts the probe and the follow-up code in the same Python block, do we still run both?

Depends on how the model formats it. The normal case is fenced blocks, and there we [cut the list of blocks](https://github.com/cuga-project/cuga-agent/blob/87e17c36/src/cuga/backend/cuga_graph/nodes/cuga_agent_core/execution/code_extraction.py#L63-L70). We keep everything up to and including the block that contains the probe call, and drop the rest. So if the model puts the probe in one block and the dependent code in the next block, only the probe runs. That's the case I was seeing in practice and it's what part B fixes ([test](https://github.com/cuga-project/cuga-agent/blob/87e17c36/src/cuga/backend/cuga_graph/nodes/cuga_agent_core/tests/execution/test_code_extraction.py#L102)).

If the model puts both in the same fenced block, both currently run. The probe only finds the block, not where the call is inside the block, so the matching block is kept whole. This is a second violation of the same instruction! The directive we inject in place of the output schema says to call the tool ALONE in its own block so it's the model ignoring the format rule on top of the ordering rule. I haven't seen this in the Vakra runs, but it's a real gap and it's the "mid-block splitting" item on my follow-up list below.

On the other hand, in case of unclosed fence or python with no fence, the recovery paths do [cut mid-blob, at statement granularity](https://github.com/cuga-project/cuga-agent/blob/87e17c36/src/cuga/backend/cuga_graph/nodes/cuga_agent_core/execution/code_extraction.py#L73-L99). There, everything is collapsed into one blob, so there are no block boundaries to cut on. Along the way I fixed a bug where a multiline probe call produced invalid Python, so the probe never ran and the isolation was bypassed. It now grows the cut to the end of the statement incrementally by testing python compilation.

> 2. When we cut off later blocks, does the model get told that happened, or does it just see the probe result?

I want to emphasise that we only cut the code when the model ignores the instructions. When there is no output schema, we add a special instruction in place of the tool's output schema, to only run the tool as a probe and to discover the output type. The code cutting is done when the model generates code that ignores this. Afterwards, we don't tell the model that the code was cut off. In the next step, the model sees that its code asked for two things, but only got one back. It is implicit, and the model is supposed to retry.

> 3. After we've seen a tool's shape once, can the model use it freely again for the rest of the session?

Yes. There is an internal list of schemaless tools, created when the tools are first discovered. Then, only when a tool is [called successfully](https://github.com/cuga-project/cuga-agent/blob/87e17c36/src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/sandbox_node.py#L52-L62), the shape is recorded _for the session_ and [logically gets skipped on any subsequent calls](https://github.com/cuga-project/cuga-agent/blob/87e17c36/src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/graph_adapter.py#L111-L112), so that the probing only happens once. Once the shape is recorded, it's added to the system prompt each turn as an observed-shapes block, and the tool is no longer isolated, so the model can chain against it freely.

> 4. What happens if the probe call fails — do we still lock that tool, or keep asking for probes?

[Still keep asking.](https://github.com/cuga-project/cuga-agent/blob/87e17c36/src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/sandbox_node.py#L52-L62) There is no cap to this.


*Tradeoffs*

> 1. Are we okay with an extra turn (and extra tokens) the first time each weak-schema tool is used?

Yes, overall effect observed from testing 5 runs on a task that consistently failed, is that it saves tokens and turns. This table, copied from the description above, shows the effects:
| Branch | Commit | Pass@1 | ExactMatch | Avg tokens | Avg LLM calls | Avg time |
|---|---|---|---|---|---|---|
| main | 8d3d3599 | 0.0% | 0.20 | 301,614 | 15.2 | 65.2s |
| feat/272-tool-output-schema-probing (A+C) | 2dc47738 | 0.0% | 0.80 | 205,719 | 11.2 | 25.2s |
| feat/272-stage2-block-isolation (A+B+C) | 4f623786 | 60.0% | 1.00 | 214,005 | 11.0 | 24.7s |

So  B trades ~4% more tokens than A+C for 0% → 60% Pass@1, and it's still ~29% below main.


> 2. Is a short blurb like "list of 3 items" enough, or do we need richer shape info?

The problem I was observing is that the python code had to guess at return types. It would generate calls like `result.get(0)` which would work fine on a dict but fail on a list, str or int. It would throw an exception. The next step would not work in many cases. The table in the previous question was for a task that failed consistently.

[What gets stored](https://github.com/cuga-project/cuga-agent/blob/87e17c36/src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/sandbox_node.py#L33-L49) is: *dicts:* key names only, first 8, no value types; *lists:* length plus first item's type and 120 chars; *strings:* length plus 120 chars; *everything else:* just the type name. It doesn't do nesting. That's plenty for flat results but pretty thin for nested response.

> 3. Could the name-matching cut the wrong code (e.g. a tool name mentioned in a comment or string)?

Yes, [it could match a comment or string literal](https://github.com/cuga-project/cuga-agent/blob/87e17c36/src/cuga/backend/cuga_graph/nodes/cuga_agent_core/execution/code_extraction.py#L59-L60). This could cause code to be truncated early, which would waste a step, but not generate any unsafe code. It does check for prefix collisions such as `get_file_2()` which would not be matched for `get_file()` ([test](https://github.com/cuga-project/cuga-agent/blob/87e17c36/src/cuga/backend/cuga_graph/nodes/cuga_agent_core/tests/execution/test_code_extraction.py#L135)).

> 4. Will users/policies now often approve a "print the structure" probe before the real work?
The new behaviour is only triggered when a tool has no output schema. I can't say what effect any particular policy would have on this. But if the tool has an output schema it will behave the same as before.

*Evidence & rollout*

> 1. Is one Vakra task × 5 runs enough to merge, or do we need the full task set first or appworld?

I have run the full Vakra. The results were like this. So between #416 and this PR, we gained 9 tasks or 4.5% while saving turns and tokens vs. main branch.
image

> 2. Which real apps/tools do we expect this to hit hardest in production?

Any tools that have no output schema. In particular, openapi tools with empty response_schemas, and MCP tools with no outputSchema. Also, [the set is computed from `tools_for_execution` rather than `tools_for_prompt`](https://github.com/cuga-project/cuga-agent/blob/87e17c36/src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/prepare_node.py#L589-L599) so tools reachable only through find_tools shortlisting are still covered.

> 3. How will we know if this is helping or hurting (more turns, more approvals, fewer crashes)?

Helping: succeeding in tasks that call such tools. Fewer wasted turns and tokens, again for such tools. Hurting: consistent or intermittent failures in such tools or more tokens and turns.

> 4. Anything we should flag as follow-up before merge (docs, metrics, mid-block splitting)?

- An issue for cugasupervisor
- Mid-block splitting in case model generates one large code block containing unprobed tool call plus follow-up usage
- Sharing across sessions
- Time-to-live or other invalidation
- Handle tools with heterogeneous output types
- Cap on retries on probing in case the tool probe never succeeds

src/cuga/backend/cuga_graph/nodes/cuga_lite/prompt_utils.py
```diff
@@ -148,6 +163,29 @@ def get_tool_params_str(tool: StructuredTool) -> str:
...
+        match on shape — that suppressed real schemas. Instead the manager tags
+        the synthetic placeholder with ``_synthetic_placeholder`` and we trust
+        that marker, leaving every genuinely-declared schema intact.
+        """
+        response_schemas = {}
+        if hasattr(tool, 'func') and hasattr(tool.func, '_response_schemas'):
+            response_schemas = tool.func._response_schemas
+
+        if not response_schemas or not isinstance(response_schemas, dict):
+            return True
```
sami-marreed, line 185: @coderabbitai what if response schema is pydantic object

src/cuga/backend/cuga_graph/nodes/cuga_lite/prompt_utils.py
```diff
@@ -148,6 +163,29 @@ def get_tool_params_str(tool: StructuredTool) -> str:
         else:
             return "**kwargs"
 
+    @staticmethod
+    def is_weak_schema_tool(tool: StructuredTool) -> bool:
```
sami-marreed, line 167: Validat edge cases and examples of multiple tools with  -> Dict, -> dict, -> str is string ok?

src/cuga/backend/cuga_graph/nodes/cuga_lite/prompt_utils.py
sami-marreed: @coderabbitai Are there edge cases could be considered here weak but are not weak

haroldship:
@sami-marreed Both items from this thread are addressed:

- **Pydantic / non-dict Mapping normalization**: `is_weak_schema_tool` now normalizes `_response_schemas` via a new `_normalize_response_schemas` helper before checking for a real schema. A Pydantic `BaseModel` (class or instance) is converted via `model_json_schema()`/`.schema()`; a non-`dict` `Mapping` (`MappingProxyType`, `UserDict`, custom `Mapping`) is treated like a dict. Neither is misclassified as weak anymore — only an empty/absent schema or the `_synthetic_placeholder` marker counts as weak. Added regression tests for both, plus explicit `-> Dict`/`-> dict`/`-> str` cases in `test_prompt_utils_weak_schema.py`.

- **Feature gating**: the whole probing feature (directive injection, code truncation, shape memory) is now behind `advanced_features.cuga_lite_weak_schema_probing_mode`, with three modes:
  - `"combine_and_execute"` (default, and the fallback for unset/invalid values) — full legacy behavior, feature off.
  - `"get_first_and_execute"` — probes and truncates every call, never remembers a shape.
  - `"truncate_at_first_probe"` — today's full behavior: probe once, remember, stop enforcing.

  An invalid value logs a warning and falls back to `"combine_and_execute"`.

Landed in `integration/m3-eval` at 56b0f6fd, with pytest coverage for the mode resolver, the gating behavior at each chokepoint, and the schema-normalization edge cases.

sami-marreed:
> @sami-marreed Both items from this thread are addressed:
> 
>     * **Pydantic / non-dict Mapping normalization**: `is_weak_schema_tool` now normalizes `_response_schemas` via a new `_normalize_response_schemas` helper before checking for a real schema. A Pydantic `BaseModel` (class or instance) is converted via `model_json_schema()`/`.schema()`; a non-`dict` `Mapping` (`MappingProxyType`, `UserDict`, custom `Mapping`) is treated like a dict. Neither is misclassified as weak anymore — only an empty/absent schema or the `_synthetic_placeholder` marker counts as weak. Added regression tests for both, plus explicit `-> Dict`/`-> dict`/`-> str` cases in `test_prompt_utils_weak_schema.py`.
> 
>     * **Feature gating**: the whole probing feature (directive injection, code truncation, shape memory) is now behind `advanced_features.cuga_lite_weak_schema_probing_mode`, with three modes:
>       
>       * `"combine_and_execute"` (default, and the fallback for unset/invalid values) — full legacy behavior, feature off.
>       * `"get_first_and_execute"` — probes and truncates every call, never remembers a shape.
>       * `"truncate_at_first_probe"` — today's full behavior: probe once, remember, stop enforcing.
>       
>       An invalid value logs a warning and falls back to `"combine_and_execute"`.
> 
> 
> Landed in `integration/m3-eval` at [56b0f6f](https://github.com/cuga-project/cuga-agent/commit/56b0f6fd58bd4d919d7f1ce972aac3c1f894ef02), with pytest coverage for the mode resolver, the gating behavior at each chokepoint, and the schema-normalization edge cases.

@coderabbitai please suggest different flag names for `cuga_lite_weak_schema_probing_mode` something like cuga_lite_code_blocks_handling_strategy ? what you think

sami-marreed (approved):

sami-marreed:
@haroldship there are cases where probing at always first block is good too, how do you reommend we add that

PR #423, opened 2026-06-30
chore(watsonx): default to openai/gpt-oss-120b

### LINKED ISSUE #422: [Bug]: Default Watsonx model should be openai/gpt-oss-120b (not llama-4-maverick)
### Issue type

Enhancement / configuration default

### What you want and why

`src/cuga/configurations/models/settings.watsonx.toml` currently defaults **all** agent blocks to `meta-llama/llama-4-maverick-17b-128e-instruct-fp8`. For deployments standardizing on **`openai/gpt-oss-120b`** on IBM Watsonx (e.g. regulated / on-prem environments where gpt-oss-120b is the supported model), the Watsonx profile should default to gpt-oss-120b so it works out-of-the-box — without requiring every user to set a `MODEL_NAME` override.

Today, with no `MODEL_NAME` in the environment, `_get_model_name` (`src/cuga/backend/llm/models.py`) falls through to the TOML `model_name`, i.e. llama-4-maverick. So the only way to run gpt-oss-120b by default is to override `MODEL_NAME` everywhere.

gpt-oss-120b has been verified working on Watsonx: server-side `response_format=json_schema` enforcement, the full task-decomposition graph, policy, and the `cuga` demo server all run cleanly against it.

### How it could work

Change `model_name` in every `[agent.*.model]` block of `settings.watsonx.toml` to `openai/gpt-oss-120b` (leaving `temperature` / `max_tokens` unchanged). Users who still want llama-4 can set `MODEL_NAME=meta-llama/llama-4-maverick-17b-128e-instruct-fp8` or edit the TOML.

### Links or extra context

Note: this is a **default-behavior change** for all Watsonx users — anyone on `settings.watsonx.toml` without a `MODEL_NAME` override would switch from llama-4 to gpt-oss-120b.

### Checklist

segevshlomovIBM:
## Chore

Fixes #422

### Summary
Default the Watsonx profile to `openai/gpt-oss-120b`: all `[agent.*.model]` blocks in `src/cuga/configurations/models/settings.watsonx.toml` switch from `meta-llama/llama-4-maverick-17b-128e-instruct-fp8` to `openai/gpt-oss-120b` (`temperature`/`max_tokens` unchanged). With no `MODEL_NAME` override, Watsonx agents now default to gpt-oss-120b; users who want llama-4 can set `MODEL_NAME` or edit the TOML.

⚠️ **Default-behavior change** for all Watsonx users (llama-4 → gpt-oss-120b unless `MODEL_NAME` is set).

### Testing
- [x] Verified locally

With `MODEL_NAME` unset (so the TOML default is exercised):
- `LLMManager.get_model(...)` resolves `model_id=openai/gpt-oss-120b` from the TOML default; a live `invoke` returned `OK` against Watsonx.
- `uv run demo` (CUGA FastAPI server): clean `Application startup complete`, **HTTP 200** on `/` and `/docs`, log line `Using model_name from TOML: openai/gpt-oss-120b`, 0 errors.

## Summary by CodeRabbit

* **New Features**
  * Updated the Watsonx-backed AI models used across several agent workflows to a newer model, which may improve response quality and consistency.

### DIFF
--- a/src/cuga/configurations/models/settings.watsonx.toml
+++ b/src/cuga/configurations/models/settings.watsonx.toml
@@ -1,60 +1,60 @@
 [agent.task_decomposition.model]
 platform = "watsonx"
-model_name = "meta-llama/llama-4-maverick-17b-128e-instruct-fp8"
+model_name = "openai/gpt-oss-120b"
 temperature = 0.1
 max_tokens = 16000
 
 [agent.shortlister.model]
 platform = "watsonx"
-model_name = "meta-llama/llama-4-maverick-17b-128e-instruct-fp8"
+model_name = "openai/gpt-oss-120b"
 temperature = 0.1
 max_tokens = 16000
 
 [agent.planner.model]
 platform = "watsonx"
-model_name = "meta-llama/llama-4-maverick-17b-128e-instruct-fp8"
+model_name = "openai/gpt-oss-120b"
 temperature = 0.1
 max_tokens = 16000
 
 [agent.chat.model]
 platform = "watsonx"
-model_name = "meta-llama/llama-4-maverick-17b-128e-instruct-fp8"
+model_name = "openai/gpt-oss-120b"
 temperature = 0.1
 max_tokens = 16000
 
 [agent.plan_controller.model]
 platform = "watsonx"
-model_name = "meta-llama/llama-4-maverick-17b-128e-instruct-fp8"
+model_name = "openai/gpt-oss-120b"
 temperature = 0.1
 max_tokens = 16000
 
 [agent.final_answer.model]
 platform = "watsonx"
-model_name = "meta-llama/llama-4-maverick-17b-128e-instruct-fp8"
+model_name = "openai/gpt-oss-120b"
 api_version ="2024-08-06"
 temperature = 0.1
 max_tokens = 32000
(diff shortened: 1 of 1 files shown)


src/cuga/configurations/models/settings.watsonx.toml
```diff
@@ -1,60 +1,60 @@
 [agent.task_decomposition.model]
 platform = "watsonx"
-model_name = "meta-llama/llama-4-maverick-17b-128e-instruct-fp8"
+model_name = "openai/gpt-oss-120b"
```
haroldship, line 3: can't we make a constant for the model name, and refer to it in each place?

haroldship (changes requested):
small nitpick but I think it's worth fixing, then I will approve :)

src/cuga/configurations/models/settings.watsonx.toml
segevshlomovIBM: We can, but I wanted to be consistent with the other setting tomel files (RITS, LITE, Open..)
It helps us to debug different agents if we need to
If you are passing the model name in the .env then it will override, so this is just the default when none is given

haroldship (approved):

PR #424, opened 2026-06-30
ci: reduce Windows stability tests to a single smoke case

sami-marreed:
## Summary
- Windows stability CI was taking ~30 minutes because it ran all 9 E2E tests sequentially (parallel was removed in ef2662c6 due to Windows subprocess contention).
- Adds a `--windows-smoke` flag and a dedicated `[[stability.windows_smoke]]` config with one fast-mode test (`test_list_my_accounts_fast`) that still exercises subprocess servers, dynamic port allocation, UTF-8 encoding, and agent streaming on Windows.
- Reduces the Windows job timeout from 60 to 15 minutes.

## Why not re-enable parallel?
Parallel on Windows was intentionally removed — subprocess spawning is slower and concurrent execution caused resource contention, often making parallel runs slower than sequential.

## Test plan
- [ ] Windows stability job completes in ~3–5 minutes instead of ~30
- [ ] Smoke test passes on `windows-latest`
- [ ] Ubuntu matrix (3 Python versions) still runs the full suite unchanged

## Summary by CodeRabbit

* **Tests**
  * Added a faster Windows smoke stability check to validate a small, targeted subset of the suite.
  * Reduced the runtime limit for this check so failures surface sooner and CI completes more quickly.
  * Improved test selection and error messaging when the Windows smoke path is used.

### DIFF
--- a/.github/workflows/stability-tests.yml
+++ b/.github/workflows/stability-tests.yml
@@ -88,10 +88,10 @@ jobs:
           retention-days: 7
 
   stability-tests-windows:
-    name: Stability Tests Windows (Python 3.12)
+    name: Stability Tests Windows Smoke (Python 3.12)
     if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
     runs-on: windows-latest
-    timeout-minutes: 60
+    timeout-minutes: 15
     strategy:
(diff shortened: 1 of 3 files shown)


laredo (approved):
Looks good to me

PR #426, opened 2026-07-01
feat(policies): ToolGuard UI viewer modal and per-policy enable/disable (#419, #421)

### LINKED ISSUE #419: [Feature] ToolGuard UI — view generated examples and code in a popup
## What you want and why

After generating a tool guard, there is currently no way to inspect what was generated from the UI. The shield icon changes colour to indicate a guard exists, but clicking it only triggers re-generation (or a toast saying guards already exist). Users need to be able to read the generated examples and policy code without leaving the Policies screen or digging into files.

A read-only popup on the shield icon would give immediate visibility into what the guard actually enforces, which is essential for review and trust before publishing.

## How it could work

- Clicking the shield icon on a policy that **already has all target tools guarded** opens a modal/side-panel instead of triggering generation.
- The modal shows a tabbed or accordion view per target tool, each containing:
  - **Violating examples** — the list of examples that should be blocked
  - **Compliance examples** — the list of examples that should be allowed
  - **Policy code** — the generated Python guard function, displayed in a read-only code block with syntax highlighting
- The modal has a **Close** button only (no editing).
- The existing `allTargetToolsHaveGuard` check in `PoliciesConfig.tsx` already identifies when to show this view — the `onClick` handler currently shows a toast; it should open the modal instead.

## Links / extra context

- Parent epic: #75
- Relevant code: `PoliciesConfig.tsx` — `allTargetToolsHaveGuard`, `onClick` handler on the shield `Button`
- Guard data is available on the policy object as `policy.tool_guards[tool].{violating_examples, compliance_examples, policy_code}`

### LINKED ISSUE #421: [Feature] ToolGuard UI — disable an individual tool guard after generation
## What you want and why

Once a tool guard is generated it is always active. There is no way to temporarily disable a single guard for a specific tool without deleting it entirely or disabling the whole policy. Users may want to test behaviour without a guard, or roll back a specific guard while keeping others active, without losing the generated code.

## How it could work

- Add a per-tool **enabled/disabled toggle** inside the tool guard view (either in the new popup from the related issue, or inline in the expanded policy row).
- The toggle maps to a new `enabled` boolean field on the `ToolGuard` model (default `true`).
- The ToolGuard runtime checks this field before enforcing the guard — if `enabled` is `false` the guard function is skipped for that tool.
- The shield badge/icon could visually distinguish a policy where some guards are disabled (e.g. a muted or strikethrough style) vs all active.
- Toggling saves via the normal policy save flow (no new endpoint needed).

## Links / extra context

- Parent epic: #75
- Relevant models: `ToolGuard` in `src/cuga/backend/cuga_graph/policy/models.py` — add `enabled: bool = True`
- Relevant runtime: `ToolGuardRuntime` in `tool_guard_runtime.py` — skip guard execution when `tool_guard.enabled is False`
- Related: ToolGuard UI popup issue (view guards)

offerakrabi:
## Feature

Closes #419
Closes #421

### Summary

**Issue #419 — ToolGuard viewer popup**

Clicking the shield icon on a `ToolGuidePolicy` that has all target tools guarded now opens a read-only viewer modal instead of showing a toast. The modal shows a tab per target tool, each containing:
- Violating examples (red tinted blocks)
- Compliance examples (green tinted blocks)
- Generated policy code (dark monospace block rendered via `marked`)

The shield icon turns green when guards are generated.

**Issue #421 — Per-policy guard enable/disable**

Added a `guards_enabled` boolean flag to `ToolGuidePolicy` (default `True`). When disabled:
- Guard enforcement is skipped at runtime — tool description enrichment remains active
- Shield icon turns red with tooltip "Guards generated but are disabled"
- Collapsed policy row badge changes from green "Guarded" to red "Guards disabled"
- A `Toggle` in the viewer modal header controls the flag; saved via the normal policy save flow

Also fixed two bugs discovered during testing:
- `guards_enabled` was being dropped in `apply_policies_data_to_storage` (field not passed through on save)
- `patch_draft_policies` route did not invalidate the ToolGuard runtime cache after saving, so toggling guards on/off had no effect until a full agent rebuild

### Testing
- [x] UI changes tested manually (viewer modal, toggle, shield states, badge states)
- [x] Guard disable/enable round-trip verified end-to-end

image
image

## Summary by CodeRabbit

* **New Features**
  * Added a per-policy toggle to enable or disable tool guard enforcement.
  * Introduced a “Tool Guard Viewer” modal to inspect violating/compliance examples and the related policy content.
  * Enhanced the policy UI with clearer guard status labels, colors, and updated viewer behavior.

* **Bug Fixes**
  * Draft policy updates now refresh tool-guard runtime behavior so updated settings take effect on subsequent tool calls.

* **Tests**
  * Added unit coverage to confirm disabled guards are skipped during guard registration/enforcement.


sami-marreed:
is this done with carbon mcp?

sami-marreed (approved):

offerakrabi:
> is this done with carbon mcp?

yes

PR #428, opened 2026-07-02
fix: deliver HITL response in tool-approval e2e resume helper

### LINKED ISSUE #427: [Bug]: tool-approval e2e helper never delivers HITL response; approval tests don't exercise post-approval execution
### Bug Description

The policy tool-approval e2e test helper `resume_graph_with_response` never actually delivers the human approval/denial to the graph. As a result, the three tool-approval e2e tests in `test_tool_approval_full_graph.py` (approve / deny / modification) **never exercise post-approval execution** — the approved code is never run, no tool is ever called, and the tests pass on the approval-prompt banner text rather than on a real completed answer.

The helper builds a throwaway local `AgentState`, sets `hitl_response` on it, and then resumes with `astream(None, …)` — so the response is dropped:

```python
# src/cuga/backend/cuga_graph/policy/tests/helpers.py  (resume_graph_with_response)
current_state = graph.graph.get_state(config)
updated_state = AgentState(**current_state.values)
updated_state.hitl_response = response          # set on a local copy…
async for event in graph.graph.astream(None, config, stream_mode="updates"):  # …never written back
    ...
```

Because the checkpoint is never updated with the response, on resume the graph re-runs from the interrupt with no approval recorded, regenerates code, and re-interrupts. `metadata["user_approved"]` is never set, so `ToolApprovalHandler.is_returning_from_approval()` returns `False` and `handle_approval_resumption()` (which routes the approved code to the sandbox) never runs.

### Steps to Reproduce

1. Add temporary logging to confirm execution (or reuse existing logs):
   - `ToolApprovalHandler.handle_approval_resumption` logs `"Returning from tool approval - skipping code generation, executing approved code"`.
   - `sandbox_node._record_weak_schema_shapes` (only reached after the approved code executes).
2. Run: `uv run python -m pytest src/cuga/backend/cuga_graph/policy/tests/test_tool_approval_full_graph.py -s`
3. Observe: `handle_approval_resumption` never logs; the sandbox execution/record path is never reached; the reported `final_answer` is the approval banner (`## ✋ Digital Sales Tool Approval …`) and contains none of the tool's data (e.g. `Acme Corp` / `acc_1`).

### Expected Behavior

`resume_graph_with_response` should deliver the `ActionResponse` to the graph via LangGraph's proper resume mechanism (e.g. `Command(resume=…)` or `graph.update_state(config, {"hitl_response": response})` before streaming), so that:
- `metadata["user_approved"]` is set on approval,
- `handle_approval_resumption` runs and executes the approved code through the sandbox,
- the approve test ends with a real answer built from the tool output, and the deny test actually cancels execution.

### Actual Behavior

The response is dropped. Approved code never executes; the tool is never called; the tests assert only on the approval-banner text (`final_answer` non-empty + not "cancelled"/"denied"), so they pass without verifying the approve/deny/modification behavior they claim to cover.

### Environment

- Repo: `cuga-project/cuga-agent`
- File: `src/cuga/backend/cuga_graph/policy/tests/helpers.py` (`resume_graph_with_response`), affecting `src/cuga/backend/cuga_graph/policy/tests/test_tool_approval_full_graph.py`
- Python 3.12, LangGraph checkpointer
- Pre-existing on `main` (independent of the weak-schema-probing work in #417; the approval handler and this helper are unchanged there).

### Error Logs / Screenshots

No crash — a silent gap. Confirmed via instrumentation: `handle_approval_resumption` and `_record_weak_schema_shapes` never fire during the approval e2e tests; `final_answer` is the approval banner with no tool data.

### Configuration

Standard test config (`settings.policy.enabled = True`, in-memory `DigitalSalesToolProvider`). No secrets involved.

### Additional Context

Surfaced while making the two tool-approval tests probing-aware in #417. The probing change only affected the *content* of the regenerated code (a structure probe with NL narration); it did not create this gap. Filing separately so the harness fix (make the tests genuinely deliver approvals and verify post-approval execution) can be scoped on its own — it touches a shared helper used by other policy e2e tests and may surface additional assertions to tighten.

sami-marreed:
## Summary
- Fixes `resume_graph_with_response` so tool-approval e2e tests actually deliver human approvals/denials to the graph via LangGraph's `Command(resume=...)` mechanism (matching the SDK resume path).
- Tightens the three tool-approval full-graph e2e tests to assert post-approval execution (tool output in final answer, no approval banner) and post-denial cancellation.

Fixes #427

## Problem
The helper built a local `AgentState`, set `hitl_response` on it, then called `astream(None, …)` without writing the response back to the checkpoint. On resume the graph re-interrupted without recording approval, so `handle_approval_resumption()` never ran and tests passed on the approval-banner text alone.

## Test plan
- [x] `uv run python -m pytest src/cuga/backend/cuga_graph/policy/tests/test_tool_approval_full_graph.py -s` — all three tests passed in CI (run 28581606688)
- [x] Verify approve flow final answer contains `Acme Corp` and not the approval banner
- [x] Verify deny flow final answer indicates cancellation and does not contain tool output
- [x] Verify modification flow approve step includes tool output after resume

## Summary by CodeRabbit

* **Tests**
  * Updated human-in-the-loop graph resume logic in test helpers to use standard interrupt/resume semantics, with looped resumption and bounded retry behavior.
  * Strengthened end-to-end tool-approval assertions across approve/deny/modify flows to consistently require the expected final answer and tool output content.
  * Added stricter checks ensuring the approval banner and cancellation/denial wording appear only in the correct scenarios.

### DIFF
--- a/src/cuga/backend/cuga_graph/policy/tests/helpers.py
+++ b/src/cuga/backend/cuga_graph/policy/tests/helpers.py
@@ -3,6 +3,7 @@
 import uuid
 from typing import Dict, List, Any, Optional
 from langchain_core.messages import HumanMessage
+from langgraph.types import Command
 
 from cuga.backend.cuga_graph.policy.storage import PolicyStorage
 from cuga.backend.cuga_graph.policy.configurable import PolicyConfigurable
@@ -461,13 +462,11 @@ async def resume_graph_with_response(
     """
     config = graph.get_config_with_policy({"configurable": {"thread_id": thread_id}})
 
-    # Update the state with the response
-    current_state = graph.graph.get_state(config)
-    updated_state = AgentState(**current_state.values)
-    updated_state.hitl_response = response
-
-    # Stream events until completion or next interrupt
-    async for event in graph.graph.astream(None, config, stream_mode="updates"):
+    async for event in graph.graph.astream(
+        Command(resume=response.model_dump()),
+        config,
+        stream_mode="updates",
+    ):
         if isinstance(event, tuple):
             namespace, state_dict = event
             if "__interrupt__" in state_dict or state_dict.get("__interrupt__"):
--- a/src/cuga/backend/cuga_graph/policy/tests/helpers.py
+++ b/src/cuga/backend/cuga_graph/policy/tests/helpers.py
@@ -452,6 +452,11 @@ async def resume_graph_with_response(
 ) -> Any:
     """Resume the graph with a human response.
 
+    Delivers the response via LangGraph's interrupt/resume protocol
+    (``Command(resume=...)``), matching the SDK path. When the user
+    confirmed approval, may resume more than once: follow-up codegen
+    (e.g. structure probes) can re-trigger tool approval on the same thread.
+
     Args:
         graph: DynamicAgentGraph instance
         thread_id: Thread identifier
@@ -459,11 +464,18 @@ async def resume_graph_with_response(
 
     Returns:
         Final state snapshot after resuming
+
+    Raises:
+        RuntimeError: If the graph keeps interrupting after repeated confirmed resumes.
     """
     config = graph.get_config_with_policy({"configurable": {"thread_id": thread_id}})
     resume_payload: Any = Command(resume=response.model_dump())
+    max_resume_attempts = 5
 
-    while True:
+    for attempt in range(max_resume_attempts):
+        # Drain the stream so LangGraph runs nodes and updates the checkpoint.
+        # We don't inspect individual events; the authoritative state is read
+        # from get_state() after the stream finishes (or hits an interrupt).
         async for _event in graph.graph.astream(
             resume_payload,
             config,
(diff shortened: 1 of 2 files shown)


src/cuga/backend/cuga_graph/policy/tests/helpers.py
```diff
@@ -460,24 +461,26 @@ async def resume_graph_with_response(
...
-                break
-        elif "__interrupt__" in event:
-            break
-
-    # Get the final state
-    state_snapshot = graph.graph.get_state(config)
-    return state_snapshot
+    resume_payload: Any = Command(resume=response.model_dump())
+
+    while True:
```
haroldship: it's not clear what this while loop is doing, nor what pulling all the events from the stream without processing is for. perhaps a comment?

haroldship (changes requested):
Requesting changes:
- None of the testing plan in the description is checked. Please check.
- I don't follow the looping logic, some comments explaining why things are done like that would help

sami-marreed:
Addressed review feedback in b6aa231c:

- **Loop comments:** Documented why we drain `astream` (LangGraph needs the stream consumed to advance the checkpoint; we read authoritative state via `get_state()` afterward) and why confirmed approvals may resume multiple times (structure-probe codegen can re-trigger tool approval on the same thread).
- **Bounded retries:** Capped at 5 confirmed resume attempts with a clear `RuntimeError` if the graph keeps interrupting.
- **Test plan:** Checked off all items — the three `test_tool_approval_full_graph.py` tests passed in CI run 28581606688.

haroldship (approved):

Issue #401, opened 2026-06-29
Cuga-lite extractor should dispatch OpenAI-style tool-call JSON as a fallback


Sergey-Zeltyn:
### Summary

Tool-call-trained models (Nemotron-3-Super, several Qwen tool-calling variants, etc.) routinely emit OpenAI-style function-call arguments in the assistant's `content` field rather than fenced Python code. Example:

```json
{
  "query": "search emails unread priority inbox",
  "app_name": "gmail"
}
```

That's exactly the args of cuga-lite's built-in `find_tools(query, app_name)` helper — but expressed in tool-call shape. Cuga-lite's extractor looks for `` ```python … ``` `` only, finds nothing, classifies the response as `Assistant_nl`, re-prompts, and the model produces the same JSON again. The loop runs until the 70-step ceiling.

### Why this is worth a distinct change

Issue #386 covers the general "unproductive loop" guard. This issue is the **happy-path complement**: when the model's "wrong-shape" output is structurally a valid tool call, dispatching it is strictly better than terminating. The model already picked the right tool and the right args; it just used a different syntax.

### Observed cases

From the most recent nemotron sweeps on AppWorld easy (43 tasks):

| Run | Failures with JSON-args mode | Notable repeat-offender task |
|---|---:|---|
| `20260628_143824_compare_nemotron_easy` (temp 0.1, top_p 1.0) | 12 / 20 failures | `e775c78_1` — 28 JSON turns, 19 byte-identical repeats, 56 LLM calls |
| `20260628_175529_compare_nemotron_easy_temp_06` (temp 0.6, top_p 0.95) | 4 / 15 failures | `e775c78_1` — still loops, 11 JSON turns |

Lowering sampling variance helps but doesn't fix it — `e775c78_1` (gmail priority-1 count) tips Nemotron into tool-call mode reliably regardless of temperature/top_p.

### Proposed fix

Add a JSON-args fallback to cuga-lite's code extractor. Pseudo-code at the extraction call site:

```python
match = extract_python_block(raw_response)
if match:
    run_python(match)
else:
    parsed = _try_parse_tool_call(raw_response)
    if parsed and parsed.tool_name in INTERNAL_TOOL_SIGNATURES:
        dispatch_tool(parsed.tool_name, parsed.args)
    else:
        # existing path: classify as Assistant_nl, re-prompt
```

Where `_try_parse_tool_call(text)`:
[...]

Issue #394, opened 2026-06-29
top_k sampling parameter cannot be set on LLM calls


Sergey-Zeltyn:
### Summary

CUGA has no way to set the **top_k** sampling parameter on LLM calls. It's not in any model TOML, not passed by any platform branch in `LLMManager`, and not accepted by `langchain_openai.ChatOpenAI` natively. For models that publish recommended sampling recipes including a top_k value (notably NVIDIA Nemotron-3 / -4 reasoning models, several Qwen variants, and DeepSeek models), runs go out at the server-side default (typically `top_k = -1`, i.e. no top-k truncation).

### What is top_k?

Top-k is one of three standard nucleus-sampling controls used during token generation:
- **temperature** scales the token logits before softmax (lower = more deterministic).
- **top_p** keeps the smallest set of tokens whose cumulative probability reaches *p*, then samples from that set.
- **top_k** keeps only the *k* highest-probability tokens at each step, then samples from those.

`top_p` is "smallest set ≥ p mass"; `top_k` is "top k regardless of mass." They're often combined — model vendors publish recommended `(temperature, top_p, top_k)` triples to control the trade-off between diversity and reliability for their specific post-training.

### Why CUGA can't set it today

- `ChatOpenAI(top_k=...)` signature check returns `False` — the OpenAI Chat Completions API spec doesn't include top_k.
- vLLM-hosted endpoints (such as RITS or AWS Bedrock-fronted vLLM) **do** accept top_k, but only via OpenAI SDK's `extra_body={"top_k": ...}` escape hatch.
- None of CUGA's platform branches in [`LLMManager._create_llm_instance`](src/cuga/backend/llm/models.py) pass `extra_body`. The `rits` and `rits-restricted` branches currently set temperature and (now) top_p but never top_k.

The 11 `top_k` references that grep finds in CUGA today (`prompt_utils.py`, `bind_tools/cap.py`) are for tool shortlisting — same word, unrelated to LLM sampling.

### Concrete impact

NVIDIA's published sampling recipe for `NVIDIA-Nemotron-3-Super-120B-A12B-FP8`:

| Param | Recommended | Settable in CUGA today |
|---|---|---|
| temperature | 0.6 | ✅ (after recent fix) |
| top_p | 0.95 | ✅ (after recent fix) |
| **top_k** | **20** | ❌ |

The same gap applies to any future model whose sampling recipe includes top_k.

### Proposed fix

[...]

Issue #386, opened 2026-06-28
[Enhancement]: Cuga-lite has no guard against unproductive iteration loops


Sergey-Zeltyn:
### Summary

When the model's output can't be parsed as a Python code block by cuga-lite's extractor, the `prepare → call_model ↔ execute` loop re-prompts up to the 70-step / token ceiling regardless of whether subsequent responses are byte-identical or match a parseable tool-call shape. The only safety net is the step limit, which is a *cost ceiling*, not a *quality signal*. Observed reproducibly across multiple models on AppWorld; consumes 30–130+ LLM calls per stuck task before the loop terminates.

### Observed cases (from recent evaluation bundles)

All on `cuga-lite` against AppWorld, lite_mode=true, with the agent-step ceiling at 70:

| Bundle | Model | Task | Symptom |
|---|---|---|---|
| `20260628_121656_compare_nemotron` | nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-FP8 | `e775c78_1` | 15 consecutive **byte-identical** JSON payloads `{"query":"search emails","app_name":"gmail"}`, 30 LLM calls, never converged |
| `20260624_153140_compare_gemma_easy_run1` | google/gemma-4-31B-it | `a30375d_1` | **132 LLM calls**, hit 70-step ceiling, agent message *"Maximum step limit (70) reached"* |

The common thread: the model produces output the cuga-lite parser can't extract code from (bare JSON, NL-only, malformed code), the loop re-prompts, the model produces the same shape again, no learning signal flows back, and the loop runs to the cost ceiling.

### Why this matters

- **Cost**: 30–130 wasted LLM calls per stuck task. On harder eval sets, mean per-task duration is dominated by 1–2 outliers (we saw run averages skewed from 80 s to 143 s by a single stuck task).
- **Eval signal**: step-limit blowups look identical to "model failed" in reports, but the root cause (model+parser format mismatch, not reasoning failure) is different and the remediation is different.
- **Debuggability**: the trajectories show identical content for 14 turns. There is no logging or telemetry that flags "model emitted the same response N times in a row" — a future-you reading the bundle has to grep manually.

### Proposed fixes

Two small, independently useful guards. Both apply to every model (no per-provider logic):

1. **Byte-equality repetition guard in `call_model`.**
[...]

PR #430, opened 2026-07-02
feat(rits): direct RITS connections with .env-driven model swapping

### LINKED ISSUE #374: [Enhancement]: Support direct RITS connections and .env-driven model swapping
- Enable two presets:
1. `settings.rits.toml` — direct RITS connection.
2. `settings.rits.local.toml` — local LiteLLM-proxy connection (existing behaviour preserved).
- Enable `.env` overriding for MODEL_NAME and RITS_BASE_URL win over the TOML, so swapping models is a one-line `.env` edit.

Sergey-Zeltyn:
Closes #374.

## Summary

Enables CUGA to talk to the RITS endpoint directly (without a local LiteLLM proxy in front of it) and lets users swap the RITS model or endpoint per run via `.env` without editing every per-role TOML block.

Before this PR, the `platform == "rits"` code path only worked behind a local LiteLLM proxy: a direct RITS call returned HTTP 403 *"Authentication parameters missing"* because RITS requires a custom `RITS_API_KEY` header, and the `ChatOpenAI` client we use sends only `Authorization: Bearer`. This PR:

- **Wires the `RITS_API_KEY` custom header** through langchain-openai's `default_headers`, so direct RITS auth works end-to-end.
- **Adds `platform == "rits"` branches** to `_get_model_name` and `_get_base_url` so `MODEL_NAME` and `RITS_BASE_URL` in `.env` override the TOML. This makes swapping between RITS-hosted models (e.g. gemma → gpt-oss → nemotron) a two-line `.env` change instead of 14 per-role TOML edits.
- **Fixes the rits constructor** to consume the already-resolved `base_url` local variable rather than re-reading `model_settings.get('url')`, so the new env override actually reaches `ChatOpenAI`.

## Presets shipped

- **`settings.rits.toml`** — direct RITS preset. Defaults to `google/gemma-4-31B-it` at the `/v1` endpoint on the RITS 3scale apicast host, cuga-baseline sampling (`temperature=0.1`, `top_p=1.0`).
- **`settings.rits.proxy.toml`** — local LiteLLM proxy preset (`http://localhost:4000`, `rits/openai/gpt-oss-120b`), for teams already running their own proxy in front of RITS.

Note the deliberate deviation from the issue text: the local-proxy preset is named `settings.rits.proxy.toml`, **not** `settings.rits.local.toml`. Dynaconf silently auto-merges any `*.local.*` file on top of its non-local sibling, which would cause the two presets to override each other rather than coexist. The `.proxy.toml` suffix escapes that convention while remaining explicit about the setup.

## Docs

Adds *"Option 7: RITS Support"* to the LLM Configuration section of the README, mirroring the format used for Options 1-6 (OpenAI / WatsonX / Azure / LiteLLM / Groq / OpenRouter). Extends the Configuration Files list to include both new presets.

## Test plan

- [x] `uv run ruff check src/cuga/backend/llm/models.py` — passes
- [x] `uv run ruff format --check src/cuga/backend/llm/models.py` — passes
- [x] Runtime smoke: model factory constructs the direct-RITS `ChatOpenAI` with the correct `base_url`, `temperature=0.1`, `top_p=1.0`, and the sampling-config INFO log line surfaces the effective values.
- [x] End-to-end AppWorld eval runs on the `gemma-4-31B-it` direct preset — many bundles produced during development; consistent behavior across easy / med / hard task splits.
- [x] `.env` override path validated: setting `MODEL_NAME` / `RITS_BASE_URL` in cuga-eval's `.env` reaches the request body (verified via Langfuse `modelParameters`).

## Related follow-ups

Not blocking this PR but worth linking:

- #386 — cuga-lite has no guard against unproductive iteration loops
- #401 — Cuga-lite extractor should dispatch OpenAI-style tool-call JSON as a fallback
- #394 — `top_k` sampling parameter cannot be set on LLM calls
- #405 — TOML-defined sampling params honored (merged; this PR builds on it)

## Summary by CodeRabbit

* **New Features**
  * Added RITS-based model configuration presets for both direct endpoint and local proxy routing.
  * Added standardized model settings for agent and memory components (including token limits and temperature).

* **Documentation**
  * Documented **Option 7: RITS Support**, including `RITS_API_KEY` setup and `.env` configuration for direct and proxy modes.
  * Clarified how updating `MODEL_NAME` may require updating `RITS_BASE_URL` to avoid reusing the previous URL path.
  * Updated the “Configuration Files” list to include the new `settings.rits.toml` and `settings.rits.proxy.toml`.

### DIFF
--- a/src/cuga/configurations/models/settings.rits.toml
+++ b/src/cuga/configurations/models/settings.rits.toml
@@ -1,81 +1,127 @@
 [agent.task_decomposition.model]
 platform = "rits"
-model_name = "rits/openai/gpt-oss-120b"
-url="http://localhost:4000"
+model_name = "google/gemma-4-31B-it"
+url="https://inference-3scale-apicast-production.apps.rits.fmaas.res.ibm.com/google-gemma-4-31b-it/v1"
 apikey_name = "RITS_API_KEY"
 temperature = 0.1
+top_p = 1.0
 max_tokens = 1000
 
 [agent.planner.model]
 platform = "rits"
-model_name = "rits/openai/gpt-oss-120b"
-url="http://localhost:4000"
+model_name = "google/gemma-4-31B-it"
+url="https://inference-3scale-apicast-production.apps.rits.fmaas.res.ibm.com/google-gemma-4-31b-it/v1"
 apikey_name = "RITS_API_KEY"
 temperature = 0.1
+top_p = 1.0
 max_tokens = 7000
 
 [agent.shortlister.model]
 platform = "rits"
-model_name = "rits/openai/gpt-oss-120b"
-url="http://localhost:4000"
+model_name = "google/gemma-4-31B-it"
+url="https://inference-3scale-apicast-production.apps.rits.fmaas.res.ibm.com/google-gemma-4-31b-it/v1"
 apikey_name = "RITS_API_KEY"
 temperature = 0.1
+top_p = 1.0
 max_tokens = 7000
 
 
 [agent.chat.model]
 platform = "rits"
-model_name = "rits/openai/gpt-oss-120b"
-url="http://localhost:4000"
+model_name = "google/gemma-4-31B-it"
+url="https://inference-3scale-apicast-production.apps.rits.fmaas.res.ibm.com/google-gemma-4-31b-it/v1"
 apikey_name = "RITS_API_KEY"
 temperature = 0.1
+top_p = 1.0
 max_tokens = 5000
 
 [agent.plan_controller.model]
 platform = "rits"
-model_name = "rits/openai/gpt-oss-120b"
-url="http://localhost:4000"
+model_name = "google/gemma-4-31B-it"
+url="https://inference-3scale-apicast-production.apps.rits.fmaas.res.ibm.com/google-gemma-4-31b-it/v1"
 apikey_name = "RITS_API_KEY"
 temperature = 0.1
+top_p = 1.0
 max_tokens = 5000
 
 
 [agent.final_answer.model]
 platform = "rits"
-model_name = "rits/openai/gpt-oss-120b"
-url="http://localhost:4000"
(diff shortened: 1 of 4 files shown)


src/cuga/configurations/models/settings.rits.toml
```diff
@@ -1,81 +1,127 @@
...
 temperature = 0.1
+top_p = 1.0
 max_tokens = 1000
 
 [agent.planner.model]
 platform = "rits"
-model_name = "rits/openai/gpt-oss-120b"
-url="http://localhost:4000"
+model_name = "google/gemma-4-31B-it"
+url="https://inference-3scale-apicast-production.apps.rits.fmaas.res.ibm.com/google-gemma-4-31b-it/v1"
```
sami-marreed, line 13: if user changes model name in .env will also change the url here?

Sergey-Zeltyn: @sami-marreed 
no, MODEL_NAME and RITS_BASE_URL are independent env vars. Changing MODEL_NAME in .env does not rewrite the URL in the TOML.

The reason both are exposed as separate overrides is that RITS's URL path is model-specific — each hosted model gets its own `<host>/<model-slug>/v1` endpoint. Swapping between two RITS-hosted models therefore requires updating both env vars together; setting only `MODEL_NAME` would silently leave you pointed at the previous model's URL.

For context on how I've been using this in practice: I put all three — `RITS_API_KEY, MODEL_NAME`, and `RITS_BASE_URL` — into cuga-eval's `.env` and run with `--dotenv`. That's the intended primary use case for these overrides: the TOML defines a sensible default preset, and the .env overrides let me swap models/endpoints per run without touching 14 per-role TOML blocks.

I've pushed **62be9a55** with a README clarification that spells the coupling out explicitly

sami-marreed (approved):

PR #433, opened 2026-07-02
feat(supervisor): multi-agent registry, composition UI, and supervisor runtime

### LINKED ISSUE #101: [Design] CugaSupervisor frontend — manage endpoint UX
## What you want and why

The CUGA frontend today is built around a single-agent experience: one agent, one set of policies, one chat interface. As the supervisor (multi-agent orchestration) matures on the backend, the frontend needs a first-class way to:

1. **Configure** a supervisor — define which sub-agents it orchestrates, set supervisor-level policies, and toggle plan approval.
2. **Connect** to the right endpoint — the supervisor exposes a different entry point than a standalone agent; the frontend must know which mode it is in and route requests accordingly.
3. **Experience** the supervisor at runtime — show the operator/user which sub-agents are running, display the plan approval gate when it fires, and surface per-agent status inline.

This issue covers **proposing a design and then implementing it**, keeping the manage-endpoint experience central to the proposal.

#### 1. Manage endpoint integration
- How does the frontend discover whether it is connected to a standalone `CugaAgent` or a `CugaSupervisor`?
- Should the `/manage` endpoint return a capability manifest (e.g. `{ "mode": "supervisor", "agents": [...], "policies": [...] }`) that the UI reads to adapt its layout?
- What changes, if any, are needed in the manage endpoint to support supervisor metadata?

#### 2. Supervisor configuration screen
- Where in the existing UI does supervisor setup live? (New top-level tab? Extension of the existing agent config panel?)
- How does a user add/remove sub-agents, set their order, and assign routing rules?
- How are supervisor-level policies (intent guard, playbook, routing, sequencing, approval) configured visually — reuse the existing policy panel or a new supervisor-specific one?
- Plan approval toggle: where does it live and what does enabling it look like at runtime?

#### 3. Runtime / chat experience
- When the supervisor is active, how is the multi-agent execution surfaced? (Collapsible sub-agent trace? Step-by-step timeline? Status chips?)
- Plan approval gate: how does the UI pause, display the proposed plan, and collect operator approval before the supervisor dispatches sub-agents?
- Error and partial-failure states: one sub-agent fails — what does the user see?

#### 4. Mode switching
- Can a user toggle between standalone-agent mode and supervisor mode in the UI, or is the mode fixed at deploy time via the manage endpoint?

### Proposed deliverables

1. **Design document / wireframes** — covering all four areas above, with a recommended approach for the manage endpoint contract
2. **Manage endpoint changes** — any backend additions needed to surface supervisor capabilities to the frontend
3. **Frontend implementation** — supervisor config screen, runtime trace, and plan approval UX

## Links or extra context

- Supervisor policy system v1 (SDK): #100
- Policy document parser epic: #99
- Existing manage endpoint and policy UI are the reference starting points — the design must be additive and not break the standalone-agent experience

sami-marreed:
## Feature

Closes #101

### Summary

Adds a multi-agent registry to the manage UI (create/list/delete agents, single or supervisor kind), a supervisor composition screen for adding internal or A2A sub-agents, and wires the supervisor into the live chat runtime end-to-end.

**Backend registry**
- `GET/POST /api/agents`, `DELETE /api/agents/{id}` — list/create/delete agents (`agents_routes.py`)

**Manage UI**
- Dashboard: create/delete agents, "Supervisor" tag, per-agent Chat link
- Composition screen: Sub-agents panel (add internal agent or A2A endpoint), plan-approval toggle, persisted via a new `PATCH /api/manage/config/draft/supervisor`

**Supervisor runtime**
- Per-request `X-Agent-ID` resolution in `/stream` with a built-graph cache, invalidated on draft-save/publish
- `DynamicAgentGraph` takes a per-agent `supervisor_agents`/`supervisor_enabled` override instead of only reading the global `settings.supervisor.enabled` YAML toggle
- **Root-cause routing fix**: `TaskAnalyzer.should_use_supervisor_mode` only ever checked the *global* setting, so a per-agent supervisor never actually routed into the `CugaSupervisor` node at runtime even though the graph was built correctly. Added a `state.supervisor_mode` override (same pattern as `lite_mode`) that `event_stream` sets when the resolved agent is a supervisor.
- Plan-approval HITL gate reusing the existing `interrupt()`/`ActionResponse` path (`create_agent_approval_action`, already had frontend support in the chat card renderer). Required adding `AgentState.supervisor_metadata`, which the resume flow depends on but which didn't exist on the outer state.
- Fixed a silent bug where hyphenated agent ids (e.g. `crm-agent`, the normal case for any UI-created agent) produced `delegate_to_crm-agent`, an invalid Python identifier that parses as subtraction — breaking both delegation itself and the plan-approval gate's code-scan matcher.

**Chat UI**
- `/chat/:agentId` route, an agent switcher in the chat header, `X-Agent-ID` threaded through `postStream` and the conversation endpoints (previously hardcoded to `cuga-default`)

**CLI**
- `cuga start manager --seed-supervisor-demo` preloads a CRM/email/filesystem sub-agent trio + supervisor (draft + published) and provisions their tools, for local testing without manual setup

**Test infra**
- `tests/unit/conftest.py` and `src/cuga/sdk_core/tests/conftest.py`: autouse fixture that isolates DB-touching tests to a temp `DBS_DIR` instead of the shared local `cuga.db` a running dev server also uses (several tests call `reset_config_db()`, which previously could delete/corrupt a live server's database)

### Testing
- [x] Tested locally; unit tests pass (`tests/unit/`, `src/cuga/sdk_core/tests/`, `src/cuga/backend/cuga_graph/nodes/cuga_agent_core/tests/`) — no regressions vs. the pre-existing baseline (verified via `git stash -u` diffing)
- [x] `ruff check` clean on all touched files
- [x] Frontend `tsc`/webpack build clean (same pre-existing baseline errors, no new ones)
- [x] Live-verified end-to-end with a real LLM (Groq) against an isolated throwaway DB: supervisor routing, delegation, execution-output streaming, and the hyphenated-agent-id fix
- [ ] Plan-approval pause → approve → resume round-trip — reconstructed from reading the LangGraph control flow and fixed real gaps found along the way, but not exercised against a live UI click-through. Please test before relying on it.

## Summary by CodeRabbit

- **New Features**
  - Added an agent registry for creating, viewing, deleting, and managing single or supervisor agents.
  - Added supervisor configuration for internal and A2A sub-agents, delegation, and optional plan approval.
  - Added agent-specific chat routes, conversation history, streaming, and direct navigation.
  - Added approval prompts before supervisor delegation, with approve and deny actions.
  - Added optional demo setup for CRM, email, and filesystem agents.

- **Bug Fixes**
  - Improved configuration isolation, graph caching, agent switching, and supervisor plan resumption.
  - Improved validation when importing supervisor and sub-agent configurations.


offerakrabi (approved):
## PR Review: #433 — feat(supervisor): multi-agent registry, composition UI, and supervisor runtime

@sami-marreed this looks solid, just one behavioral question i had, it looks like we persist the memory for sub-agents between runs. is this on purpose? is it the behavior we want?

**Merge.** The approach is sound — the per-agent graph cache, routing fix, and plan-approval gate all hang together correctly. The singleton isolation (no non-default publish mutating `app_state`) is well-tested.
Nothing blocks. #1 is the author's call.

|---|---|---|---|---|
| 1 | Minor | O(n) referrer scan on every draft-save/publish | `manage_routes/helpers.py:_referrer_ids_for_agent` | Loads every agent's config from the DB on each save to find supervisors referencing the changed sub-agent → fine for a small registry, but grows linearly with agent count. |

## Questions

1. Sub-agents are invoked with a fixed thread ID `supervisor_conversational_{agent_name}`, shared across all users and all supervisor runs for the lifetime of the process. If `crm-agent` produced wrong output in one user's session — say it hallucinated a customer record — the next delegation from any other user's supervisor starts from a conversation history that already contains that bad context. Is this intentional, or should thread IDs be scoped per supervisor turn or per user?

2. `PATCH /config/draft/supervisor` is the only endpoint the composition UI can use to update sub-agents and plan-approval. The same endpoint does not require `registry_enabled=true` to be called — it only checks the flag at the top, so it returns 404 when the registry is off. Is there a plan for supervisors configured via the global YAML path to also use this endpoint, or does that path remain YAML-only?

## Notes

- Plan-approval HITL was not verified against a live UI click-through (PR body explicitly flags this). The logic reads correctly and the unit tests cover the state-machine paths, but a live round-trip should be confirmed before this feature is relied on in production.

PR #434, opened 2026-07-02
fix: reject invalid generated Python before sandbox exec

### LINKED ISSUE #438: [Feature] Fail-fast syntax validation for invalid LLM-generated CodeAgent Python
## What you want and why

CugaLite CodeAgent runs LLM-generated Python in a sandbox via `exec()`. When the model produces **invalid syntax**—especially truncated or malformed triple-quoted f-strings while building long markdown/JSON reports—the failure happens at runtime with a noisy traceback and no actionable guidance for the model to retry.

**Observed failure (Instana/ServiceNow multi-phase report workflow):**
- Phase 1 succeeded with an inline `phase1_md = f"""..."""` block.
- Phase 2 failed with `SyntaxError: unterminated triple-quoted f-string literal` (~919-char script, never closed `"""`).
- Previously, `SecurityValidator.validate_imports` logged a syntax warning and **continued to exec anyway**, wasting a turn and returning unhelpful output.

**Why it matters:**
- Valid f-strings should keep working (e.g. `print(f"total={n}")`, short report snippets).
- Long report blobs in `f"""..."""` are a high-failure LLM pattern (truncation, unescaped `{` from JSON, quote nesting)—but the platform should fail fast with a clear hint rather than executing broken code.
- Auto-repairing unterminated strings is unsafe (silent wrong reports); detect-and-retry is the right model.

## How it could work

1. **Pre-exec syntax guard** — `compile(code, flags=PyCF_ALLOW_TOP_LEVEL_AWAIT)` before sandbox wrap/exec; reject invalid Python early.
2. **Actionable errors** — line number, offending line, hint for unterminated f-strings / bad indentation (prefer `'\n'.join([...])`, `json.dumps()`, or `await write_file(...)`).
3. **Typed exception** — `CodeSyntaxError` instead of string-matching error messages in `LocalExecutor.format_error`.
4. **Prompt guidance** — rule discouraging long markdown/JSON in triple-quoted f-strings inside inline code blocks (short f-strings remain fine).
5. **Agent retry loop** — syntax error returned as execution output → model self-corrects on next `call_model` turn.

**Out of scope:** auto-closing quotes, auto-escaping braces in f-strings (fragile, risks wrong output).

## Links or extra context

- Fix PR: https://github.com/cuga-project/cuga-agent/pull/434
- Branch: `fix/code-syntax-guard-before-exec`
- Key files: `executors/common/security.py`, `executors/code_executor.py`, `prompts/mcp_prompt.jinja2`
- Related: policy DB `cuga_policies` missing table was a separate non-blocking warning in the same logs

sami-marreed:
Fixes #438

## Summary
- Add a syntax guard that rejects invalid LLM-generated Python **before** sandbox `exec()`, instead of logging a warning and running anyway
- Return clean, actionable errors (line number + hint for unterminated f-strings / bad indentation) so the agent can retry with `write_file` or `'\n'.join()` patterns
- Allow top-level `await` in validation (`PyCF_ALLOW_TOP_LEVEL_AWAIT`) so normal CodeAgent scripts still pass

Fixes the failure mode seen when the model generates truncated `phase2_md = f"""...` report blocks (SyntaxError at exec time with noisy tracebacks).

## Test plan
- [x] `uv run pytest src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/tests/test_code_executor.py`
- [ ] Re-run Instana/ServiceNow multi-phase report workflow; confirm syntax errors surface with hint and agent self-corrects on retry
- [ ] Confirm normal async tool calls (`await find_tools(...)`, etc.) still execute


src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/common/security.py
```diff
@@ -96,6 +95,37 @@ class SecurityValidator:
         (r'\\u[0-9a-fA-F]{4}', 'unicode encoded string (potential obfuscation)'),
     ]
 
+    @staticmethod
+    def format_syntax_error(code: str, exc: SyntaxError, *, filename: str = "<code>") -> str:
```
offerakrabi: i don't think `filename` is used in this function

sami-marreed: Fixed in 4d459e59 — removed the unused `filename` param from `format_syntax_error`. `validate_syntax` still passes `filename` to `compile()` where it belongs.

sami-marreed:
Updates addressing review (HEAD `4d459e59`):

**Addressed**

- **offerakrabi** — Removed unused `filename` from `format_syntax_error`; `validate_syntax` still uses it for `compile()`.
- **CodeRabbit** — Replaced fragile `"Python syntax error" in str(error)` check with a dedicated `CodeSyntaxError` type (`security.py` → `code_executor.py` / `local_executor.py`).

**Also in this push**

- Added **MULTILINE REPORTS** rule (#10) to `mcp_prompt.jinja2` — do not build long markdown/JSON in triple-quoted f-strings; use `join` / `json.dumps` / `write_file`.

Ready for another look.

sami-marreed:
@offerakrabi — following up on the syntax guard / f-string discussion, here's how we're thinking about it for this PR.

**TL;DR:** Valid f-strings are fully supported at runtime. We're not blocking them. The guard only rejects *invalid* Python (syntax errors). The prompt rule + error hints steer the model away from one specific high-failure pattern: huge report blobs in triple-quoted f-strings.

---

### What already works

The sandbox runs any code that passes `compile()` — including f-strings. Phase 1 in the Instana/ServiceNow run actually succeeded with an inline `phase1_md = f"""..."""` block. So small/medium f-strings with a few interpolations are fine and should stay that way.

Examples we want to keep:
```python
print(f"open_sev10_untracked={summary['instana_events']['open_sev10_untracked']}")
phase1_md = f"# Phase 1\n- Total: {summary['instana_events']['total']}\n"
```

---

### What broke in Phase 2 (and why we're cautious)

Phase 2 failed with `unterminated triple-quoted f-string` — the generated script was ~919 chars and never closed the `"""`. That's not "f-strings don't work"; it's the model producing *invalid* Python.

The pattern we're discouraging (not forbidding) looks like:
```python
phase2_md = f"""
# Phase 2 — long markdown...
{some_dict_or_json}   # braces, quotes, backticks everywhere
... hundreds more lines ...
"""
```

This shape fails often for a few reasons:

1. **Truncation** — long inline code blocks get cut off before the closing `"""`.
2. **Brace escaping** — in f-strings, `{` starts an expression. JSON/dict literals need `{{`/`}}` everywhere unless you're only interpolating a couple of variables. One missed brace → SyntaxError.
3. **Quote nesting** — markdown + JSON + triple quotes is easy to break.
4. **High cost of failure** — a broken one-liner f-string loses one print; a broken report f-string loses an entire phase.

Safer alternatives for the model (and what the error hint points to):
```python
lines = ["# Phase 2", f"- Untracked: {n}", json.dumps(section, indent=2)]
content = "\n".join(lines)
await write_file("./output/phase2.md", content)
```

---

### What we're *not* doing (and why)

We considered "supporting" broken f-strings (auto-closing quotes, auto-escaping braces, etc.) and decided against it — auto-repair would turn syntax errors into silently wrong reports, which is worse than a clean retry.

So the stack for this PR is:
- **Runtime:** valid f-strings run; invalid code → `CodeSyntaxError` with a hint
- **Prompt:** rule #10 nudges away from long report f-strings specifically
- **Retry:** model self-corrects on the next turn

Happy to adjust the prompt wording if you think it's too broad — e.g. we could soften it to "for multiline *reports*, prefer join/write_file; short f-strings for single values are fine" instead of anything that reads like a blanket ban.

offerakrabi (approved):

PR #437, opened 2026-07-05
fix(watsonx): correct gpt-oss-120b context profile to prevent negative max_tokens

### LINKED ISSUE #439: [Bug]: watsonx gpt-oss-120b fails with negative max_tokens on long CugaLite runs
## What you want and why

Long CugaLite sessions against watsonx with `openai/gpt-oss-120b` fail mid-run with:

```
Invalid input argument for Model 'openai/gpt-oss-120b': max_tokens must be at least 1, got -5178.
```

The agent should keep running on large investigations (many tool calls, large variable payloads, context summarization) without the LLM call failing because watsonx receives an invalid `max_tokens` value.

## How it could work

**Root cause:** `ChatWatsonx` (langchain_ibm) ships with a stale `profile.max_input_tokens = 8192`. langchain_ibm computes `max_tokens = profile.max_input_tokens - input_tokens`. Once the prompt exceeds 8K tokens (~13K in the failing case), `max_tokens` goes negative and watsonx returns HTTP 400.

Context summarization made this worse: it triggered at “1407% usage” against the wrong 8K window instead of the model’s real 131K context, and `_setup_model_profile` skipped correcting the profile when one already existed.

**Expected fix:**
- Prefer known context sizes for `gpt-oss-120b` (131072) over provider-supplied profiles
- Set the correct profile when creating `ChatWatsonx` instances
- Resolve `model_id` (not just `model_name`) for watsonx in context summarization
- Merge existing profile keys when updating `max_input_tokens`

## Links or extra context

- Fix in progress: https://github.com/cuga-project/cuga-agent/pull/437
- Observed on watsonx `openai/gpt-oss-120b` with context summarization enabled (`trigger_fraction = 0.75`)
- Example log line before failure: `Total context tokens: 115296 (1407.4% usage)`

sami-marreed:
Fixes #439

## Summary
- Fix watsonx `max_tokens must be at least 1, got -5178` failures when using `openai/gpt-oss-120b` on long CugaLite runs
- `ChatWatsonx` from langchain_ibm ships with a stale 8K `max_input_tokens` profile; langchain_ibm then computes `max_tokens = profile - input_tokens`, which goes negative once prompts exceed 8K
- Prefer known model context sizes (131K for gpt-oss-120b) over provider profiles, set the correct profile at watsonx model creation, and resolve `model_id` for context summarization

## Root cause
On large agent sessions, context summarization triggered against an 8K window (1407% usage) while the actual model supports 131K. Even after summarization, the next watsonx call still used the 8K profile, producing negative `max_tokens`.

## Test plan
- [x] `uv run pytest tests/unit/test_token_counter.py -q` (21 passed)
- [x] `uv run pytest tests/unit/test_context_summarizer.py -q` (46 passed)
- [ ] Restart server with watsonx + `openai/gpt-oss-120b` and rerun a long CugaLite workflow that previously failed with `max_tokens ... got -5178`

### DIFF
--- a/src/cuga/backend/cuga_graph/utils/context_summarizer.py
+++ b/src/cuga/backend/cuga_graph/utils/context_summarizer.py
@@ -86,10 +86,7 @@ class ContextSummarizer:
             self.middleware = None
 
     def _setup_model_profile(self):
-        """Setup model profile with context size for fraction-based triggers."""
-        if not (hasattr(self.config, 'trigger_fraction') and self.config.trigger_fraction):
-            return
-
+        """Setup model profile with correct context window for the resolved model."""
         resolved_name = resolve_model_identifier(self.model, fallback_name=self.model_name)
         try:
             context_size = ensure_model_context_profile(self.model, resolved_name)
--- a/src/cuga/backend/cuga_graph/utils/context_summarizer.py
+++ b/src/cuga/backend/cuga_graph/utils/context_summarizer.py
@@ -14,7 +14,11 @@ from langchain.agents.middleware import SummarizationMiddleware
 from loguru import logger
 
 from cuga.config import settings
-from cuga.backend.cuga_graph.utils.token_counter import TokenCounter
+from cuga.backend.cuga_graph.utils.token_counter import (
+    TokenCounter,
+    ensure_model_context_profile,
+    resolve_model_identifier,
+)
 from cuga.backend.cuga_graph.utils.message_utils import convert_to_proper_message_type
 
 # Optional imports for middleware invocation (may not be available in all LangChain versions)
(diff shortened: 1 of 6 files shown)


src/cuga/backend/cuga_graph/utils/context_summarizer.py
```diff
@@ -86,19 +90,10 @@ def _setup_model_profile(self):
         if not (hasattr(self.config, 'trigger_fraction') and self.config.trigger_fraction):
             return
```
offerakrabi: this return will not go through `ensure_model_context_profile`, meaning in this edge case the fix will not work

src/cuga/backend/cuga_graph/utils/token_counter.py
```diff
@@ -289,6 +289,72 @@
...
+    for key in sorted_keys:
+        if normalized_name.startswith(key):
+            return MODEL_CONTEXT_SIZES[key]
+
+    return None
+
+
+def ensure_model_context_profile(model: Optional[Any] = None, model_name: Optional[str] = None) -> int:
+    """Ensure *model.profile* reflects the known context window for the resolved model name."""
+    resolved_name = resolve_model_identifier(model, fallback_name=model_name or "")
```
offerakrabi: i think this is called twice, once here and once again after this function returns in `_setup_model_profile`

it doens't break anything, just redundant

src/cuga/backend/cuga_graph/utils/token_counter.py
```diff
@@ -289,6 +289,72 @@
...
+                model.profile = {"max_input_tokens": context_size}
+                logger.debug(
+                    "Set model profile: max_input_tokens={} for {}",
+                    context_size,
+                    resolved_name or model_name or "unknown",
+                )
+        except Exception as exc:
+            logger.warning(f"Failed to set model.profile: {exc}")
+
+    return context_size
```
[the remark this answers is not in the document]
sami-marreed: Merged in de546d23: `ensure_model_context_profile` now copies existing profile keys and only updates `max_input_tokens`.

src/cuga/backend/cuga_graph/utils/context_summarizer.py
sami-marreed: Removed the `trigger_fraction` early return so profile setup always runs in `_setup_model_profile` (de546d23). Creation-time fix in `models.py` already covered the main watsonx invoke path; this closes the summarizer-only gap.

src/cuga/backend/cuga_graph/utils/token_counter.py
sami-marreed: Fixed in de546d23: `ensure_model_context_profile` uses the passed `model_name` directly when provided instead of re-resolving from the model instance.

offerakrabi (approved):

PR #443, opened 2026-07-07
feat(policy): reuse pre-generated ToolGuard code across environments via policy export/import

### LINKED ISSUE #418: [Feature] ToolGuard — import pre-generated tool guards from a file
## What you want and why

Every time policies are imported or the environment is reset, tool guards must be re-generated from scratch via LLM. Each generation takes ~1 minute per tool and consumes LLM credits. If a user has already generated and reviewed guards in one environment (e.g. dev), they should be able to export and import them into another (e.g. staging/prod) without re-running generation.

This also enables guards to be committed to source control and shared across a team as stable, reviewed artefacts.

## How it could work

**Export:**
- An **Export guards** button (or inclusion in the existing policy export JSON) that serialises the `tool_guards` field for all Tool Guide policies into a portable JSON file.
- Format: the existing `policy.tool_guards` structure — `{ tool_name: { violating_examples, compliance_examples, policy_code } }` — keyed by policy ID.

**Import:**
- An **Import guards** button in the Policies UI that accepts a guards JSON file and merges the `tool_guards` into the matching policies (matched by policy ID).
- Uses the existing `merge_tool_guards` helper in `tool_guard_policy_updates.py` to preserve any locally generated guards not present in the import.
- Saves via the normal config-store sync path — no new endpoint required.

**Validation:**
- Warn if a policy ID in the import file does not exist in the current config.
- Do not overwrite guards for tools not present in the import file (non-destructive merge).

## Links / extra context

- Parent epic: #75
- `merge_tool_guards` helper: `src/cuga/backend/cuga_graph/policy/tool_guard/tool_guard_policy_updates.py`
- Existing policy import: `importPolicies` in `PoliciesConfig.tsx`
- Related: batch generation issue (alternative to re-generating after import)

offerakrabi:
## Feature

Closes #418

### Summary

Saving and loading policies with ToolGuard code and examples was not possible — guards were silently dropped on export and lost on server restart. In this PR we enable exporting and importing policies with ToolGuard code and examples, both through the UI and the SDK, without triggering any LLM regeneration.

**Root cause:**
- `_policy_to_frontend_dict` omitted `guards_enabled` from the `tool_guide` branch, so it was dropped from every exported JSON
- `_policy_to_markdown` did not write `guards_enabled` to the `.md` YAML frontmatter
- `create_tool_guide_from_markdown` did not read `tool_guards` or `guards_enabled` back from frontmatter, so both were lost on server restart

**Changes:**
- `server/main.py` — add `guards_enabled` to the `tool_guide` branch of `_policy_to_frontend_dict` (defaults to `True` for legacy exports)
- `filesystem_sync.py` — write `guards_enabled` into YAML frontmatter when saving a `ToolGuide` to a `.md` file
- `folder_loader.py` — read `tool_guards` and `guards_enabled` back from frontmatter in `create_tool_guide_from_markdown`
- `tests/unit/test_policy_frontend_dict_guards_enabled.py` — unit tests for the frontend-dict fix
- `tests/unit/test_toolguard_reuse_roundtrip.py` — regression tests covering all three transport paths: frontend export→import, SDK `export_policies_to_json`→import, and markdown filesystem round-trip
- `src/scripts/run_tests.sh` — add new test files to the CI test suite

### Testing
- [x] Tested locally; all 9 unit tests pass
- [x] Verified manually through the UI — exported a policy with pre-generated guards, imported it, confirmed guards were present and no regeneration was triggered
- [x] Verified via an SDK e2e smoke script (not committed) — 17 checks across 5 scenarios: load from JSON, retrieve via `agent.policies.get()`, SDK export, re-import into a fresh agent, and `guards_enabled=False` preservation — all pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Tool guard settings are now preserved consistently across markdown, JSON, and frontend policy views, including correct `guards_enabled` defaulting/normalization (e.g., when missing or `null`) and safe handling of invalid `tool_guards` entries.
  * Markdown frontmatter extraction is more resilient, preventing delimiter collisions from affecting policy content.
* **New Features**
  * Tool guide markdown now supports `tool_guards` and exports the `guards_enabled` flag.
* **Tests**
  * Added/expanded unit tests covering frontend flag behavior and end-to-end ToolGuard/`guards_enabled` round-trips via multiple import/export paths.


sami-marreed:
## Code Review

**Overall: Good fix, but one bug should be fixed before merging.**

### What's good
- The main fix works. It correctly saves and loads the `guards_enabled` setting in all three places: the frontend export, the SDK export, and the markdown file. Old files without this setting still default to "enabled," so nothing breaks for existing users.
- The tests are solid. They don't just check that data comes back the same — one test actually checks that importing a policy does NOT trigger the AI to regenerate the guard code, which is the main point of this PR.
- I ran all 5 new tests and they pass.
- The extra, unmentioned change in `utils.py` (handling bad guard entries one-by-one instead of failing the whole batch) turned out to be a real, needed fix, not random extra code. It just wasn't described in the PR, so it's worth a quick note there.

### Bug to fix
**Markdown files can break if the guard code contains "---".**
The code that reads the markdown file splits the file on the first two `---` it finds, anywhere in the text. If the saved guard code has something like a comment `# --- section ---`, that gets mistaken for the end of the frontmatter, and the whole file fails to load. I tested this and reproduced the error.

This matters because it's the exact bug this PR is trying to fix (guards getting silently lost) — just triggered a different way. Since LLM-generated code often has comment dividers like this, it's likely to happen in practice.

Suggested fix: split the frontmatter only on lines that are exactly `---` (e.g. a regex anchored to the start of a line), not any `---` substring anywhere in the file. Also add a test using guard code that contains `---`.

### Smaller things worth doing
- No test checks what happens with a broken/invalid guard entry (the PR claims bad entries get skipped, but nothing proves it).
- `guards_enabled = False` is only tested for the markdown path, not for the frontend or SDK JSON export/import. Worth testing there too, since a bug could silently flip it back to `True`.
- The `guards_enabled` value handling changed slightly in a way that isn't mentioned in the PR description — worth a one-line explanation of the new behavior for values that aren't a plain `true`/`false`.
- A couple of very minor style nits (using the safer `yaml.safe_dump`, and a bit of unreachable defensive code) — not blocking.

### Summary
Solid fix for the core problem, and well tested overall. Please fix the `---` splitting bug (or explicitly flag it as a known limitation) before merging, and add a couple of the missing test cases above.

offerakrabi:
@sami-marreed 
Addressed in 829244e3:

1. Markdown frontmatter delimiter parsing — Updated [`parse_markdown_with_frontmatter()`](src/cuga/backend/cuga_graph/policy/folder_loader.py:28) to treat only lines exactly equal to `---` as YAML frontmatter delimiters. This prevents ToolGuard guard code comments/strings containing `---` from breaking markdown reload.

2. YAML dumping — Switched policy markdown frontmatter serialization in [`filesystem_sync.py`](src/cuga/backend/cuga_graph/policy/filesystem_sync.py:171) from unsafe YAML dumping to safe YAML dumping.

3. Guard code with `---` regression — Added coverage in [`test_toolguard_reuse_roundtrip.py`](tests/unit/test_toolguard_reuse_roundtrip.py) proving markdown round-trip preserves ToolGuard policy code containing `# --- section ---`.

4. Invalid guard entries — Added coverage in [`test_toolguard_reuse_roundtrip.py`](tests/unit/test_toolguard_reuse_roundtrip.py) proving invalid per-tool guard entries are skipped while valid guards on the same policy are preserved.

5. `guards_enabled=False` JSON round trips — Strengthened frontend export/import and SDK export/import coverage in [`test_toolguard_reuse_roundtrip.py`](tests/unit/test_toolguard_reuse_roundtrip.py) to verify `guards_enabled=False` remains false.

6. `guards_enabled=null` behavior — Added coverage in [`test_toolguard_reuse_roundtrip.py`](tests/unit/test_toolguard_reuse_roundtrip.py) confirming JSON null defaults to enabled, matching legacy/missing-field behavior.

7. `merge_tool_guards` coverage — No code change. JSON/markdown policy import currently treats imported policies as snapshots via [`apply_policies_data_to_storage()`](src/cuga/backend/cuga_graph/policy/utils.py:114), not as partial ToolGuard patch updates via [`merge_tool_guards()`](src/cuga/backend/cuga_graph/policy/tool_guard/tool_guard_policy_updates.py:11). Adding merge semantics would be a behavior change, so I left it out of this PR.

Ruff format + Ruff check --fix pass locally. Focused tests pass locally: 8 passed for [`test_toolguard_reuse_roundtrip.py`](tests/unit/test_toolguard_reuse_roundtrip.py) and [`test_policy_frontend_dict_guards_enabled.py`](tests/unit/test_policy_frontend_dict_guards_enabled.py).

sami-marreed (approved):

PR #450, opened 2026-07-08
fix(frontend): finalize stream and clear loading state on Stop

### LINKED ISSUE #449: [Bug]: Stop button freezes web UI and blocks new chat requests
## What you want and why

Clicking the **Stop** button in the CUGA web UI chat should cancel the in-flight agent request and return the chat to an idle state so the user can send a new message immediately.

Instead, after Stop is clicked the entire page appears frozen and no further requests can be sent. This makes the chat unusable whenever a long-running task is cancelled mid-stream.

## How it could work

1. User sends a message and the agent begins streaming a response.
2. User clicks **Stop**.
3. The frontend should:
   - Abort the active `/stream` fetch via the Carbon `AbortSignal`
   - Call the backend `/stop` endpoint for the thread
   - Clear Carbon's loading indicator (`updateIsMessageLoadingCounter`)
   - Finalize any in-progress streaming shell message with a `complete_item` / `final_response` so Carbon exits streaming mode
4. The composer re-enables and the user can submit a new request.

**Suspected root cause:** `customSendMessage.ts` defers shell message creation until the first SSE event and uses a loading counter during the pre-stream window (#265). On cancellation, the stream is aborted but the shell message is never finalized when `signal.aborted` is detected mid-loop, leaving Carbon stuck in an active streaming state. The loading counter may also remain incremented if cancellation happens before the first event.

**Relevant files:**
- `src/frontend_workspaces/frontend/src/carbon-chat/customSendMessage.ts` — stream abort / finalize logic
- `src/frontend_workspaces/frontend/src/carbon-chat/CarbonChat.tsx` — `STOP_STREAMING` bus event handler

## Links or extra context

- Repro: open chat (or Manage page "Try your configuration" panel), send a message that takes noticeable time, click **Stop** while streaming, then try to send another message.
- Related prior fix: `fix(ui): show Carbon loading indicator during pre-stream window` (#265) introduced deferred shell creation + loading counter that must be cleaned up on cancel.

sami-marreed:
## Bug fix

Fixes #449

### Summary

Clicking **Stop** during an in-flight chat request left the Carbon AI Chat component stuck in streaming mode. The deferred shell message (introduced in #265) was never finalized on cancel, and the pre-stream loading counter could remain incremented — freezing the UI and blocking new messages.

This change:
- Adds `finalizeStream` / `handleCancellation` helpers in `customSendMessage.ts`
- Finalizes the in-progress shell message with a cancellation response when `AbortSignal` fires
- Always clears the loading indicator in `finally`, including pre-stream cancellations
- Routes `AbortError` through the same cancellation path instead of adding a dangling message

### Type of Changes

### Root Cause

After #265 deferred shell message creation until the first SSE event, cancellation mid-stream exited the read loop without sending `complete_item` / `final_response`. Carbon remained in an active streaming state and disabled the composer.

### Solution

On cancellation (signal aborted or `AbortError`), clear the loading counter and either finalize the existing shell stream or post a standalone cancellation message if streaming never started.

### Testing
- [x] Confirmed cancellation before first SSE event clears loading state
- [x] Confirmed cancellation mid-stream finalizes shell message and re-enables composer

### Checklist
- [x] Code follows project conventions
- [x] Linked to issue #449

## Summary by CodeRabbit

* **New Features**
  * Added a new option to stop an active chat agent request.

* **Bug Fixes**
  * Improved handling of cancelled requests so the chat UI clears correctly and shows a consistent cancellation state.
  * Prevented loading indicators from lingering after a request ends or is interrupted.
  * Made stream completion more reliable when messages finish normally or are aborted.

Test evidence

image


nir-mashkif:
I confirm that the UI behaves correctly, and code is valid

nir-mashkif (approved):
I confirm code is fine , system behave right

PR #458, opened 2026-07-09
feat(policy): add context_variables field to ToolGuide for Jinja2 rendering

hantablack9:
## Feature

Closes #

### Summary

Adds an optional `context_variables: Dict[str, Any]` field to `ToolGuide`. When non-empty, `guide_content` can be rendered as a Jinja2 template with those variables at enactment time. Empty dict (the default) preserves existing behavior — fully backwards compatible.

This allows tool descriptions to include live session values (e.g. result counts, current stage) without hardcoding them into the policy definition.

### Testing
- [x] Three new unit tests added in `tests/unit/test_tool_guide_context_variables.py`
- [x] Ruff format and lint checks pass

## Summary by CodeRabbit

* **New Features**
  * Added support for passing contextual variables into tool guidance content, enabling template-style rendering at enactment time when variables are provided.
  * Updated the guidance field to document and clarify this templating behavior.

* **Tests**
  * Added unit coverage for contextual variables, including defaulting to an empty set, storing provided values correctly, and ensuring existing guidance text is preserved unchanged.

### DIFF
--- a/src/cuga/backend/cuga_graph/policy/models.py
+++ b/src/cuga/backend/cuga_graph/policy/models.py
@@ -239,7 +239,19 @@ class ToolGuide(BaseModel):
             "tools use the app name 'runtime_tools', so scoped policies for those tools must include it."
         ),
     )
-    guide_content: str = Field(..., description="Markdown content to append to tool descriptions")
+    guide_content: str = Field(
+        ...,
+        description="Markdown content to append to tool descriptions. Supports Jinja2 templates when context_variables is non-empty.",
+    )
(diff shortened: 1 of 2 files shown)


sami-marreed:
Thanks for the contribution — closing this as not mergeable.

`ToolGuide.context_variables` is added on the model, but nothing in policy enactment renders `guide_content` as Jinja2, so the field would be unused. The accompanying tests only assert model storage, not end-to-end rendering.

If you want to pursue templated tool guides, please open an issue first with the intended enactment path and we can discuss the design.

PR #467, opened 2026-07-09
feat(sdk): Run Receipt — per-run token/timing visibility (fixes silent token-tracking crash)

### LINKED ISSUE #466: [Feature] Run Receipt: per-run token/cost/timing visibility; fix silent token-tracking crash
### What you want and why

Today a developer running CUGA (especially cuga-lite / API mode) cannot answer two basic questions after a run: **"what did this run cost?"** and **"where did the time go?"** — without standing up a self-hosted Langfuse stack.

Concretely:

1. **Token tracking silently crashes on the default providers.** `TokenUsageTracker.on_llm_end` (`src/cuga/backend/cuga_graph/utils/agent_loop.py:50`) dereferences `response.llm_output` without a None-guard. For LiteLLM/watsonx/streaming responses, `llm_output` is `None` (or lacks `token_usage`), the callback raises internally, and the token count is silently lost. It also never falls back to `AIMessage.usage_metadata`, the modern LangChain usage field that all our installed providers populate.
2. **No per-run token breakdown.** Usage is only accumulated into a single process-global counter (`ActivityTracker.token_usage`), so concurrent runs cross-contaminate and per-run attribution is impossible.
3. **No cost computation anywhere in the repo.** There is no price table and no `cost` calculation in the live serving path — only the optional offline Langfuse evaluation flow.
4. **No LLM-vs-tool wall-time split per run**, even though per-tool `duration_ms` is already recorded by `ToolCallTracker`.

### How it could work

Add an opt-in **Run Receipt** behind a new `advanced_features.run_receipt` flag (default **off** — zero behavior change when disabled). When enabled, `CugaAgent.invoke()` returns `result.receipt`, a structured object with a pretty `str()` rendering:

```
┌─ Run Receipt ────────────────────────────┐
│ model: gpt-4.1                           │
│ tokens: 18,342 in / 2,101 out (20,443)   │
│ est. cost: $0.0535                       │
│ steps: 6   llm calls: 7   tool calls: 4  │
│ time: 9.4s (llm 6.1s · tools 2.8s)       │
│ slowest tool: crm.search_accounts 1.9s   │
└──────────────────────────────────────────┘
```

Implementation sketch (minimal diff, reuses existing plumbing):

- A per-invoke LangChain callback handler collects `usage_metadata` (fallback: `llm_output["token_usage"]`) and LLM wall time for every model call in the run — no changes to graph nodes or state schemas, no cross-run contamination.
- Tool timings reuse the existing `duration_ms` from tool-call tracking.
- A small static price map (`$ / 1M tokens` per model pattern) estimates cost; unknown models degrade gracefully to `est. cost: n/a`.
- Every receipt code path is fail-safe (guarded), so a receipt failure can never affect a run.
- Independently of the flag: fix the `llm_output` None-guard bug in `TokenUsageTracker` with a `usage_metadata` fallback.

Out of scope (follow-ups): emitting the receipt as a final SSE event on the server `/stream` path, and receipt support for `agent.stream()`.

### Links or extra context

- Bug site: `src/cuga/backend/cuga_graph/utils/agent_loop.py:50`
- Existing tool timing: `src/cuga/backend/cuga_graph/nodes/cuga_lite/tracking/tracker.py` (`record_call(duration_ms=...)`)
- Global-only counter: `src/cuga/backend/activity_tracker/tracker.py` (`collect_tokens_usage`)

segevshlomovIBM:
## Summary

Adds a **Run Receipt** — per-run token usage and LLM-vs-tools time split — behind a new `advanced_features.run_receipt` flag (default **off**, zero behavior change when disabled), and fixes a silent token-tracking crash on the default providers.

Closes #466

### The problem

- `TokenUsageTracker.on_llm_end` dereferenced `response.llm_output` without a None-guard, so token usage silently crashed and was lost on LiteLLM/watsonx/streaming responses.
- Token usage was only accumulated into a single process-global counter — no per-run breakdown, concurrent runs cross-contaminate.
- No per-run token accounting existed in the live serving path; answering "how many tokens did this run use?" required deploying Langfuse.
- No per-run LLM-vs-tool wall-time split, despite tool `duration_ms` already being recorded.

### What this PR does

When `run_receipt = true`, every `CugaAgent.invoke()` returns `result.receipt`:

```
┌─ Run Receipt ──────────────────────┐
│ model: openai/gpt-oss-120b         │
│ tokens: 7,450 in / 208 out (7,658) │
│ llm calls: 2   tool calls: 1       │
│ time: 6.1s (llm 3.1s / tools 0.4s) │
│ slowest tool: get_accounts 0.4s    │
└────────────────────────────────────┘
```

(real output from a watsonx `gpt-oss-120b` run)

- **Per-invoke callback collector** (`RunMetricsCollector`) — one instance per `invoke()`, so concurrent runs never share counters. Reads `AIMessage.usage_metadata` with a legacy `llm_output["token_usage"]` fallback, and captures every LLM call in the run (main loop, reflection, summarizer).
- **Tokens, not cost.** Per team review the receipt deliberately reports counts only: CUGA runs against self-hosted and internal deployments whose price we don't know, so any figure would be wrong for some users. Multiply `input_tokens` / `output_tokens` by your own rates. `cache_read_tokens` and `reasoning_tokens` are surfaced when the provider reports them, since they're usually billed differently.
- **Tool timings** reuse the existing `duration_ms` from tool-call tracking (same semantics as `track_tool_calls`); the returned `result.tool_calls` stays gated on the caller's flag — unchanged behavior.
- **Fail-safe by design**: every receipt code path is guarded; any failure yields `receipt=None`, never affects the run.
- **Bug fix (flag-independent)**: None-guard + `usage_metadata` fallback in `TokenUsageTracker` — kept as a separate first commit so it can be split out if desired.

### Verification

- 15 new tests: guard fix (4), receipt/collector (9), SDK integration incl. flag-off zero-change (2).
- Full CI-equivalent gate (`run_tests.sh --skip-stability`) green locally.
- Manual E2E, offline (`CUGA_MOCK_LLM=true`) and real (watsonx `gpt-oss-120b` — the exact provider whose token tracking previously crashed): flag on → receipt above; flag off → `receipt is None`, behavior identical to `main`.
- Security review of the diff: no findings.

### Out of scope / follow-ups

- Emit the receipt as a final SSE event on the server `/stream` path.
- Receipt support for `agent.stream()` (e.g. `agent.last_receipt`).
- Optional hardening: a `timings_only` tracker mode so receipt-forced tracking records only `{name, duration_ms}` instead of full tool arguments/results in thread state.

## Summary by CodeRabbit

* **New Features**
  * Added optional “Run Receipt” on agent invocation, including token counts, model name, and per-run LLM vs tool timing.
  * Added “timings-only” tool-call tracking to capture timing metadata without storing tool payloads.
* **Bug Fixes**
  * Hardened token-usage extraction for providers that return missing or malformed usage fields.
* **Documentation**
  * Documented the `run_receipt` flag and receipt fields/behavior.
* **Tests**
  * Added unit and end-to-end integration coverage for receipts, token tracking, and timings-only mode.

---

**Update (team review):** cost estimation has been removed — `cost_usd` and `backend/llm/pricing.py` are gone, and the receipt reports input/output token counts only. This also retires both open review threads about pricing sources (litellm-vs-static table, RITS list prices): there is no price lookup left to get wrong.


src/cuga/backend/cuga_graph/utils/agent_loop.py
```diff
@@ -47,7 +47,17 @@ def __init__(self, tracker: ActivityTracker):
     async def on_llm_end(self, response: LLMResult, **kwargs):
         generation = response.generations[0][0].text
         self.tracker.collect_prompt(role="assistant", value=generation)
-        self.tracker.collect_tokens_usage(response.llm_output.get("token_usage").get("total_tokens"))
+        # llm_output is None (or lacks token_usage) for LiteLLM/watsonx/streaming
+        # responses; fall back to the message's usage_metadata before giving up.
+        token_usage = (response.llm_output or {}).get("token_usage") or {}
+        total_tokens = token_usage.get("total_tokens")
+        if total_tokens is None:
+            message = getattr(response.generations[0][0], "message", None)
```
haroldship: generations[0][0] can crash if malformed response

src/cuga/backend/cuga_graph/utils/run_receipt.py
```diff
@@ -0,0 +1,180 @@
...
+    tool_time_s: float = 0.0
+    wall_time_s: float = 0.0
+    slowest_tool: Optional[str] = None
+    tool_timings: List[ToolTiming] = Field(default_factory=list)
+
+    def __str__(self) -> str:
+        cost = f"${self.cost_usd:.4f}" if self.cost_usd is not None else "n/a (unknown model)"
+        lines = [
+            f"model: {', '.join(self.models) if self.models else 'unknown'}",
+            f"tokens: {self.input_tokens:,} in / {self.output_tokens:,} out ({self.total_tokens:,})",
```
haroldship: could there be cached or reasoning tokens?

src/cuga/backend/llm/pricing.py
```diff
@@ -0,0 +1,56 @@
...
+(lowercased, provider prefixes such as ``openai/`` or ``rits/openai/`` stripped),
+longest pattern first, so put the more specific variant (``gpt-4o-mini``) rather
+than relying on ordering.
+"""
+
+from typing import Optional, Tuple
+
+# (substring pattern, USD per 1M input tokens, USD per 1M output tokens)
+MODEL_PRICES: list[Tuple[str, float, float]] = [
+    ("gpt-4.1-mini", 0.40, 1.60),
```
haroldship: Since litellm is already a core dependency, could `litellm.get_model_info()` / `litellm.model_cost` be the primary price source, with this table only as a fallback for models litellm doesn't know? It's maintained upstream and includes cache-read pricing. Two caveats from trying it:

1. Bare open-weight names aren't keys — `gpt-oss-120b` needs a provider prefix (`groq/openai/gpt-oss-120b`, `azure_ai/gpt-oss-120b`), so the lookup should try the raw (unstripped) model name before the normalized one.
2. This table already disagrees with litellm on both gpt-oss models (120b: 0.15/0.75 here vs 0.15/0.60 on groq/azure/bedrock in litellm; 20b: 0.10/0.50 vs 0.07/0.30). These look like Groq's launch prices, since reduced — which is the drift this hard-coded approach invites. (minimax-m3 and all the gpt-4.x/gpt-4o entries match litellm exactly — only the two gpt-oss models drifted.)

Note: this is the bundled/GitHub price map in the litellm package, not our LiteLLM proxy — no network dependency beyond an optional GitHub fetch that already falls back to a local copy (`LITELLM_LOCAL_MODEL_COST_MAP=True` forces offline). If we ever pull live prices from our proxy's `/model/info` instead, that one is behind VPN and would need fail-safe + caching.

src/cuga/backend/llm/pricing.py
```diff
@@ -0,0 +1,56 @@
+"""Static price estimates for the models shipped in configurations/models/*.toml.
+
+Prices are public list prices in USD per 1M tokens and are estimates by design —
```
haroldship: rits/openai/gpt-oss-120b normalizes to gpt-oss-120b and gets the public list price, but the docstring says self-hosted RITS deployments "resolve to None." If pricing RITS usage at notional market rate is intended, the docstring should say so; if not, RITS prefixes should short-circuit to None.

src/cuga/sdk.py
```diff
@@ -2297,12 +2345,17 @@ async def invoke(
...
 
+            # No pre-run snapshot exists on the resume path, so the receipt's
+            # tool timings are thread-cumulative here.
             return InvokeResult(
                 answer=final_answer,
                 tool_calls=tool_calls,
                 thread_id=thread_id,
                 error=error_msg,
                 variables=_hitl_variables,
+                receipt=self._build_run_receipt(
```
haroldship, line 2596: On the resume path the collector is fresh (tokens/llm_time cover only this resume segment) but tool timings cover the whole thread history, so tool_time_s can exceed wall_time_s and tool_call_count includes calls from before the interrupt and from earlier turns. The code comment acknowledges it, but the receipt gives its consumer no signal that its fields have different scopes. This is fixable the same way as the normal path: read len(state.values.get("tool_calls", [])) via self.graph.get_state(run_config) (already called later at line 2331) before ainvoke and slice.

src/cuga/sdk.py
```diff
@@ -2225,6 +2256,21 @@ async def invoke(
...
+        # can never affect the run itself.
+        receipt_collector: Optional[RunMetricsCollector] = None
+        receipt_started_at: Optional[float] = None
+        try:
+            if settings.advanced_features.run_receipt:
+                receipt_collector = RunMetricsCollector()
+                receipt_started_at = time.monotonic()
+                # Tool durations for the receipt require tracking; the returned
+                # result.tool_calls stays gated on the caller's track_tool_calls.
+                run_config["configurable"]["track_tool_calls"] = True
```
haroldship: Forcing track_tool_calls=True doesn't just collect timings. ToolCallTracker.record_call stores the full sanitized arguments and result of every tool call (tracking/tracker.py:98-107), and sandbox_node.py:241 appends those records to state.tool_calls, which is persisted by the checkpointer and accumulates across turns on a thread (that's why this PR needs the snapshot-slice at line 2455). So a metrics flag turns on full payload capture the caller never opted into: unbounded state growth on long-lived threads, and retention of potentially sensitive tool outputs. Consider a lightweight tracking mode (name + duration_ms only) when tracking is receipt-forced, or at minimum document this — the README says "zero overhead when disabled" but not what enabling costs.

haroldship (changes requested):
Thanks — this is a well-built feature: the fail-safe discipline is consistent, the per-run collector correctly avoids the global ActivityTracker's concurrency problem, and test coverage of the aggregation logic is solid. The on_llm_end crash fix alone is worth landing.

Requesting changes for two behavioral issues, plus a few accuracy/design questions in the inline comments:

**Blocking**

1. Enabling run_receipt silently turns on full tool-payload capture (sdk.py). Forcing track_tool_calls=True means every tool call's arguments and results are appended to state.tool_calls and persisted by the checkpointer, accumulating unboundedly across turns on a thread — data retention and state growth the caller never opted into. A metrics flag shouldn't capture payloads; a lightweight mode (name + duration only) or at minimum prominent documentation is needed.
2. Resume-path receipt mixes scopes (sdk.py). Tokens/LLM time cover only the resume segment while tool timings are thread-cumulative, so tool_time_s can exceed wall_time_s and counts include earlier turns. The consumer has no way to tell. The normal path's snapshot-slice approach can be applied here too via graph.get_state() before resuming.

**Non-blocking, see inline comments**

- Pricing: consider litellm.model_cost (already a core dependency, maintained upstream) as the primary source with the static table as fallback — the table already disagrees with litellm on both gpt-oss models.
- The docstring says self-hosted RITS models "resolve to None", but rits/openai/gpt-oss-120b normalizes to a priced entry — which behavior is intended?
- Cached/reasoning token details (input_token_details / output_token_details) are ignored, so cost overestimates on cache-heavy runs.
- response.generations[0][0] in TokenUsageTracker.on_llm_end can still IndexError on an empty generations list — same silent-loss class this PR fixes; a guard would make the fix complete.

haroldship (dismissed):
Wow - looks great

segevshlomovIBM:
All of the review points are fixed, verified end-to-end, and pushed.

I added a timings_only tracking mode so that when the receipt forces tool tracking, only tool name/app/duration reach the checkpointed state, never arguments, results, or errors, and the resume path now snapshots the prior tool_calls count via graph.get_state() so the receipt covers only the resume segment (both blockers). Pricing switched to litellm's bundled model_cost map as the primary source (static table as fallback, gpt-oss prices corrected, rits/... short-circuits to None), with cache-read tokens now collected and billed at the cache-read rate. I also guarded the empty-generations crash in both usage collectors, updated docs/tests (95 passing, re-verified E2E on watsonx), and pushed the 3 commits to the PR

haroldship:
@sami-marreed I've approved this PR

src/cuga/backend/cuga_graph/utils/agent_loop.py
segevshlomovIBM: Fixed: generations is now guarded (`first = generations[0][0] if generations and generations[0] else None`) with the usage fallback reading from the message only when present — same silent-loss class as the llm_output guard below it. The receipt collector applies the same pattern.

src/cuga/backend/cuga_graph/utils/run_receipt.py
segevshlomovIBM: Yes to both, now counted: cache-read tokens (input_token_details.cache_read) are reported per model, priced at the model's cache-read rate when litellm knows it, and shown as "N% cached"; reasoning tokens (output_token_details.reasoning) are reported per model and shown as "N reasoning" in the tokens line. Covered by test_collector_accumulates_cache_read_tokens and test_collector_accumulates_reasoning_tokens.

src/cuga/backend/llm/pricing.py
segevshlomovIBM: Done as suggested: litellm's bundled cost map is now the primary source (raw model name tried first, then lowercased, then normalized — covering the provider-prefixed-keys caveat), with the static table as fallback only. The table's gpt-oss entries were synced to litellm's current values (120b 0.15/0.60, 20b 0.07/0.30), and cache_read_input_token_cost is used for cached-input pricing when available. All lookups stay fail-safe (any litellm error → static fallback → None).

src/cuga/backend/llm/pricing.py
segevshlomovIBM: RITS is internal-IBM self-hosted only, so notional market pricing would be misleading — `rits/` prefixes now short-circuit to None before any lookup, and the docstring says so explicitly ("never billed at a notional market rate").

src/cuga/sdk.py
segevshlomovIBM: Fixed exactly as you suggested: the resume path now snapshots `len(state.values.get("tool_calls", []))` via `self.graph.get_state(run_config)` before resuming and slices the result, so the receipt's tool timings/count cover only this resume segment — matching the token/llm-time scope of the fresh collector. The normal path does the same via the initial state's tool_calls count.

src/cuga/sdk.py
segevshlomovIBM: Implemented the lightweight mode you proposed: when the receipt needs timings but the caller didn't pass track_tool_calls=True, the SDK sets the configurable to "timings_only" and ToolCallTracker records only name/app/duration — arguments, results, and error payloads are stored as None (tracking/tracker.py, timings_only contextvar). result.tool_calls stays empty unless the caller explicitly opted in, and what accumulates on thread state in this mode is the redacted records only (needed for the per-run slice), so no sensitive payload retention the caller never asked for.

src/cuga/backend/llm/pricing.py
segevshlomovIBM: Superseded by the team review: the receipt no longer estimates cost at all. `cost_usd` and `backend/llm/pricing.py` are removed — CUGA runs against self-hosted and internal deployments whose price we don't know, so any figure would be wrong for some users. The receipt now reports input/output token counts and callers apply their own rates.

That resolves this thread by removing the code it was about: there's no price lookup left to pick a source for. Thanks for pushing on it — the litellm suggestion was right for as long as we were pricing, and `cache_read_tokens` / `reasoning_tokens` from your other comment are kept, since those are counts rather than prices.

src/cuga/backend/llm/pricing.py
segevshlomovIBM: Superseded by the team review: the receipt no longer estimates cost at all. `cost_usd` and `backend/llm/pricing.py` are removed — CUGA runs against self-hosted and internal deployments whose price we don't know, so any figure would be wrong for some users. The receipt now reports input/output token counts and callers apply their own rates.

That resolves this thread by removing the code it was about: there's no price lookup left to pick a source for. Thanks for pushing on it — the litellm suggestion was right for as long as we were pricing, and `cache_read_tokens` / `reasoning_tokens` from your other comment are kept, since those are counts rather than prices.

sami-marreed (approved):

PR #472, opened 2026-07-12
feat(cuga-lite): native function calling (opt-in)

### LINKED ISSUE #471: [Bug]: cuga-lite silently mishandles native tool_calls: parallel calls dropped, preamble+tool_calls ignored, bind_tools NotImplementedError crashes the run
### What happened

When `cuga_lite_bind_tools_mode` binds tools to the model (which the **gpt-oss-20b runtime profile does automatically** — `model_runtime_profile.py:9-14`), a model that responds with native `AIMessage.tool_calls` — the encoding function-calling-trained models naturally emit — is silently mishandled by cuga-lite's execution side. Three distinct defects:

- **D1 — parallel tool calls silently dropped.** The recovery transpiler reads **`tool_calls[0]` only** (`src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/response_utils.py`, `tool_call = tool_calls[0]`) and converts it to `result = await tool(args)` for the sandbox. Every sibling call in the same turn is discarded with no error, no log, no trace — and the model then reports success for work that never happened.
- **D2 — text preamble + tool_calls → tool_calls ignored entirely.** `normalize_response` only attempts tool-call recovery when `response.content` is empty (`adapter/graph_adapter.py`, `if not content:`). A natural preamble like *"I'll notify both customers now."* means **zero tools execute** and the announcement itself becomes the final answer.
- **D9 — models without `bind_tools` support crash the run.** `langchain_core.BaseChatModel.bind_tools` raises `NotImplementedError` by default, and `NotImplementedError` is a subclass of `RuntimeError` — so the deliberate `except RuntimeError: raise` for cap errors in `helpers/bind_tools.py` re-raises it instead of falling back to the unbound model. `call_model` dies instead of degrading.

Observed demo output (scripted deterministic FC model, task "notify customers acme and globex", one `notify(customer)` tool, `cuga_lite_bind_tools_mode="all"`):

```
=== SCENARIO A: parallel tool_calls (empty content) ===
agent's final answer : 'Done — both acme and globex have been notified.'
tools ACTUALLY run   : ['acme']
VERDICT: FAIL — globex silently dropped, answer claims FALSE SUCCESS        (D1)

=== SCENARIO B: text preamble + tool_calls ===
agent's final answer : "I'll notify both customers now."
tools ACTUALLY run   : []
VERDICT: FAIL — ZERO tools ran; the announcement became the final answer   (D2)

=== SCENARIO C: baseline code-act (no binding) ===
tools ACTUALLY run   : ['acme', 'globex']
VERDICT: PASS — code-act handles it
```

Scenario A is the worst failure class an agent can exhibit: **silent partial execution reported as full success**. Scenario C shows the failures are specific to the native-FC encoding, not the task.

### Motivation / why this matters now

- The gpt-oss-20b profile enables binding with zero configuration, so every deployment on that model exercises these paths today.
- The in-tree pad-to-cap measurements ("0 tool calls vs 5-7 without padding", `bind_tools/cap.py`) prove real bound models do emit native tool_calls — despite the system prompt forbidding it (`prompts/mcp_prompt.jinja2:138`), which is its own inconsistency.
- The transpiler bridging native FC to execution has **no test coverage** (nothing imports `response_utils`; the empty-content branch of `normalize_response` is never exercised).
- Full analysis, defect table (D1–D9), and a phased fix design: `docs/reports/function_calling_cuga_lite.md` on branch `function_calling`.

### How to reproduce

Reproducing tests are on branch **`function_calling`** as strict-xfail tests (executable documentation — they assert the *correct* behavior, xfail today, and flip to hard failures once fixed):

```bash
git checkout function_calling
uv run pytest src/cuga/backend/cuga_graph/nodes/cuga_lite/tests/test_native_tool_calls_execution.py -rxX -q
# => 1 passed, 3 xfailed
#    XFAIL test_parallel_tool_calls_all_execute            (D1)
#    XFAIL test_text_preamble_with_tool_calls_still_executes (D2)
#    XFAIL test_model_without_bind_tools_falls_back_gracefully (D9)
#    PASSED test_code_act_control_executes_both            (control)
```

The tests use a deterministic scripted `BaseChatModel` (no API keys, no network) with `CugaAgent(tools=[notify])` and `config={"configurable": {"cuga_lite_bind_tools_mode": "all"}}`.

### Environment

- Branch: `main` (`df173678`), reproduced on branch `function_calling`
- Python 3.12, macOS (darwin), local executor (no Docker/e2b required)
- No model API needed — deterministic scripted model

### Expected behavior (fix sketch)

1. All `tool_calls` in a turn execute (batch block; explicit per-call success/error reporting so partial execution is impossible to miss), not just `[0]`.
2. `text + tool_calls` responses keep the text as reasoning and still execute the calls.
3. `NotImplementedError` from `bind_tools` falls back to the unbound model instead of crashing (catch it before the cap's `RuntimeError` re-raise, or use a dedicated cap exception type).
4. Generated batch code must not splice tool names into source (`await {name}(...)`) — names may not be valid Python identifiers; call through a lookup map.

See `docs/reports/function_calling_cuga_lite.md` §5 for the full design (P0 correctness → P1 typed SDK surface → P2 modes/prompt).

segevshlomovIBM:
# feat(cuga-lite): native function calling (opt-in)

Closes #471

## Why

CUGA-lite is a code-act agent: the model calls tools by writing Python run in a sandbox. It also had a `bind_tools` subsystem, but native tool calls were mishandled on execution — only the first of several was run (the rest silently dropped while the agent reported success), a text preamble suppressed the calls entirely, and a model whose `bind_tools` raised `NotImplementedError` crashed the run. This PR turns native function calling into a correct, opt-in capability without changing the default code-act path.

## Scope (what this PR does)

- **Native FC execution.** When opted in, every native tool call in a turn is executed (not just the first), a text preamble no longer suppresses the calls (it's preserved as reasoning), and calls run through the same guarded, tracked, variable-aware sandbox pipeline as code. Args are emitted via dict-splat (`await tool(**{...})`) so keyword / non-identifier parameter names (Gmail `from`, OData `$filter`, `page-size`) can't produce a SyntaxError.
- **Dedicated native prompt.** A sibling template `mcp_prompt_native.jinja2`, structurally parallel to the code-act prompt (same role, apps, tools, knowledge/skills/todos, pagination and data-source rules — including the skill install-ordering) but with all sandbox/code/Python language removed and tools reframed as native functions. It's selected only when FC is on; the code-act prompt is untouched (byte-identical to `main`).
- **SDK surface — `ToolCalling`.** `CugaAgent(tools=[...], tool_calling=ToolCalling(mode="native"))` and a per-invoke override `invoke(..., tool_calling=...)`. Options: `mode` (`code`|`native`|`hybrid`), `native_tools`/`apps` selection, `include_find_tools`, `max_bound_tools`, and `tool_choice` (`auto`|`required`|provider-specific — nudges/forces native calls for code-preferring models, with capability-detection fallback). Also readable as a global setting `advanced_features.cuga_lite_tool_invocation_mode`.
- **Safe fallback.** A model that doesn't support `bind_tools` degrades to the unbound (code-act) model instead of crashing `call_model`; providers that reject `tool_choice` fall back to a plain bind. Every new path is `try/except`-guarded and degrades to code-act.
- **Full `mode="code"` opt-out.** An explicit `ToolCalling(mode="code")` forces the code-act prompt/handling **and unbinds tools** (`cuga_lite_bind_tools_mode="none"`), overriding a global `native` default and any model profile that would otherwise bind (e.g. gpt-oss-20b) — so an explicit code opt-out is truly code-act.

## Default is unchanged (verified)

Off by default at every layer (`settings.toml`, config validator, and `ToolCalling` default all `"code"`). With no opt-in: the code-act prompt is byte-identical to `main` (0-line diff), the legacy tool-call recovery is byte-identical (single-call, empty-content-only), few-shots inject as before, and no tools are bound. Concurrency-safe: the invocation mode is derived **per call** from `configurable` (not stored on the shared adapter). Dedicated "flag-off" tests lock this in.

## Not in this PR (deferred, additive/non-breaking)

- **Direct dispatch** (a ToolNode-style executor that invokes tools without the sandbox) — keeps the sandbox as the universal tool bus for now.
- **Parallel fan-out** via `asyncio.gather` (multi-call turns run sequentially) — needs a sandbox-injected helper since `import asyncio` is restricted.
- **Args-schema pre-validation** against the bound `StructuredTool.args_schema`.
- **Run-receipt `invocation` breakdown** (native vs code counts).
- **`on_unsupported="error"`** (currently only the `fallback_code` degradation ladder is implemented).

## Testing

- Focused FC suite green (transpiler incl. keyword/non-identifier args & legacy parity; `normalize_response` on/off × preamble/empty/code-wins/concurrency; `_safe_bind` + `tool_choice` fallback; `ToolCalling`→configurable + mode resolver + code opt-out unbinds the gpt-oss-20b profile; native-prompt framing, install-ordering, and template-selection wiring; few-shot skip). `ruff check` clean.
- Full `run_tests.sh unit_tests` (CI unit job) green.
- Live e2e against watsonx `gpt-oss-120b`: with the native prompt the model emits native tool calls that CUGA executes (both tools run, correct answer). Also fixes a pre-existing `errors.py` body-form `tool_use_failed` detection bug surfaced during review.

## Notes for reviewers

- The `docs/reports/security_secret_protection_poc.{md,html}` untracked files are from a separate task and are **not** part of this PR.
- A design/analysis report is included at `docs/reports/function_calling_cuga_lite.md`.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in native function calling to the CugaAgent SDK via `ToolCalling` (`code`/`native`/`hybrid`), including native tool selection and `tool_choice`.
  * Added a runnable native function-calling example.
* **Bug Fixes**
  * Improved native tool-call execution to support multiple tool calls, preserve accompanying text as reasoning, and respect explicit code blocks.
  * Enhanced robustness and fallback when tool binding features aren’t supported.
* **Documentation**
  * Expanded “Native Function Calling” guidance with per-invocation overrides and configuration options.
* **Tests**
  * Added unit/integration coverage for mode resolution, transpilation, prompt rendering, safety rules, and non-leaking concurrency behavior.


segevshlomovIBM:
@coderabbitai thanks for the thorough review — all 7 comments are addressed in `79ac30fb`.

**Major**
- **Fail closed on serialization failure** — `tool_calling_to_configurable` now returns the code/unbind config (`{cuga_lite_tool_invocation_mode: "code", cuga_lite_bind_tools_mode: "none"}`) on any error instead of `{}`, so a serialization failure can never leave native binding enabled via a global setting or model profile. Added a regression test.
- **`include_find_tools=False` serialized explicitly** — the boolean is now always pinned (True or False) for an explicit native/hybrid selection, so a global setting / profile that enables `find_tools` can't widen it. Added a regression test.

**Minor / stability**
- **Non-dict `function` / non-string `name` guard** — `_parse_one_tool_call` now coerces a truthy non-dict `function` to `{}` and requires `name` to be a non-empty `str`; malformed provider entries return `None` rather than raising up through `normalize_response`/`call_model`. Added a regression test covering a non-dict `function`, a non-string `name`, and a non-dict entry (both `multi=True` and the legacy `multi=False` path).
- **RUF012** — the scripted-model test fields now use `Field(default_factory=list)` in both test files.
- **`ToolCalling` default docstring** — corrected: `None` → `{}` (no-op); an explicit `ToolCalling(mode="code")` emits `"code"` + `"none"`.
- **Global-setting opt-in documented** — the report now describes both opt-in paths (SDK `ToolCalling` and `advanced_features.cuga_lite_tool_invocation_mode`) resolved by `resolve_tool_invocation_mode` (`configurable > global setting > "code"`).
- **Nested Markdown fence** — fixed with a four-backtick outer fence.

Verification: focused FC suite + `ruff check` (incl. RUF012) + `run_tests.sh unit_tests` all green.

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/graph_adapter.py
```diff
@@ -146,16 +147,35 @@ async def resolve_bind_tools(
             logger.warning(f"AgentGraphAdapter.resolve_bind_tools failed: {exc}")
         return None
 
-    def normalize_response(self, response: Any) -> Tuple[str, Optional[str]]:
+    def normalize_response(
+        self, response: Any, configurable: Optional[dict] = None
```
haroldship, line 151: `configurable` parameter is never explained. In `graph_nodes.py` it just says "not used here". should explain in the docstrings what different values are allowed and what they do

src/cuga/backend/cuga_graph/nodes/cuga_lite/helpers/bind_tools.py
```diff
@@ -18,6 +18,39 @@
     apply_bind_tools_cap_and_merge,
     bind_tools_max_count_from_settings,
 )
+from cuga.backend.cuga_graph.nodes.cuga_lite.tool_calling import native_tool_calls_enabled
+
+
+def _safe_bind(model: BaseChatModel, tools: List[StructuredTool], tool_choice: Any = None) -> BaseChatModel:
```
haroldship, line 24: this guard only covers half the failure. it fires when bind_tools() itself raises — but hosted providers (OpenAI, watsonx, ...) don't validate tool schemas at bind time, binding is just local config. they validate when the chat request is actually sent, so a tool with a loose/invalid schema (common with MCP tools — missing param types, odd keywords) comes back as a 400 from inside `ainvoke_model,` where only the Groq tool_use_failed shape is caught — everything else re-raises and kills the run. so a tool set that works fine in code-act (schemas never leave the prompt) can crash in mode="native", which breaks the "every new path degrades to code-act" guarantee.

suggestion: when FC is opted in, catch provider bad-request errors at invoke time, log that the tool schemas were rejected, and retry once without the bound tools. note ainvoke_model only gets the bound model, so either the unbound model needs to be passed in, or the catch/retry belongs in call_model where both are in scope. (alternatively note it as a known gap in the report's deferred list — but I think it belongs in this PR since it's the ladder's biggest hole.)

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/graph_adapter.py
```diff
@@ -146,16 +147,35 @@ async def resolve_bind_tools(
...
-            if tool_code:
-                logger.warning("Empty content with tool_calls detected; recovering tool call as Python code")
-                content = tool_code
         reasoning = normalize_assistant_text(
             (getattr(response, "additional_kwargs", None) or {}).get("reasoning_content")
         )
+        # Derived per call from configurable/settings (no adapter state: the
+        # adapter is shared across concurrent invokes with possibly different
+        # tool_calling modes).
+        allow_native = native_tool_calls_enabled(configurable)
```
haroldship, line 160: Related gap just above in `ainvoke_model`: this method only recovers the Groq `tool_use_failed` error. Under native/hybrid, a provider rejecting the bound tool schemas (400 at invoke time) re-raises and crashes the run instead of degrading to code-act — see my comment on _safe_bind.

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/response_utils.py
```diff
@@ -34,30 +35,87 @@ def tool_call_kwarg_literal(value: Any) -> str:
...
+        function = {}
+    name = tool_call.get("name") or function.get("name")
+    if not isinstance(name, str) or not name:
+        return None
+    args = tool_call.get("args") or function.get("arguments") or {}
     if isinstance(args, str):
         try:
             args = json.loads(args)
         except json.JSONDecodeError:
             args = {}
```
haroldship: if args is a string that fails json parsing, the tool still gets called — just with no args. same for the line below: a non-dict args becomes {}. for a tool with required params that's just a TypeError the model can retry, but for a tool where all params are optional it will run with its defaults instead of the intended args, and the agent reports success. that's the same silent-wrong-execution problem this PR is fixing (D1).

I don't think returning None works either — _parse_tool_calls silently drops None entries, so the call would silently not run at all. better: on the multi/native path, skip the call and emit a visible marker like _tool_call_statements already does for non-identifier names, e.g. print("[skipped tool call with unparseable args: notify]"), so the model sees it and retries. (multi=True only — the legacy path has to stay byte-identical to main.)

src/cuga/backend/cuga_graph/nodes/cuga_lite/helpers/bind_tools.py
```diff
@@ -192,7 +225,27 @@ async def resolve_model_with_bind_tools(
         settings_tool_names_fn=_bind_tools_tool_names_from_settings,
         settings_include_fn=lambda: _bind_include_find_tools_from_config({}),
     )
-    max_count = bind_tools_max_count_from_settings()
+    # Couple native function calling to binding (issue #471): if FC is enabled
+    # (e.g. via the global cuga_lite_tool_invocation_mode setting) but no bind
+    # mode was chosen, bind all tools — otherwise the native prompt tells the
+    # model to call tools that were never bound. The SDK ToolCalling path already
+    # sets a concrete bind mode, so this only fills the global-setting gap.
+    if mode in ("", "none", "false", "0", "off") and native_tool_calls_enabled(cfg):
```
haroldship, line 233: defaulting to bind "all" puts the cap/shortlister machinery on the main SDK path, and I don't think it's ready for that. binding is resolved inside call_model, so with more than `cuga_lite_bind_tools_max_count` (default 128) (aside, I added that because of m3) tools every turn re-collects all tools and makes an extra LLM round-trip to shortlist them — hidden latency and cost, per turn. and the shortlister was deliberately built strict for benchmarks: hallucinated names or a missing query raise RuntimeError, which `resolve_bind_tools` deliberately re-raises — so the run crashes instead of degrading to code-act, which breaks the "every new path degrades to code-act" guarantee. this isn't a corner case for us: in m3/vakra there are domains with several hundred tools, so ToolCalling(mode="native") — the exact one-liner the README advertises — hits the shortlister on every turn from the start.

suggestion: when the mode came from ToolCalling (i.e., it's not a benchmark config), catch the cap RuntimeError and degrade to the unbound code-act model with a log warning. and/or reconsider the "all" default (same default in tool_calling_to_configurable, tool_calling.py:114) or at least warn on ToolCalling that large tool sets mean a per-turn shortlister call and a possible hard failure, so callers know to pass native_tools=[...]/apps=[...] or max_bound_tools=0.

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/graph_adapter.py
```diff
@@ -146,16 +147,35 @@ async def resolve_bind_tools(
...
+            # as reasoning, not lost (issue #471 D2).
+            tool_code = extract_code_from_response_tool_calls(response, multi=True)
+            if tool_code:
+                if content:
+                    reasoning = f"{reasoning}\n{content}".strip() if reasoning else content
+                    logger.warning("Tool calls emitted alongside a text preamble; executing the tool calls")
+                content = tool_code
+        elif not content:
+            # Legacy path (FC off, or FC on with a code block): unchanged behavior.
+            tool_code = extract_code_from_response_tool_calls(response, multi=allow_native)
```
haroldship: nit: the elif is only reachable when allow_native is False (empty content can't contain a fence), so this is always multi=False — write that literally.

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/graph_adapter.py
```diff
@@ -146,16 +147,35 @@ async def resolve_bind_tools(
...
+        # adapter is shared across concurrent invokes with possibly different
+        # tool_calling modes).
+        allow_native = native_tool_calls_enabled(configurable)
+        if allow_native and "```python" not in content:
+            # Native FC opted in: honor tool_calls even alongside a text preamble,
+            # unless the model already emitted a PYTHON code block (explicit
+            # code-act intent wins). A non-python fence (```json / ```text) in the
+            # preamble must NOT suppress the tool calls. The preamble is preserved
+            # as reasoning, not lost (issue #471 D2).
+            tool_code = extract_code_from_response_tool_calls(response, multi=True)
```
haroldship, line 167: when "```python" in content → the elif is skipped for non-empty content, tool_calls are ignored, and nothing is logged. If a model emits both a fence and tool_calls, that's a D1-shaped silent drop by design; add a logger.warning so it's diagnosable.

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/prepare_node.py
```diff
@@ -598,6 +610,17 @@ async def _wrapped(*args, **kwargs):
                 t for t in (tools_for_prompt or []) if getattr(t, "name", None)
             ]
 
+        # Native function calling (issue #471): when enabled (resolved above),
```
haroldship, line 617: non-blocking: hybrid and native select the same template here, and mcp_prompt_native.jinja2 has all code/Python language removed — it even says output must be tool calls or plain text. so the model in hybrid mode is never told it may write code, which makes hybrid indistinguishable from native at the prompt level. the docs say hybrid means "both encodings are offered; the model picks per step". either give the native template a {% if hybrid %} section that offers the python block as an option, or re-document hybrid as "native prompt, code tolerated if the model emits it". both are fine with me, but docs and behavior should match.

haroldship (commented):
really solid work overall — the gating is careful, the default-unchanged claim holds up, and the test coverage on the new paths is thorough. requesting changes on three things, all inline:

1. schema rejection at invoke time isn't handled — providers validate tool schemas when the request is sent, not at bind, so a loose MCP schema 400s inside ainvoke_model and kills the run instead of degrading to code-act (comment on _safe_bind).

2. unparseable tool-call args become {} and the tool still executes — for all-optional-param tools that means running with defaults and reporting success (comment on response_utils.py).

3. native mode defaults to binding all tools, which routes big tool sets into the per-turn shortlister and its deliberate hard-raise. in m3/vakra some domains have several hundred tools, so this hits us immediately (comment on bind_tools.py).

all three break the "every new path degrades to code-act" promise in the PR description, and all three look like small fixes.

the rest are non-blocking: hybrid renders the pure-native prompt so it's not really hybrid (decide + document), a couple of small nits inline.

segevshlomovIBM:
**Split per team process:** the two standalone bug fixes in this PR (the `bind_tools` `NotImplementedError` crash → graceful fallback, and the body-form `tool_use_failed` detection in `llm/errors`) have been extracted to #524 so they can be reviewed and land on their own merit.

This PR remains the **proposal** for the opt-in native function-calling capability (dedicated native prompt, `ToolCalling` SDK surface, multi-call execution, `tool_choice`) — flagged off by default — pending design review with the team and an evaluation run (AppWorld + digital-sales, code-act vs native vs hybrid) before any recommend-on decision. Once #524 lands, this branch will be updated on `main` and the duplicated fixes will fall out of the diff.

segevshlomovIBM:
Parking this as a draft — it needs a couple more design discussions before it's ready, per today's meeting. Not abandoned; converting to draft so it stops reading as "awaiting review" while those happen.

State for whoever picks it up:

**Green** on CI, but **35+ commits behind main** with conflicts in `sdk.py` and `__init__.py` (main's lazy-import refactor landed since).

**Three blocking items from @haroldship's review, still open:**
1. Provider **400s at invoke time** aren't handled — providers validate tool schemas when the request is sent, not at bind, so a loose MCP schema kills the run instead of degrading to code-act.
2. `mode="all"` routes large tool sets into the shortlister's deliberate hard-raise (several hundred tools in m3/vakra), so the advertised `ToolCalling(mode="native")` one-liner hard-fails.
3. **Hybrid mode renders the pure-native prompt**, so hybrid is indistinguishable from native at the prompt level — needs either a real `{% if hybrid %}` section or a docs correction.

Plus 8 inline threads from 12 July that still need replies.

Note: #524 merged since, so `BindToolsUnsupportedError` now exists upstream and makes item 2 materially cheaper than when it was first raised.

segevshlomovIBM:
Superseded by #777, which implements native function calling as a switchable execution mode with a real tool-exec loop instead of routing tool_calls through the sandbox. Closing; the branch stays for reference.

PR #490, opened 2026-07-14
fix: route workspace tree to host when sandbox_mode is native/local

### LINKED ISSUE #484: [Bug]: /api/workspace/tree logs opensandbox import error and returns 503 in native sandbox mode
## What happened

While running the server locally, the UI (or any client) polling workspace files triggers noisy logs and a failed API response:

```
2026-07-14 10:39:44.853 | WARNING  | cuga.backend.server.main:get_workspace_tree:3505 - Sandbox workspace tree failed: No module named 'opensandbox'
INFO:     127.0.0.1:58071 - "GET /api/workspace/tree?thread_id=512b7a24-24dd-4863-9014-c094b7300624 HTTP/1.1" 503 Service Unavailable
```

The agent session itself keeps working — this is not blocking execution — but the workspace tree panel cannot load and the terminal spams warnings on each poll.

## How to reproduce

1. Use default-ish local settings where `advanced_features.opensandbox_sandbox = true` but `advanced_features.sandbox_mode = "native"` (current `src/cuga/settings.toml` ships this combination).
2. Start the server **without** installing the optional OpenSandbox extra (`uv sync` without `--extra opensandbox`, so `opensandbox` is not importable).
3. Open a chat thread and let the frontend call `GET /api/workspace/tree?thread_id=<thread_id>`.

**Expected:** Workspace tree is served from the native/host thread workspace (same as code execution in `native` mode).

**Actual:** `get_workspace_tree` takes the OpenSandbox branch because `workspace_tree_is_sandbox_backed()` only checks `opensandbox_sandbox`, tries `CodeExecutor._get_opensandbox_executor()`, import fails, host fallback returns empty, and the endpoint responds **503**.

Relevant code:

- `workspace_tree_is_sandbox_backed()` — `src/cuga/backend/server/workspace_sandbox.py:22` (true when `opensandbox_sandbox` is set, ignores `sandbox_mode`)
- `get_workspace_tree` fallback — `src/cuga/backend/server/main.py:3499-3508`

## Environment

- OS: macOS (darwin 25.3.0)
- CUGA: local dev on `main`
- `sandbox_mode = "native"`, `opensandbox_sandbox = true`
- OpenSandbox SDK not installed (`No module named 'opensandbox'`)

```
WARNING  | cuga.backend.server.main:get_workspace_tree:3505 - Sandbox workspace tree failed: No module named 'opensandbox'
GET /api/workspace/tree?thread_id=... HTTP/1.1" 503 Service Unavailable
```

## Suggested fix (optional)

Align workspace API routing with execution policy: when `sandbox_mode` is `native` or `local`, use `fetch_native_workspace_tree` / host paths instead of requiring the OpenSandbox SDK. Alternatively, treat missing `opensandbox` as a soft fallback with a single debug log instead of warning + 503 on every poll.

sami-marreed:
## Bug fix

Fixes #484

### Summary
`/api/workspace/tree` was taking the OpenSandbox path whenever `opensandbox_sandbox=true`, even with default `sandbox_mode=native` (and shell tools off). That caused noisy import errors and 503s when the optional OpenSandbox SDK was not installed.

Workspace routing now follows `sandbox_mode`: `native`/`local` use the host thread workspace; OpenSandbox is only required when mode is actually opensandbox-backed.

### Testing
- [x] `uv run pytest tests/unit/test_workspace_sandbox.py` — 31 passed
- [x] `uv run pytest tests/unit/test_workspace_upload.py` — 16 passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Workspace behavior now correctly respects the selected sandbox mode.
  * Native and local modes consistently use the native workspace, even when shell tools are disabled or sandbox settings are enabled.
  * Other modes continue to select sandbox-backed workspaces when configured.

### DIFF
--- a/src/cuga/backend/server/workspace_sandbox.py
+++ b/src/cuga/backend/server/workspace_sandbox.py
@@ -20,7 +20,16 @@ _LEGACY_DISPLAY_ROOTS = {"tmp", "cuga_workspace"}  # kept for backward-compat pa
 
 
 def workspace_tree_is_sandbox_backed() -> bool:
-    return bool(getattr(settings.advanced_features, "opensandbox_sandbox", False))
+    """True when workspace APIs should use the OpenSandbox SDK.
+
+    ``opensandbox_sandbox`` alone is not enough: when ``sandbox_mode`` is
+    ``native`` or ``local``, files live on the host even if the OpenSandbox
+    flag is set for other features.
+    """
+    if not bool(getattr(settings.advanced_features, "opensandbox_sandbox", False)):
+        return False
+    mode = str(getattr(settings.advanced_features, "sandbox_mode", "opensandbox") or "opensandbox")
+    return mode not in ("native", "local")
 
 
 def workspace_tree_is_native_backed() -> bool:
@@ -30,12 +39,12 @@ def workspace_tree_is_native_backed() -> bool:
     plan = ExecutionRouter.resolve(settings)
     if plan.filesystem_backend == "host" or plan.shell_backend in ("native", "local"):
         return True
+    mode = str(getattr(settings.advanced_features, "sandbox_mode", "opensandbox") or "opensandbox")
+    if mode in ("native", "local"):
+        return True
     if workspace_tree_is_sandbox_backed():
         return False
-    mode = getattr(settings.advanced_features, "sandbox_mode", "opensandbox")
-    return bool(
-        getattr(settings.advanced_features, "enable_shell_tool", False) and mode in ("native", "local")
-    )
+    return bool(getattr(settings.advanced_features, "enable_shell_tool", False))
 
 
 def _hidden_parts(parts: tuple[str, ...]) -> bool:
--- a/tests/unit/test_workspace_sandbox.py
+++ b/tests/unit/test_workspace_sandbox.py
@@ -13,56 +13,99 @@ from cuga.backend.server.workspace_sandbox import (
 )
 
 
+def _settings(
+    *,
+    opensandbox_sandbox: bool,
+    sandbox_mode: str,
+    enable_shell_tool: bool = False,
+    enable_filesystem_tools: bool = False,
+):
+    class Adv:
+        pass
+
+    adv = Adv()
+    adv.opensandbox_sandbox = opensandbox_sandbox
+    adv.sandbox_mode = sandbox_mode
+    adv.enable_shell_tool = enable_shell_tool
+    adv.enable_filesystem_tools = enable_filesystem_tools
+
+    class Settings:
+        advanced_features = adv
+        execution = type(
+            "Exec",
+            (),
+            {
+                "shell_backend": None,
+                "filesystem_backend": None,
+                "python_backend": None,
+                "workspace_root": None,
+            },
+        )()
+
+    return Settings()
+
+
+@pytest.mark.unit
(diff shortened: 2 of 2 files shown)


offerakrabi:
# PR Review: #490 — fix: route workspace tree to host when sandbox_mode is native/local

## Summary

**Verdict: merge with fixes.** The core fix is correct and minimal: gating `workspace_tree_is_sandbox_backed()` on both `opensandbox_sandbox` and `sandbox_mode` directly addresses the reported 503s. The refactoring of the test helper into `_settings()` removes eight lines of repeated class boilerplate and makes all four new tests visibly more readable. `test_workspace_tree_uses_native_when_sandbox_mode_native_without_shell_tools` directly reproduces the exact config from the issue report.

**Strengths:**
- `workspace_tree_is_sandbox_backed()` fix is minimal: two lines, correct semantics, and the function now has a docstring explaining the intent.
- All four new `@pytest.mark.unit` tests assert both predicates (`is_native_backed` and `is_sandbox_backed`) in tandem, verifying they stay mutually consistent rather than testing in isolation.
- `test_workspace_tree_uses_native_when_sandbox_mode_native_without_shell_tools` directly exercises the reported bug config (`opensandbox_sandbox=True`, `sandbox_mode=native`, shell tools off), giving an explicit regression guard.
- 47 tests pass with no regressions.

**What to resolve before merge:**

Minor:
1. MIN-1 — `workspace_tree_is_native_backed()` now routes any `sandbox_mode=native/local` config to `fetch_native_workspace_tree`, which silently skips the `ensure_thread_workspace_seeded()` call that `fetch_host_workspace_tree` makes. For deployments using workspace fixture seeding (`SEED_MODES` env var), the first tree poll on a new thread will return an empty workspace instead of pre-populated fixtures.


## Issues


**[MIN-1] `workspace_tree_is_native_backed()` now routes `sandbox_mode=native/local` to `fetch_native_workspace_tree`, which skips `ensure_thread_workspace_seeded()` that `fetch_host_workspace_tree` calls**
`src/cuga/backend/server/workspace_sandbox.py:42-44`

**What's wrong:** The new guard at lines 42–44 (`if mode in ("native", "local"): return True`) fires before the `workspace_tree_is_sandbox_backed()` check, and it fires for *any* config with `sandbox_mode=native/local` — including `opensandbox_sandbox=False` with no shell tools, which previously fell through to `main.py:3511 → fetch_host_workspace_tree()`. `fetch_host_workspace_tree` (in `workspace_upload.py:280`) calls `ensure_thread_workspace_seeded(tid)` before scanning; `fetch_native_workspace_tree` (in `workspace_sandbox.py:176`) does not. These two functions scan the same physical directory (`<cwd>/cuga_workspace/<thread_id>/`), so the outputs are identical in the common case. But the seeding call is silently dropped.

**What it can cause:** When `SEED_MODES` is configured (a CRM or CI fixture-injection env var), the first workspace-tree API call for a new thread normally seeds fixture files into the thread directory. After this PR, any deployment with `sandbox_mode=native/local` (regardless of `opensandbox_sandbox`) skips that seeding entirely. The workspace tree panel will appear empty on first load instead of showing the pre-seeded fixtures. The root cause is invisible from the API response — it just returns an empty tree with no error.


---

offerakrabi (approved):

Issue #444, opened 2026-07-07
[Epic] Knowledge, production-grade: citations, reindex reliability & ingest performance


dolev31:
## What & why

Knowledge **v1** (#128) made document knowledge work end-to-end — ingest, retrieval, agent/session scoping, and the SDK + UI surfaces. This epic (**v2**) hardens that foundation into a **trustworthy, production-grade** retrieval system: answers you can trace to their sources, reindexing that never corrupts or confuses state, ingest that scales, and correct tokenization across every embedding provider.

**Predecessor:** #128 — Knowledge v1: integration & foundations (delivered).

### 1. Citations & provenance
Trace every answer back to the documents and passages it used.
- ⬜ #154 — Citations & sources data model for knowledge retrieval *(design)*
- ⬜ #155 — Inline citation markers + hover previews + sources dropdown (UI) *(design)*

### 2. Reindex reliability & error-state UX
Reindexing must be race-free, correctly scoped, and legible when it fails.
- ✅ #396 — reindex concurrency race: hold `_agent_draft_lock` across migrate+reindex
- ✅ #397 — draft-status state leak across modal close / agent switch / publish
- ✅ #398 — `ReindexBusyError` maps to its own FE code (`reindex_busy`), not `reindex_failed`
- ✅ #402 — reindex tile shows filenames (not `task_xxx`); rows persist on terminal state
- ⬜ #403 — Documents tab count diverges from reindex worker count after failed embeds

### 3. Next-gen ingest performance
Take ingest from "fast enough" to "scales".
- ⬜ #319 — content-hash cache for re-ingest (skip embed on unchanged content)
- ⬜ #320 — streaming ingest pipeline via Docling `page_chunk_size`
- ⬜ #321 — CUDA Graphs + warm worker for query-time embed (avoid cold-start cliff)

### 4. Tokenizer & splitter correctness
Correct chunk boundaries across every embedding provider.
- ✅ #387 — HybridChunker tokenizer mismatch for litellm/watsonx-routed HF embedders
- ⬜ #395 — user-facing tokenizer override based on cl100k-fallback telemetry *(design)*
- ⬜ #399 — extend token-aware splitter to openai/azure (tiktoken) + canary parity

### 5. Knowledge-profile UX
Make scope, cost, and configuration obvious to operators.
- ⬜ #303 — Knowledge profile UX: per-scope overrides + cost disclosure
[...]

PR #497, opened 2026-07-19
fix(knowledge): make the embedding-model cache durable and self-healing

### LINKED ISSUE #496: [Refactor] Embedding-model cache is not durable or self-healing (bricks knowledge search)
Part of #444 (Knowledge, production-grade).

## What you want and why

A user running `cuga start demo_knowledge` was met with:

> **Embedder unavailable** — Your 4 indexed documents can't be searched — the active embedder (fastembed/BAAI/bge-small-en-v1.5) isn't reachable. Open Configure knowledge base to check its API key / connection and run Test connection.

Knowledge search was dead, and the suggested remedy was impossible to follow: `fastembed` runs a local ONNX model with **no API key and no endpoint**. Whether a model is cached is an implementation detail a user should never have to know about, let alone repair by hand.

Four independent gaps combined:

1. **Prevention** — `FASTEMBED_CACHE_PATH` is set only in `Dockerfile.ubi`, so local installs fall back to fastembed's default of `tempfile.gettempdir()/fastembed_cache`. On macOS that is `/var/folders/...`, which the OS periodically purges — silently deleting hundreds of MB of ONNX models from a working install.
2. **Recovery** — a purged or interrupted download leaves a dangling snapshot symlink plus a 0-byte `.incomplete` blob. HuggingFace still believes the model is cached, so onnxruntime raises `NoSuchFile` on *every* subsequent start. The install is permanently bricked; restarting never helps, and the only fix is deleting a hash-named directory under `/var/folders` — something no user will ever do.
3. **State** — health is binary available/unavailable, so a legitimate first run downloading a 66 MB model renders as a **red error**. Working software looks broken.
4. **Diagnosis** — the backend already returns `embedder_error`, but the UI discards it and prints hardcoded API-key/connection advice regardless of provider.

## How it could work

1. Default `FASTEMBED_CACHE_PATH` to a persistent per-user location (`~/.cache/cuga/fastembed`). Not under `PACKAGE_ROOT` — that is `site-packages` for a pip install, the wrong home for large blobs and possibly read-only.
2. On a model-load failure whose signature indicates an incomplete/corrupt cache, purge that one entry and re-fetch **exactly once**. Bounded, skipped under `HF_HUB_OFFLINE`, and narrow enough that a bad key, unsupported model, or offline hub still surfaces rather than silently re-downloading. The purge target must be derived from the failing path in the error (the fastembed model name does not match its backing repo dir) and constrained to the cache dir, since it is passed to `rmtree`.
3. Add a `preparing` state (`disabled` / `preparing` / `available` / `unavailable`) so a cold start reports progress instead of failure, without blocking the polled health endpoint or triggering a duplicate download.
4. Make the banner provider-aware and surface the scrubbed `embedder_error`.

## Links or extra context

Observed traceback:

```
onnxruntime.capi.onnxruntime_pybind11_state.NoSuchFile: [ONNXRuntimeError] : 3 : NO_SUCHFILE :
Load model from /var/folders/.../T/fastembed_cache/models--qdrant--bge-small-en-v1.5-onnx-q/
snapshots/5239.../model_optimized.onnx failed. File doesn't exist
```

Cache state at the time — the snapshot symlink pointed at a blob that never finished downloading:

```
model_optimized.onnx -> ../../blobs/51f1bd0...2431      (dangling)
51f1bd0...2431.incomplete                    0 bytes
```

dolev31:
## Bug fix

Fixes #496 — part of #444 (Knowledge, production-grade).

### Summary

A user hit **"Embedder unavailable — your 4 indexed documents can't be searched"**, with the banner telling them to check an *API key and connection* for `fastembed/BAAI/bge-small-en-v1.5` — a local ONNX model that has neither. Knowledge search was dead and the suggested fix was impossible to follow.

Root cause was four gaps stacking up, fixed here in one pass:

**1. Prevention — models were stored where the OS deletes them.**
`FASTEMBED_CACHE_PATH` was set only in `Dockerfile.ubi`, so local installs fell back to `$TMPDIR`. macOS periodically purges `/var/folders`, deleting the model out from under a working install. Local runs now default to `~/.cache/cuga/fastembed`. Deliberately *not* `PACKAGE_ROOT` (where `dbs/` and `logging/` live): for a pip install that is `site-packages`, the wrong home for 66 MB blobs and possibly read-only. An explicitly-set `FASTEMBED_CACHE_PATH` still wins, so container images are unaffected.

**2. Recovery — a poisoned cache bricked the install permanently.**
A purged or interrupted download leaves a dangling snapshot symlink and a 0-byte `.incomplete` blob. HuggingFace still believes the model is cached, so onnxruntime raised `NoSuchFile` on *every* subsequent start — restarting never helped, and the only escape was deleting a hash-named dir under `/var/folders`, which no user will do. That entry is now purged and re-fetched **exactly once**.

Three deliberate constraints on the retry:
- **Bounded** — a second failure is a genuine fault (offline, disk full, upstream gone) and must surface, not loop.
- **Narrow** — only cache-corruption signatures match. A bad API key, an unsupported model name, or a gated repo still fails immediately; silently re-downloading would mask real misconfiguration.
- **Contained** — the purge target is parsed from the failing path in the error (the fastembed model name does not match its backing repo dir, so this is the only reliable link) and is constrained to the cache dir, since it is fed to `rmtree`.

Skipped entirely under `HF_HUB_OFFLINE=1`, which is an explicit promise not to hit the network.

**3. State — a cold start looked like a failure.**
Health only knew available/unavailable, so a legitimate first run downloading a 66 MB model rendered as a red error. Adds a `preparing` state (`disabled` / `preparing` / `available` / `unavailable`), reported without blocking the polled health endpoint or kicking off a duplicate download of the same model.

**4. Diagnosis — the UI threw away the real reason.**
The backend already returned `embedder_error`; `ManagePage` read only `embedder_available` and printed hardcoded key/connection advice. The banner is now provider-aware — local providers have no key or endpoint to check — and surfaces the scrubbed reason.

### Testing

- **Reproduced and healed the real failure**: poisoned a live cache exactly as it failed in the wild (deleted the blob, left the dangling symlink and 0-byte `.incomplete`), then watched it recover against real `fastembed` in 11.8s:
  ```
  WARNING | Embedding model cache for BAAI/bge-small-en-v1.5 is incomplete or corrupt
            (models--qdrant--bge-small-en-v1.5-onnx-q); purging and re-downloading once.
  INFO    | Embedding model cache for BAAI/bge-small-en-v1.5 recovered after re-download
  ✅ loaded + embedded, dim=384
  ```
- **21 new unit tests** covering: corruption-signature classification (and that auth/network/unsupported-model errors are *not* misread as corruption), cache-root extraction, the containment guard against `rmtree` on an out-of-cache path, retry-exactly-once, offline-mode opt-out, the `$TMPDIR` default, and `preparing` vs `unavailable`.
- **End-to-end** `cuga start demo_knowledge` asserted on positive signals (`Embeddings initialized`, `Knowledge prewarm complete`) rather than absence of errors.
- **No regressions**: frontend `tsc` shows 19 unique errors on this branch vs 19 on `main` (zero new); the `tests/unit` suite shows no new failures versus a stashed clean-`main` baseline — the two deltas seen were `test_langfuse_tracing` order-dependent flakes that pass in isolation and are unrelated to this change.

Bundle rebuilt (`NODE_ENV=production`) and committed, matching the convention in #465.

## Summary by CodeRabbit

* **New Features**
  * Added richer Knowledge embedder health reporting (disabled, preparing, available, unavailable) and exposed it via the health API.
  * Improved the Knowledge embedder UX with clearer notifications for cold-start versus true unavailability, including tailored guidance for local vs non-local providers.
  * Introduced a configurable, persistent cache directory for embedding models.
* **Bug Fixes**
  * Added self-healing when embedding model caches are corrupt by safely purging affected cache/lock data and retrying once.
  * Ensured partial/failed embedder initialization doesn’t leave the system in a broken state.
* **Tests**
  * Expanded unit coverage for cache resilience, offline behavior, health/probe states, and cache configuration defaults/env overrides.


dolev31:
@offerakrabi @sami-marreed — this is a **complementary follow-up stacked on #468**, ready for review alongside it. Just re-synced onto the updated #468 base (6bb2bb10) so it's `MERGEABLE`.

**What it does:** makes the embedding-model cache durable + self-healing — a purged/half-downloaded model no longer bricks knowledge search with a bare `disk I/O error`. Fixes #496. Small backend diff (engine.py + config.py) + tests.

Also applied the same test-marker convention you asked for on #468 (`pytest.mark.unit` + CI allowlist) here. Reviews requested — it merges right after #468.

dolev31:
@sami-marreed — flagging you for review here (you're CODEOWNER for `src/cuga/backend/knowledge/`). This is the PR that **fixes the knowledge model-cache issue** — the one where a purged/half-downloaded embedding model bricked knowledge search with a bare `disk I/O error` (root cause was a dangling snapshot symlink in `$TMPDIR` after macOS reaped the cache). Fixes #496.

Exactly where the fix lives:

**1. Prevention — stop storing models where the OS deletes them**
`src/cuga/config.py:47-50` — `FASTEMBED_CACHE_DIR` now defaults to `~/.cache/cuga/fastembed` instead of fastembed's `$TMPDIR` fallback (which macOS periodically purges). Containers already pinned `FASTEMBED_CACHE_PATH`; this gives local installs the same durability.

**2. Recovery — self-heal a corrupt/incomplete cache** (the core fix), `src/cuga/backend/knowledge/engine.py`:
- `_is_corrupt_model_cache_error` (`:867`) — recognizes the NoSuchFile / incomplete-blob signature (and only that — a bad key / offline hub still surfaces).
- `_model_cache_root_from_error` (`:875`) — pulls the `models--…` dir out of the failing path, **constrained to the cache dir** before it's handed to rmtree.
- `_purge_model_cache` (`:902`).
- `_FastEmbedEmbeddings.__init__` (`:909`, purge-and-retry at `:945-952`) — on a corrupt-cache load failure, purges that one entry and re-downloads **exactly once**; a second failure surfaces. Skipped under `HF_HUB_OFFLINE`.

**3. Concurrency** — `_ensure_embeddings` (`engine.py:2011-2019`, lock at `:1778`): double-checked lock so two `to_thread` callers can't both load the model (CodeRabbit Major, already addressed).

**4. Diagnosis** — `src/frontend_workspaces/frontend/src/ManagePage.tsx:1871` (`isLocalEmbeddingProvider` at `:260`): the "Embedder unavailable" banner is now provider-aware — for a local model it says "no key or connection to check — model files missing" instead of the impossible "check your API key".

Tests: `tests/unit/test_embedder_cache_resilience.py` (21 cases, marked `unit` + in CI). Reproduced the real failure and watched it self-heal.

It's stacked on #468 and `MERGEABLE`. 🙏

offerakrabi:
# PR Review: #497 — fix(knowledge): make the embedding-model cache durable and self-healing

These are two small findings, consider whether to fix or deffer

## Summary

**Verdict: merge with fixes** (one minor, one NTH). This is a carefully engineered fix for a real, reproducible user-facing failure. The four-gap decomposition in the PR body is accurate and the implementation matches it precisely. The self-heal logic in `_FastEmbedEmbeddings.__init__` is deliberately bounded, narrow in scope, and correctly skipped under `HF_HUB_OFFLINE`. The double-checked lock in `_ensure_embeddings` is correct and the `preparing` state is properly threaded through the backend and surfaced in the UI.

**Strengths:**
- The corruption-signature classifier (`_is_corrupt_model_cache_error`, `_model_cache_root_from_error`) is deliberately narrow and tested for false-positive cases — auth errors, unsupported model names, and offline mode are explicitly verified not to trigger re-download.
- The containment guard in `_model_cache_root_from_error` (path constrained to `cache_dir` via `.relative_to`) prevents `rmtree` from walking outside the cache directory even if the error text were injected.
- 21 unit tests exercise the real code paths — classification, cache-root extraction, purge, the self-heal retry sequence, offline opt-out, and the `preparing` state — none relying on mocks of the logic under test.
- The double-checked lock correctly prevents the duplicate ONNX session + download problem that CodeRabbit identified on first review, and the second commit addresses it cleanly.
- `FASTEMBED_CACHE_DIR` placement in `config.py` is correct: `~/.cache/cuga/fastembed` is durable, writable, and not `site-packages`.

**What to resolve before merge:**

*Minor:*
1. **MIN-1** — "Preparing embedder" info banner fires with zero indexed documents, confusing first-time users who haven't ingested anything yet. Add a `knowledgeDocCount > 0` guard at `ManagePage.tsx:1837` to match the sibling error banner.

*Nice to have:*
2. **NTH-1** — Pre-existing: if `create_embeddings` succeeds but `_get_embedding_dim` raises, `_default_embeddings` is set non-None while `_default_embedding_dim` stays `None`, permanently skipping re-init and later causing an `IntegrityError` at the ingest path. Not introduced by this PR, but `_ensure_embeddings` was refactored here — worth noting for a follow-up. *Needs a decision*: fix now (reset `_default_embeddings = None` in an except clause) or track as follow-up.

## Issues

### 🟡 Minor

**[MIN-1] `preparing` notification has no document-count guard**
`src/frontend_workspaces/frontend/src/ManagePage.tsx:1837`

**What's wrong:** The "Preparing embedder" info banner fires on `knowledgeEmbedderState === "preparing"` alone with no check on `knowledgeDocCount`. The sibling "Embedder unavailable" error banner at line 1862 correctly gates on `knowledgeEmbedderAvailable === false && knowledgeDocCount > 0`. The `preparing` banner has no equivalent guard, so it fires for any user who opens the Manage page during a first-run model download, even if they have never ingested a document and have no reason to care about embedding readiness.

**What it can cause:** A first-time user opens the Manage page while CUGA is downloading the ONNX model in the background. They see a blue info banner about the embedder warming up before they have done anything with knowledge. The banner is not an error, but it is unexpected and disappears without explanation when the model finishes loading — which looks like a bug. The user's natural question is "warming up what? I haven't added any documents."

---

### 🔵 Nice to Have

**[NTH-1] `_ensure_embeddings` half-init on `_get_embedding_dim` failure is unguarded at the ingest path**
`src/cuga/backend/knowledge/engine.py:2025-2028`

**What's wrong:** If `create_embeddings` succeeds (line 2026) but the immediately following `_get_embedding_dim` call raises (line 2027 — it calls `embed_query("test")`, a live network/ONNX call), `_default_embeddings` is set to a non-`None` value while `_default_embedding_dim` stays `None`. The `finally` clears `_embedder_initializing`. Any subsequent call to `_ensure_embeddings` sees `_default_embeddings is not None` at line 2020 and returns immediately, so `_default_embedding_dim` is never populated. This path is pre-existing (inherited from before this PR's lock refactor); the refactor made the permanently-stuck outcome slightly more deterministic but did not create it.

**What it can cause:** An ingest that calls `_resolve_collection` at line 2275 passes `self._default_embedding_dim` (which is `None`) to `set_collection_config` at line 2280. The metadata store schema has `embedding_dim INTEGER NOT NULL`, so the insert raises an `sqlite3.IntegrityError` — a confusing secondary failure whose root cause is the original `embed_query` error, not a DB problem. The guard at line 3782 (`if self._default_embedding_dim is not None`) shows awareness of this possibility in the config-apply path, but the ingest path at line 2280 has no equivalent guard.

---

sami-marreed:
rebase and resolve conflicts

dolev31:
@sami-marreed done ✅ — **rebased onto `main`** (which now carries #468 as the squash-merge `6a8ca48`) and resolved the conflicts. The PR diff is now just #497's own changes — no leftover #468 content. Now `MERGEABLE`.

@offerakrabi both of your findings are addressed (`5f337ce`):
- **MIN-1** — the *"Preparing embedder"* info banner is now gated on `knowledgeDocCount > 0`, matching the sibling *"Embedder unavailable"* banner, so a first-run user with no indexed documents never sees an unexplained warm-up notice.
- **NTH-1** — decided to **fix now**. `_ensure_embeddings` now resets `_default_embeddings`/`_default_embedding_dim` to `None` if `_get_embedding_dim` (the live `embed_query` probe) raises after the embedder object is built — so the lock-free fast path can't skip re-init forever and later feed `dim=None` into the `NOT NULL` column at ingest. Added a regression test that drives the half-init path and asserts a clean retry.

CodeRabbit's two first-round items are also resolved: the check-then-act race → the double-checked lock in `_ensure_embeddings` (you confirmed this), and the cfg-reload test staleness → the test now `addfinalizer`s a post-teardown `importlib.reload(cfg)`.

All 22 unit tests pass locally. Re-requesting review 🙏

offerakrabi (approved):

PR #498, opened 2026-07-19
feat(knowledge): open a cited PDF at the cited page

### LINKED ISSUE #499: [Feature] Open a cited document at the cited page
Part of #444 (Knowledge, production-grade).

## What you want and why

Clicking **Open document** on a citation always lands on **page 1**, even when the citation itself is labelled `p.7`. The user then has to scroll through the PDF hunting for the passage the agent just quoted — which undercuts the point of citations: letting someone verify a claim quickly.

The page number is already known. It is carried end-to-end today (`sources.py` → `MessageSource.page`) and rendered as `p.N` in the sources panel by `pageLabel()`. The document opener simply discards it.

## How it could work

Append the page as a URL fragment when opening the document:

```ts
const url = URL.createObjectURL(blob) + pageFragment(source);
window.open(url, "_blank", "noopener");
```

Design notes:

- **PDF-only.** For `.md` / `.txt` / `.csv` / `.docx`, `page` is a *chunk ordinal* rather than a real page, so a fragment would scroll somewhere arbitrary. Opening at the top is better than landing confidently wrong. `pageLabel()` already hides the label for those formats, so the helper should live beside it and share the same rule — otherwise the label and the jump can drift apart.
- **Fragment, not query param.** `/api/knowledge/documents/file` requires `X-Agent-ID` / `X-Thread-ID` headers that `window.open` cannot set, so the file is fetched via `fetch` and opened as a blob URL. The page therefore has to ride in as a fragment.

## Links or extra context

**Known risk to validate:** `#page=N` is reliable on `http(s)` URLs but historically inconsistent on `blob:` URLs, and this depends on the browser's built-in PDF viewer — no automated test can settle it. Needs a manual check: ask something answered by a multi-page PDF, click a citation showing `p.N`, confirm it opens at page N.

If a viewer ignores the fragment it degrades to current behaviour (opens at the top), so there is no regression either way. If it proves unreliable across browsers, the fallback is bundling PDF.js to render in-app — a much larger change (~1 MB on a vendors bundle already at 6.9 MB), only worth it if passage-level highlighting is also wanted.

**Explicitly out of scope:**

- Highlighting the cited sentence rather than just opening the page — needs a custom PDF.js viewer with text-layer search. `#:~:text=` is not a shortcut: Chromium-only, and does not work on `blob:` URLs.
- Non-PDF formats — no page concept to jump to without an in-app renderer.

dolev31:
## Feature

Fixes #499. Stacked on #468 (citations). Part of #444.

### Summary

Clicking **Open document** on a citation always landed on **page 1**. Following a citation labelled "p.7" meant scrolling to find the passage the agent had just quoted — which undercuts the point of citations.

The page number was already carried end-to-end (`sources.py` → `MessageSource.page`, rendered as `p.N` by `pageLabel`). The opener just dropped it. This appends it as a URL fragment.

### How it works

Adds `pageFragment()` beside `pageLabel()` in `citations/types.ts` — deliberately next to the existing pageless-format rule so the two can't drift:

```ts
const url = URL.createObjectURL(blob) + pageFragment(source);
window.open(url, "_blank", "noopener");
```

**PDF-only, on purpose.** For `.md` / `.txt` / `.csv` / `.docx`, `page` is a *chunk ordinal*, not a real page — a fragment would scroll somewhere arbitrary. Opening at the top beats landing confidently wrong. `pageLabel` already hides the label for those formats, so the two behaviours now agree.

**Why a fragment and not a query param:** `/api/knowledge/documents/file` requires `X-Agent-ID` / `X-Thread-ID` headers that `window.open` can't set, so the file is fetched via `fetch` and opened as a blob. The page therefore has to ride in as a fragment on the blob URL.

### ⚠️ One thing to confirm in review

`#page=N` is reliable on `http(s)` URLs but **historically inconsistent on `blob:` URLs**, and that's the one thing no automated test can settle — it depends on the browser's built-in PDF viewer.

**Please sanity-check when reviewing:** ask something answered by a multi-page PDF, click a citation showing `p.N`, confirm it opens at page N.

If a viewer ignores the fragment it **degrades to exactly today's behaviour** (opens at the top) — no regression either way. If it turns out unreliable across browsers, the fallback is bundling PDF.js to render in-app, which is a much bigger change (~1 MB on a vendors bundle already at 6.9 MB) and only worth it if we also want passage-level highlighting.

### Not in scope

- **Highlighting the cited sentence** — needs a custom PDF.js viewer with text-layer search. `#:~:text=` is not a shortcut: Chromium-only and doesn't work on `blob:` URLs.
- **Non-PDF formats** — no page concept to jump to without an in-app renderer.

### Testing

- **10 new vitest cases** in `pageFragment.test.ts`: PDF happy path, case-insensitive extension, every pageless format returns `''`, null/undefined/0/negative pages rejected (no `#page=0`), and a shape guard that any non-empty result matches `^#page=[1-9]\d*$` so `blobUrl + fragment` is always a valid URL.
- Full citations vitest suite green (17 tests: 7 existing + 10 new).
- Frontend `tsc`: 22 unique errors on this branch vs 22 on the `feat/knowledge-citations` base — zero new.
- Bundle rebuilt (`NODE_ENV=production`) and committed, per the convention in #465; verified the fragment is present in the shipped `main.*.js`.

## Summary by CodeRabbit

* **New Features**
  * Cited PDF documents now open directly to the referenced page.
  * Added support for pageless formats like `.docx` while keeping page links PDF-only.
* **Bug Fixes**
  * Prevented invalid page fragments (including missing, non-integer, or non-positive values) so links never show `#page=0` or malformed output.
  * Improved tab opening behavior and safer link handling using `noopener,noreferrer`, with more reliable cleanup.
* **Tests**
  * Added a comprehensive test suite to validate page fragment generation rules and safety.


dolev31:
@offerakrabi @sami-marreed — **complementary follow-up stacked on #468**, ready for review with it. Re-synced onto the updated #468 base (5d3595a1) → `MERGEABLE`.

**What it does:** clicking a citation opens the cited PDF **at the cited page** instead of page 1 — directly completes the citations UX in #468. Fixes #499. Tiny diff (a `pageFragment` helper + one call site) + 10 vitest cases.

One thing worth a reviewer's eye is called out in the PR body: `#page=N` on a `blob:` URL is browser-viewer-dependent — it degrades to opening at the top, so no regression either way. Reviews requested — merges right after #468.

offerakrabi:
# PR Review: #498 — feat(knowledge): open a cited PDF at the cited page

## Summary

**Verdict: merge with fixes** — the core design is correct and the implementation is clean. `pageFragment()` is a pure, well-tested function that sits exactly where it belongs, next to `pageLabel()`. The 10 vitest cases are specific and honest, including a shape guard that prevents any concatenation-site surprises. The 60-second revocation window is a reasonable heuristic for the `window.open` + blob URL detour. One real bug needs fixing before merge.

**Strengths:**
- `pageFragment` is placed in `citations/types.ts` directly beside `pageLabel`, making the "pageless format" rule physically impossible to drift — both functions will be edited together.
- The guard `typeof source.page !== 'number' || source.page < 1` correctly rejects `null`, `undefined`, `0`, and negatives, so `#page=0` can never appear in a blob URL.
- Tests use a regex shape guard (`/^#page=[1-9]\d*$/`) that protects the concatenation site from any future breakage, not just the current inputs.
- Export through `citations/index.ts` keeps the public API surface clean and consistent with the existing `pageLabel` / `scopeLabel` pattern.
- 

## Issues


### 🟠 Major

**[MAJ-1] `revokeObjectURL` is called with a fragment-appended URL and may not free the blob**
`src/frontend_workspaces/frontend/src/carbon-chat/CarbonChat.tsx:892`

**What's wrong:** `url` is constructed as `URL.createObjectURL(blob) + pageFragment(source)`, producing a string like `blob:http://localhost:3000/abc123#page=7`. `URL.revokeObjectURL` is then called with this fragment-bearing string. The WHATWG File API spec requires the exact blob URL (no fragment) to locate and release the object in the browser's URL store. If the browser does not strip the fragment before the lookup — which Firefox and Safari do not guarantee — the revocation silently fails, the blob remains live for the entire page session, and the 60-second timer is wasted. Each "Open document" click that lands on a PDF leaks one blob for the session lifetime.

**What it can cause:** A user browses a long agent conversation with many PDF citations, clicking "Open document" on each. Every click adds a permanently live blob to memory. The tab never reclaims that memory until the page is reloaded. On Firefox/Safari this is a deterministic leak; on Chromium it's luck depending on version. The fix is to split the URL: `const blobUrl = URL.createObjectURL(blob); window.open(blobUrl + pageFragment(source), ...); setTimeout(() => URL.revokeObjectURL(blobUrl), 60_000)`.

---

### 🟡 Minor

**[MIN-1] `window.open` uses `"noopener"` but the codebase convention is `"noopener,noreferrer"`**
`src/frontend_workspaces/frontend/src/carbon-chat/CarbonChat.tsx:890`

**What's wrong:** Every other `window.open` call in this codebase that opens a new tab uses `"noopener,noreferrer"` (e.g. `KnowledgeConfig.tsx:3216`, `ToolReview.tsx:81`). This call omits `noreferrer`. For blob URLs the referrer is typically empty anyway, but omitting it is inconsistent and will confuse reviewers looking for a reason later.

**What it can cause:** No security regression is likely for blob URLs specifically. The inconsistency is the main issue: the next reviewer will wonder whether the omission was intentional, and a linter rule that enforces `noreferrer` would flag this file only.

---

### 🔵 Nice to Have

**[NTH-1] `pageLabel` and `pageFragment` are inconsistent for `.docx`**
`src/frontend_workspaces/agentic_chat/src/citations/types.ts:13`

**What's wrong:** `PAGELESS_EXTENSIONS` does not include `.docx`. As a result, `pageLabel` returns `p.N` for `.docx` sources (the label renders in the sources panel), but `pageFragment` correctly returns `''` for them. The PR description and the `pageFragment` doc-comment both claim "the two behaviours now agree" — they don't for `.docx`. The gap is pre-existing in `pageLabel`, but the PR introduces `pageFragment` as the enforcer of that rule and explicitly makes the "they agree" claim.

**What it can cause:** A `.docx` source with `page: 5` shows "p.5" in the sources panel but clicking "Open document" opens at the top. A user sees a page label that implies a navigation target, but the document opens at page 1. The PR explicitly intended to close this kind of gap ("Better to open at the top than to land confidently wrong") — the inconsistency is directly contrary to that goal.

---

dolev31:
@offerakrabi all three findings addressed (`6d2ecef`), and the PR is **rebased onto `main`** (post-#468 squash-merge) → now `MERGEABLE`:
- **MAJ-1 (blob leak)** — split the blob URL from the fragment. `window.open` gets `blobUrl + pageFragment(source)`, but `revokeObjectURL` is called with the **bare** `blobUrl`, so Firefox/Safari's exact-URL lookup actually frees the blob instead of leaking one per click. Applied your exact suggested shape.
- **MIN-1** — `window.open` now uses `"noopener,noreferrer"`, matching the codebase convention.
- **NTH-1 (.docx)** — added `.docx` to `PAGELESS_EXTENSIONS` so `pageLabel` and `pageFragment` finally agree: Word pagination is a render-time artifact, so both now suppress rather than showing *"p.5"* while the doc opens at the top. Added tests that lock the agreement.

Rebuilt the served frontend bundle from the merged source and verified all three fixes are compiled in. 22 citation vitest cases pass. Re-requesting review 🙏

@sami-marreed this one's the citations-UX completion (open a cited PDF at the cited page) — ready alongside #497 whenever you have a moment.

offerakrabi (approved):

dolev31:
@coderabbitai good catch — addressed in `f1ec27e`. The old guard (`typeof !== 'number' || page < 1`) let `NaN`, `±Infinity`, and fractional pages slip through and emit `#page=NaN` / `#page=2.5` into the blob URL. Tightened to `typeof page !== 'number' || !Number.isInteger(page) || page < 1`, so only a real positive integer produces a fragment. Added parametrized tests covering `NaN`, `Infinity`, `-Infinity`, `2.5`, `0.9` (27 citation cases pass), and rebuilt the served bundle.

dolev31:
@sami-marreed thanks for merging #468 and #497 🙏 — **#498 is the last one in the citations stack** and it's ready:
- ✅ **CI fully green** (11/11 — the earlier red was the known-flaky WatsonX policy e2e test `test_healthcare_family_claims`, unrelated to this frontend-only change; it passed on re-run).
- ✅ **offerakrabi approved**, all 3 of his findings addressed (blob-leak MAJ-1, `noopener,noreferrer`, `.docx` pageless) + CodeRabbit's `pageFragment` hardening.
- Touches only `backend/knowledge/`-adjacent frontend citations code — you're CODEOWNER.

It completes the citations UX: clicking a citation opens the cited PDF **at the cited page** instead of page 1 (fixes #499). `MERGEABLE`, no conflicts. Your review is the only remaining gate 🙏

sami-marreed (changes requested):
@dolev31 Quick follow-up on the served UI bundle:

This branch is currently **2 commits behind `main`** (`#508`, `#497`). `#497` includes a `ManagePage.tsx` change, so the committed `src/cuga/frontend/dist` here isn’t built from the latest merged frontend source.

Could you:
1. Merge/rebase onto latest `main`
2. Rebuild with `src/frontend_workspaces/frontend/build.sh` (or equivalent `NODE_ENV=production` build + copy into `src/cuga/frontend/dist`)
3. Commit the updated dist

Thanks — the citation `#page=N` change itself looks good; this is just to keep the shipped bundle in sync.

dolev31:
@sami-marreed done — good catch, addressed in `1ec217c5`:

1. **Merged latest `main`** (picked up #497, #508, and the other 4 that landed) — clean, no conflicts.
2. **Rebuilt** `src/cuga/frontend/dist` via a `NODE_ENV=production` webpack build + copy into the package dir.
3. **Committed the updated dist.**

The rebuilt bundle is now built from the current merged frontend source, so it includes **#497's `ManagePage.tsx` doc-count guard** (which `main`'s committed bundle was actually still missing) alongside this PR's citation changes. Verified the compiled bundle contains the citation fixes (`pageFragment`/`#page=`, the bare-blob `revokeObjectURL`, and `noopener,noreferrer`).

`MERGEABLE`, no conflicts. Re-requesting review 🙏

PR #514, opened 2026-07-20
feat: preserve agent configs from DB on startup when storage is prod

### LINKED ISSUE #511: [Feature]: Persist and preload agent drafts/configs from DB on server startup
### What you want and why

On CUGA agent server startup, the process always overrides existing drafts instead of loading the agent configurations already stored in the database.

That means a restart can wipe or replace saved draft/config state, so operators lose previously configured agent settings and have to re-enter them. The server should treat the DB as the source of truth for existing agent configs/drafts and restore them on boot.

### How it could work

1. On startup, load existing agent configs (and drafts) from the DB before applying any default/bootstrap config writes.
2. Only create or seed drafts when none exist for that agent (or when explicitly requested), instead of always overwriting.
3. Keep env/file bootstrap as a fallback for first-time setup, not as a wipe of persisted state.
4. Make the preload path deterministic and logged so it's clear whether configs came from DB vs defaults.

### Links or extra context

Related: #322 (saved-config startup / force_env interactions). This request is specifically about not overriding existing drafts and preloading persisted agent configs from the DB on every server start.

### Important Notes

1. Consider taking instance id tenant id
2. Do it only when Postgres is enabled e.g. storage it prod ( maybe configure it here to work on which storage prod, local, any ) and keep default prod

### Checklist

sami-marreed:
## Feature

Closes #511

### Summary

On manager/demo bootstrap, CUGA always reseeds drafts and published configs, which overwrites persisted Postgres rows under `storage.mode=prod`. This change treats the DB as source of truth when `storage.preserve_configs_on_startup` matches the current mode (default `prod`): if any `agent_configs` row exists for the current `tenant_id` + `instance_id` + `agent_id`, skip `reset_config_db` / `save_draft` / `save_config` and log that configs were preserved from the DB. Empty DBs still get seeded from defaults. Local demos keep the existing reset/reseed behavior unless the setting is changed to `local` or `any`.

### Testing
- [x] Tested locally; tests pass (`uv run pytest tests/unit/test_preserve_configs_on_startup.py`)

## Summary by CodeRabbit

* **New Features**
  * Added `storage.preserve_configs_on_startup` (default: `prod`) to control whether startup preserves existing tenant/instance/agent configuration.
  * Travel Agent and demo/manager startup seeding now skip overwriting when preservation is enabled and `storage.mode` matches (or `any` is selected).

* **Bug Fixes**
  * Prevent draft/config resets when an existing config row is already present.
  * Ensure preservation checks are correctly scoped to the active tenant and instance.

* **Tests**
  * Added unit tests covering mode/preservation combinations, invalid-value fallback, async existence checks, and seeding vs. preservation behavior.


sami-marreed:
Addressed CodeRabbit review:

- Updated `setup_demo_manage_config` docstring for conditional preserve/reseed behavior.
- Extracted `resolve_preserve_existing()` in `config_store.py` and wired both demo setup and Travel Agent CLI through it.

Skipped the docstring-coverage 0% warning — new helpers already document their behavior.

offerakrabi:
# PR Review: #514 — feat: preserve agent configs from DB on startup when storage is prod


The Minor findings are marginal in my opinion, can be deferred.


## Summary

**Verdict: merge with fixes**

The design is sound and directly addresses the issue requirements — scoping preservation to `tenant_id + instance_id + agent_id`, defaulting to `prod`-only, and treating an empty DB as a fresh-seed case. `resolve_preserve_existing` is a clean extraction point used consistently by both `setup_demo_manage_config` and the Travel Agent CLI path. The `should_preserve_existing_configs` unit tests are thorough and cover all valid/invalid/fallback combinations. SQL uses `?` placeholders throughout, which `ProdRelationalStore` rewrites to `$N` at query time — consistent with the rest of the module.

**Strengths**
- `should_preserve_existing_configs` is fully unit-tested: all three valid values plus invalid-fallback, across both `mode=prod` and `mode=local`.
- `has_any_config` is scoped to `tenant_id + instance_id + agent_id`; the test `test_has_any_config_respects_tenant_and_instance` confirms isolation across tenants.
- The integration tests exercise all three `setup_demo_manage_config` branches: preserve active, preserve-on but empty DB (seed anyway), and preserve gate off (overwrite).
- `resolve_preserve_existing` is correctly placed in `config_store.py` alongside the functions it gates, keeping the preservation logic co-located with the storage layer.

## Issues

### 🟠 Major

**[MAJ-1] `asyncio.run()` crashes when called from an already-running event loop**
`src/cuga/backend/server/demo_manage_setup.py:530`

**What's wrong:** `asyncio.run(resolve_preserve_existing(agent_id))` creates a new event loop and blocks the calling thread. This works fine when `setup_demo_manage_config` is called from a synchronous CLI entrypoint, but if it is ever invoked from within a coroutine or any async context (e.g. a test that uses `asyncio.run()` around a fixture that calls setup, or a future refactor that wires setup into an async startup hook), it raises `RuntimeError: This event loop is already running`. The same pattern appears again at line 758 (`asyncio.run(_setup())`), and in `src/cuga/cli/main.py:1749`. This is consistent with the rest of `setup_demo_manage_config` which already used `asyncio.run(_setup())`, but the PR adds a second `asyncio.run()` call in the same synchronous function, so the function now has *two* sequential top-level loop creations that cannot be combined if a caller already has a loop.

**What it can cause:** A future FastAPI/Starlette startup event that calls `setup_demo_manage_config` during server boot (an `on_startup` handler is an async context) will raise `RuntimeError: This event loop is already running` before any config is checked or seeded, crashing startup entirely. It also makes the function untestable from `pytest-asyncio` fixtures without a `nest_asyncio` workaround.

---

### 🟡 Minor

**[MIN-1] `resolve_preserve_existing` log path has no test coverage**
`tests/unit/test_preserve_configs_on_startup.py`

**What's wrong:** The test file exercises `resolve_preserve_existing` only as a side-effect of calling `setup_demo_manage_config`. All three branches of the function are exercised indirectly, but no test uses `caplog` or any logging capture to assert that the "Preserving existing agent configs from DB" log message is emitted when the function returns `True`. That log line is the operator's sole runtime signal that preservation fired rather than a reset.

**What it can cause:** If the `logger.info(...)` call in `resolve_preserve_existing` were accidentally removed or placed under the wrong condition, no test would fail. Operators who rely on that log entry to confirm production configs were not overwritten after a restart would get silent behavior with no indication of whether preservation actually ran.

---

**[MIN-2] Travel Agent `start` path does not cover the preserve-existing branch in any test**
`src/cuga/cli/main.py:1749-1778`

**What's wrong:** The test file (`test_preserve_configs_on_startup.py`) has thorough coverage for `setup_demo_manage_config` but no test exercises the Travel Agent `start` command's equivalent preserve-gate (lines 1749–1778 of `main.py`). The Travel Agent path has a slightly different structure: it does not call `setup_demo_manage_config`, and it skips seeding a `save_config` (published version) entirely — only `save_draft` is written. None of this is tested.

**What it can cause:** A regression in the Travel Agent's preserve logic (e.g., the `if not preserve_existing:` block inverted or the `asyncio.run` call removed) would not be caught by any unit test. The PR description says tests pass against `tests/unit/test_preserve_configs_on_startup.py`, which does not touch this code path.

---

sami-marreed:
Addressed review feedback:

**[MAJ-1] (offerakrabi)** — Added `config_store.run_sync()` (same get_running_loop → thread-pool pattern as knowledge adapter / secrets sync wrappers). Demo setup and Travel Agent bootstrap now use it instead of bare `asyncio.run`. Covered by `test_run_sync_works_when_event_loop_already_running`.

**Seeding log (CodeRabbit)** — Message now logs `preserve_gate` + `mode` instead of claiming “no existing rows” when the gate is simply off.

**Deferred (per review):** MIN-1 preserve log assertion, MIN-2 Travel Agent path unit test.

**Not taking (CodeRabbit seed-if-absent / drop global reset):** `reset_config_db()` only deletes local SQLite (`cuga.db` + wal/shm), not Postgres. Under `storage.mode=prod` the wipe does not touch prod rows; under local with preserve gate off, full demo reset remains intentional. Scoped transactional seed-if-absent is out of scope for #511.

offerakrabi (approved):

PR #545, opened 2026-07-26
fix(appworld): auth/token lifetime, sandbox clock freeze, schema types, and prompt scoping

### LINKED ISSUE #541: [Bug]: Optional tool parameters are all rendered to the model as `str` (anyOf wrapper not unwrapped)
## What happened

Every **optional** parameter of every tool was presented to the model as type `str`, regardless of its real type.

Pydantic renders an optional field as `anyOf: [{...}, {"type": "null"}]` with **no top-level `type` key**. The renderers read the type with `prop.get("type", "string")`, so the lookup missed and silently fell back to the `"string"` default.

Affected sites:

- `cuga_lite/prompt_utils.py` — `get_tool_docs()` / `find_tools` Parameters block
- `tools_env/registry/mcp_manager/mcp_manager.py` — `_convert_mcp_parameters_to_openapi_format`
- `tools_env/registry/mcp_manager/mcp_manager.py` — `_convert_trm_parameters_to_openapi_format`

Concrete impact on AppWorld `321ec38_1` ("add a phone alarm 2 hours before my flight day and time"): `phone_create_alarm` has no date field, so `repeat_days` (a **list**) is the only place the day can be expressed. It rendered as:

```
- `repeat_days`: str (optional) -
```

with an empty description. The agent concluded "the alarm tool only takes HH:MM", dropped the day, and failed the assertion `repeat_days: None == ["wednesday"]`.

The same fallback also mistyped numeric params — `page_index`, `page_limit`, `min_rating`, `min_duration` all showed as `str`.

## How to reproduce

```python
from typing import Optional
from pydantic import BaseModel, Field
from langchain_core.tools import StructuredTool
from cuga.backend.cuga_graph.nodes.cuga_lite.prompt_utils import PromptUtils

class Args(BaseModel):
    time: str = Field(description="HH:MM")
    repeat_days: Optional[list] = Field(None, description="")
    snooze_minutes: Optional[int] = Field(15, description="Snooze minutes.")

tool = StructuredTool(name="t", description="d", args_schema=Args, func=lambda **k: None)
print(PromptUtils.get_tool_docs(tool)[0])
```

Before the fix every optional line reads `str`; `repeat_days` should be `list` and `snooze_minutes` should be `int`.

## Environment

Any model, any tool surface — the bug is in doc rendering, not in the model or a specific benchmark. Surfaced on AppWorld but not AppWorld-specific.

Also affects OpenAPI 3.1 style `type: ["string", "null"]`, which is a list rather than a string.

Fix: resolve the type through a shared helper that unwraps `anyOf`/`oneOf`/`allOf` and the 3.1 list form before falling back to `"string"`. `openapi_parser_v0.py` and `mcp_manager/adapter.py` already handle unions correctly and are unaffected.

### LINKED ISSUE #542: [Bug]: Reflection recommends out-of-scope actions and endorses substring matching for record selection
## What happened

Two distinct failure modes, both traced to `cuga_lite/reflection/prompts/reflection_system.jinja2`. Reflection output is appended inside observations, and the executor generally complies with it.

### 1. It invents work when the task is already done

Three rules push hard against ever concluding, with only a weak completion escape hatch:

- "**When no results are found**: Do not quit early or declare the task impossible."
- "**No Results Handling**: ... Never recommend quitting when results are empty."
- "**Skepticism and Non-Triviality Check**: Always approach the `Current Task` with a healthy dose of skepticism ... Do not assume simplicity; instead, proactively look for reasons why the task might be more involved than it seems."

With the task complete and nothing legitimate left to recommend, reflection produces an out-of-scope directive. On AppWorld `4d12842_1` ("delete all my archived gmail threads from before this month") it recommended, twice:

> Use the `simple_note_create` tool (or any note-taking tool) to store the `summary` string

The agent complied. Its Gmail work was correct, but the run failed on an unexpected state change:

```
{"simple_note.Note", "gmail.UserEmailThread"} == {"gmail.UserEmailThread"}
In left but not right: ["simple_note.Note"]
```

Same pattern on `5238afc_1`. Neither task mentions notes.

### 2. It tells the agent to substring-match names

The "Keyword Search Strategy" tip reads:

> When performing keyword searches on results, use broader search terms to catch more values. For example, if a user mentions "t-shirt of type x", use broader terms like "shirt" or "x" ... rather than the exact phrase.

Broadening is correct for **discovery** and wrong for **selection**. On `5238afc_1` ("order all weightlifting benches in my cart") the agent built `bench_keywords = ["bench", ...]` and matched `product_name`, pulling a **workbench** into the order:

```
{"workbench", "weightlifting bench"} == {"weightlifting bench"}
{820, 821, 823, 825, 1686}           == {820, 821, 823, 825}
```

Related selection failures in the same run, all name- or position-based rather than field-based:

| Task | Mistake |
|---|---|
| `6588a51_1` | Took the first t-shirt order returned instead of the most recent — "last ordered" needs an explicit sort |
| `16be9ce_1` | Ignored "ordered on my sister's birthday last year" and priced search result #1 ($90 vs $769) |
| `9bf2c8a_1` | Hardcoded `quantity=1` moving cart→wishlist instead of carrying the source quantity of 3 |
| `baeb104_1` | Added the user herself to the recipient list of an email meant for her roommates |

## How to reproduce

Run AppWorld `test_easy` with `DYNACONF_ADVANCED_FEATURES__REFLECTION_ENABLED=true` (set in `benchmarks/appworld/config/appworld.env`; `settings.toml` defaults it to false). `5238afc_1` and `4d12842_1` reproduce the unrequested-note failure; `5238afc_1`, `6588a51_1`, `16be9ce_1` reproduce the selection failures.

## Environment

Any benchmark or deployment with reflection enabled. Surfaced on AppWorld, not AppWorld-specific.

Suggested direction:

- Add a **scope discipline** rule: never recommend an action the task did not ask for; recording summaries, writing notes, sending confirmations and tidying up mutate state the task is judged on. Make "task complete, stop" an explicitly valid conclusion.
- Split the keyword tip into **discovery vs. selection** — keep broadening for discovery, and require exact selection on the authoritative structured field, with explicit sorting for ordinal/superlative wording ("last", "cheapest"), qualifiers treated as required filters, attribute preservation when moving items, and precise recipient/membership sets.

### LINKED ISSUE #543: [Bug]: Agent can mistake its internal todo list for a real task-manager app and mutate user data
## What happened

`create_update_todos` is the agent's internal planning scratchpad, but nothing tells the agent that. On AppWorld `5238afc_1` the agent fetched the user's **real Todoist Inbox** and tried to complete tasks whose titles matched its own internal plan items:

```python
tasks_resp = await todoist_show_tasks_projects_project_id_tasks_get(project_id=0)
target_titles = {
    "Search Amazon for weightlifting benches",
    "Add each found bench to the Amazon cart",
    "Place the order for all items in the cart",
}
task_ids = [t["task_id"] for t in tasks if t["title"] in target_titles]
for tid in task_ids:
    await todoist_update_task_tasks_task_id_patch(task_id=tid, is_completed=True)
```

Those three strings are the agent's own todos, not anything the user asked about — the task was only "Place an order for all weightlifting benches in my amazon cart."

No `todoist.Task` appears in the run's changed-model set, so nothing was actually mutated. That is luck, not safety: on any user whose Todoist happens to hold a similarly titled task, this silently marks their real task complete.

## How to reproduce

Run AppWorld `test_easy` task `5238afc_1` with `enable_todos` on. The final code block performs the Todoist lookup and patch loop above. Reproduces more generally wherever a task-manager app (Todoist, Simple Note) is connected and the agent is planning with `create_update_todos`.

## Environment

Any deployment with `enable_todos` and a connected task-manager application. Not AppWorld-specific — the risk is highest outside benchmarks, where the user has real data.

The guidance needs to live on the `StructuredTool`'s `description=` in `cuga_agent_core/execution/todos.py`, since that is what reaches the model — the inner function's docstring is **not** bound, so documenting it there alone has no effect. Two prompts also describe the tool independently and would each need it:

- `cuga_lite/prompts/mcp_prompt.jinja2`
- `cuga_supervisor/prompts/supervisor_lite_prompt.jinja2`

Suggested wording: the list is a private scratchpad, is not a connected application, is never shown to the user, must never be mirrored or synced into a real task manager, and its titles must never be matched against records in a connected app. Tasks living in a connected app are in scope only when the user's request refers to them.

### LINKED ISSUE #544: [Bug]: AppWorld benchmark reliability — file_system token TTL, action-task answer matching, and sandbox wall-clock freeze
## What happened

Grouped tracker for AppWorld-specific reliability gaps found while smoke-testing sandbox authentication and the full tool surface. Each blocks or distorts benchmark runs; none is a model-capability problem.

### 1. `file_system_access_token` expires mid-task

AppWorld app tokens have a TTL as short as 10 minutes. Long tasks outlive it, and the auto-injected `file_system_access_token` — which `api_overrides` threads into cross-app receipt and attachment calls — goes stale mid-run, failing calls that would otherwise succeed. Tokens need refreshing ahead of the shortest TTL, `auth_apps` should be best-effort rather than fatal, and the token should be hidden from the agent-facing OpenAPI so it is never treated as a parameter the agent must supply.

### 2. Action tasks fail answer matching against a null ground truth

AppWorld tasks split into QUERY tasks (a value is the answer) and ACTION tasks (the world state is the answer, ground-truth answer is null). The final-answer path submits an answer for both, so ACTION tasks get compared against null and fail on answer matching despite performing the actions correctly. They need classifying as `N/A` before final-answer extraction, on both the plain FinalAnswerAgent and SDK paths.

### 3. Sandbox wall-clock freeze hangs `await`

Benchmark mode freezes the wall clock so tasks see a fixed "today". Freezing `monotonic`/`perf_counter` along with `datetime`/`date`/`time` means any `await` inside user code — notably LLM-backed `find_tools` — never observes elapsed time and blocks until the sandbox timeout. The wall clock needs freezing without the monotonic clocks.

## How to reproduce

Run a slice of AppWorld hard tasks end to end:

```bash
# benchmarks/appworld/eval_config.toml -> eval_key = "test_challenge_hard"
./benchmarks/appworld/eval.sh
```

1. Any task running past ~10 minutes and touching a cross-app attachment or receipt call hits the stale token.
2. Any ACTION task (e.g. `36705ae_1`, `988af8e_1`) fails answer matching while its state assertions pass.
3. Any task calling `find_tools` inside `await` under frozen time hangs to `SANDBOX_EXECUTION_TIMEOUT`.

## Environment

AppWorld benchmark only (`DYNACONF_ADVANCED_FEATURES__BENCHMARK=appworld`), via `benchmarks/appworld/config/appworld.env`. Observed on both Opus and gpt-oss-120b, so not model-specific.

Two further AppWorld problems found alongside these are **not fixable in this repo**, recorded here so they are not re-investigated:

- **`988af8e_{1,2,3}` are unpassable.** `ground_truth/evaluation.py:122` calls `models.changed_fields`, which does not exist — `appworld/collections/models.py` defines only `changed_model_names`, `changed_records`, and `changed_field_names`, and the same file uses the correct name 19 lines earlier. It raises instead of asserting, inside `for subtask in updated_subtasks:` — so it fails for any agent that completes the sub-tasks and passes vacuously for one that skips them. Ceiling is 0.958. A sweep of all task evals found no other broken `models.*` call. Upstream AppWorld fix.
- **`3650990_1` scores 0.0 on a correct answer.** The agent answered `"Yes"` correctly, but the eval harness discards `agent_answer` and forces `score=0.0` whenever a task errors, ignoring the real pass percentage. Harness-side, fixed in `cuga-eval`.

sami-marreed:
## Bug fix

Fixes #541, fixes #542, fixes #543, fixes #544

### Summary

AppWorld reliability work: auth/token lifetime, sandbox wall-clock freeze, answer
classification, JSON-Schema type resolution, and two prompt-scoping fixes found by
analysing failing trajectories.

Three commits, each independently reviewable:

| Commit | Scope | Issue |
|---|---|---|
| `92b5cd78` | AppWorld reliability and agent plumbing (auth, classifier, sandbox freeze, LLM plumbing, schema nits) | #544, #541 |
| `a0a98ec3` | Reflection prompt: stop suggesting out-of-scope work; tighten record selection | #542 |
| `b4b2febd` | Todo list is a private scratchpad, not a connected task app | #543 |

### 1 · AppWorld reliability and plumbing (`92b5cd78`) — #544, #541

- **Auth / file_system token** — AppWorld app tokens have a TTL as short as 10 minutes, so long tasks outlive them and the auto-injected `file_system_access_token` goes stale mid-run. Now refreshed ahead of the shortest TTL; `auth_apps` is best-effort; the token is hidden from agent-facing OpenAPI via overrides.
- **ACTION vs QUERY classifier** — ACTION tasks have a null ground-truth answer, but the final-answer path submitted one anyway, so they failed answer matching despite performing the actions correctly. Now classified as `N/A` before final-answer extraction, on both the plain FinalAnswerAgent and SDK paths.
- **Sandbox wall-clock freeze** — freezing `monotonic`/`perf_counter` alongside `datetime`/`date`/`time` meant any `await` inside user code (notably LLM-backed `find_tools`) never observed elapsed time and blocked until sandbox timeout. Only the wall clock is frozen now.
- **LLM plumbing** — skip native `json_schema` for Claude/Bedrock/GCP via `ChatOpenAI`; detect Azure/LiteLLM `gpt-5*` / `o4` as reasoning models so non-default temperature is omitted.
- **Prompt / schema nits** — uncap `find_tools` result count; resolve optional/`anyOf` JSON-Schema param types via a shared `json_schema_type` helper (`MCPManager` + `PromptUtils`); tighten NL auto-continue for multi-phase turns; add Gemma 4 context window sizes.

The `anyOf` fix (#541) matters more than it looks. Pydantic renders optionals as `anyOf: [{...}, {"type": "null"}]` with **no top-level `type`**, so the `prop.get("type", "string")` lookup missed and fell through to its default — meaning **every optional parameter of every tool was shown to the model as `str`**. On `321ec38_1` the agent skipped a list-typed `repeat_days` because it read as free text; numeric optionals like `page_index` and `min_rating` were mistyped too.

### 2 · Reflection prompt (`a0a98ec3`) — #542

Reflection output is appended inside observations, and the executor generally complies with it.

**Scope discipline.** Three existing rules push hard against ever concluding ("do not quit early", "never recommend quitting", "do not assume simplicity") with only a weak completion escape hatch. With nothing legitimate left to do, the reflection invented work — on `5238afc_1` and `4d12842_1` it recommended writing a `simple_note` summary the task never asked for. The executor complied and both failed on the unexpected `simple_note.Note` state change.

**Discovery vs. Selection.** The old "Keyword Search Strategy" tip told the agent to broaden its terms, which is right for discovery and wrong for deciding membership. On `5238afc_1` a substring match on `"bench"` pulled a *workbench* into an order of weightlifting benches.

### 3 · Todo list scoping (`b4b2febd`) — #543

On `5238afc_1` the agent fetched the user's **real Todoist Inbox** and tried to complete tasks whose titles matched its own internal plan items. No real task happened to match, so nothing was mutated — but on a user whose Todoist held a similarly titled task it would have silently changed their data. Nothing in the tool description or prompts said the list was internal.

## Solution

- **#541** — a shared `json_schema_type` helper unwraps `anyOf`/`oneOf`/`allOf` and the OpenAPI 3.1 `type: [...]` list form before falling back to `"string"`, applied at all three affected sites. `openapi_parser_v0.py` and `mcp_manager/adapter.py` already handled unions and were left alone.
- **#542** — added **Scope Discipline** as a fifth responsibility, making "task complete, stop" an explicitly valid conclusion. Replaced the keyword tip with **Discovery vs. Selection**: broadening kept for discovery only, plus exact selection rules for type/category, ordinal and superlative wording ("last", "cheapest"), qualifiers that pin down a specific record, attribute preservation when moving items, and recipient/membership sets.
- **#543** — stated in the tool's bound `description=` plus the two prompts that describe the tool separately (`mcp_prompt.jinja2`, `supervisor_lite_prompt.jinja2`).
- **#544** — token refresh, ACTION/QUERY classification, and selective clock freeze as described above.

> Reviewer note for #543: the guidance **must** live on the `StructuredTool`'s `description=`. The inner function's docstring is not what gets bound, so documenting it there alone never reaches the model.

## Evaluation

Single pass, no retries. **Both runs were executed against `92b5cd78` and predate the two prompt commits** — they are the baseline that motivated `a0a98ec3` and `b4b2febd`, not validation of them.

| Model | Set | Tasks | Accuracy (tasks fully passed) | Mean score | Run |
|---|---|---|---|---|---|
| **gpt-oss-120b** | `test_easy` | 43 | **74.4%** (32/43) | 0.895 | `test_easy_26-07--12h14m27s047ms` |
| **Opus** | `test_hard` | 10 | **70.0%** (7/10) | 0.961 | `test_hard_26-07--0h50m28s446ms` |

Accuracy counts a task only when every assertion passes; mean score is the average pass percentage across assertions. The gap between the two columns is the point — most failures here are a single wrong field or one unrequested side effect on an otherwise correct trajectory, which is why the mean stays high while the pass rate does not.

Two caveats on the accuracy figures, both detailed under *Known gaps*: `988af8e_1` cannot reach 1.0 because of an upstream eval bug, and `3650990_1` answered correctly but was forced to 0.0 by a harness error path. Excluding those two, Opus is 8/9 and gpt-oss-120b is 33/43.

What the failures surfaced, and where each is handled:

| Failure | Model | Addressed by |
|---|---|---|
| Unrequested `simple_note.Note` (`5238afc_1`, `4d12842_1`) | gpt-oss-120b | `a0a98ec3` · #542 |
| `"bench"` matched a workbench (`5238afc_1`) | gpt-oss-120b | `a0a98ec3` · #542 |
| Took first t-shirt order, not the latest (`6588a51_1`) | gpt-oss-120b | `a0a98ec3` · #542 |
| Dropped `quantity` when moving cart→wishlist (`9bf2c8a_1`) | gpt-oss-120b | `a0a98ec3` · #542 |
| Added the user herself as a recipient (`baeb104_1`) | gpt-oss-120b | `a0a98ec3` · #542 |
| Touched the real Todoist Inbox (`5238afc_1`) | gpt-oss-120b | `b4b2febd` · #543 |
| `repeat_days` skipped as `str` (`321ec38_1`) | Opus | `92b5cd78` · #541 (partial — see below) |

## Known gaps, not fixed here

- **`321ec38_1` is not closed by the type fix.** Verified post-fix that `repeat_days` now renders as `list`, and the agent still omits it — the AppWorld spec gives the parameter an empty description, so it has the right type and no signal about its purpose. Needs a behavioural rule, not a renderer change.
- **`988af8e_1` has a ceiling of 0.958.** Its AppWorld eval calls `models.changed_fields`, which does not exist (`collections/models.py` defines only `changed_model_names`, `changed_records`, `changed_field_names`). It raises rather than asserts, so the test is unpassable for any agent that completes the sub-tasks and passes vacuously for one that skips them. Affects `988af8e_{1,2,3}`; a sweep of all task evals found no other broken call. Upstream fix, recorded in #544.
- **`3650990_1` scored 0.0 on a correct answer.** The agent answered `"Yes"` correctly; the eval harness discards `agent_answer` and forces `score=0.0` whenever a task errors, ignoring the real pass percentage. Harness-side, fixed separately in `cuga-eval`, recorded in #544.
- **`find_tools` output trimming** was prepared and deliberately left out of this PR — it changes what every tool looks like to the model, so it ships on its own.

### Testing
- 284 tests pass (`cuga_lite`, `tools_env/registry`, `cuga_supervisor`)
- All three modified jinja templates render
- `ruff check` / `ruff format --check` clean
- Verified the todos guidance reaches the *bound* tool description, not just the docstring

## Summary by CodeRabbit

* **New Features**
  * AppWorld action-only tasks can now complete without generating an unnecessary final response.
  * Tool discovery can return all relevant tools for a workflow (no fixed cap).
  * Authentication tokens refresh automatically when stale, with safe fallback if refresh fails.
* **Bug Fixes**
  * Time-freezing during code execution preserves asyncio behavior and restores timing safely.
  * Improved handling for additional model providers and model name variants (reasoning-model detection, JSON-schema typing).
* **Tests**
  * Added coverage for AppWorld action classification, token refresh staleness, JSON-schema typing, and time-freezing correctness.


Sergey-Zeltyn:
**Extension of the clock-monkeypatch concern: a lock is not enough**

CodeRabbit flagged concurrent sandbox executions interleaving the frozen clock (`code_wrapper.py`). There's a second failure path that changes what the right fix is: with `sandbox_mode=native` (in-process `NativeSandboxExecutor`), *during* any `await` inside frozen user code, **cuga's own coroutines on the same event loop see the frozen wall clock** — timings, logs, and notably this PR's own `_is_stale()`, which computes `time.time() - ts` as a large negative age and silently stops refreshing tokens while the freeze is active.

Serializing executions fixes the interleaving but not this — cuga coroutines share the event loop even with serialized sandbox runs. Suggest confining the freeze to subprocess-isolated executors, and making `native` + `fake_datetime` an explicit error (or a no-op with a loud warning).

Sergey-Zeltyn:
**SDK classifier gate keys on the string `"N/A"`, not the decision (`sdk.py`)**

In the AppWorld final-answer path:

```python
if final_answer != "N/A":
    # ... extraction chain runs
```

If the planner's `final_answer` already happens to be the literal `"N/A"` while the classifier says QUERY (or the classifier is disabled/fails), final-answer extraction is silently skipped — a behavior change vs. pre-PR, where extraction always ran in this path. Suggest gating on an explicit `is_action` boolean set by the classifier branch, so the skip only ever happens on a positive ACTION classification.

Sergey-Zeltyn:
**Known gap missing from the PR body: real-timestamp leak via Variables-Summary**

The freeze covers user-code clock reads, but `variable_utils.py` stamps real wall-clock `Created:` times into the variables summary shown to the model, so the actual date still reaches the agent around the frozen clock.

This demonstrably caused a full task failure in a pre-PR run (`6474048_1`: reflection concluded "all payment cards … are expired (expiry dates in 2022 or 2023, and the current date is June 2026)" from those stamps and gave up), and the stamps are still present in traces from this branch (`Created: 2026-07-26 …` in Variables Summary blocks). The model happened to out-reason it in the post-PR run, but it remains a latent failure source. Cheap fix (stamp with the task's frozen datetime, or drop the timestamp from the summary) — or at least a bullet under *Known gaps, not fixed here*.

Sergey-Zeltyn (commented):
Testing results were good:
Easy Appworld 90.7%
Med Appworld 65%

Please take a look at my comments and then we, most probably, can approve it

sami-marreed:
Thanks Sergey — good catches, and thanks for running the evals. Checked all three against the branch.

## ✅ SDK `"N/A"` gate — agreed, fixing

Confirmed against `main`: extraction ran unconditionally pre-PR. `FinalAnswerAgent` already gates on a boolean; `sdk.py` is the divergent copy. Switching to an explicit `is_action` flag.

Scanned all 101 tasks in the three recent runs — **zero** literal-`"N/A"` planner answers, so this changes no observed trajectory.

## 📝 Variables-Summary timestamp — real, documenting as a known gap

Right about the leak, wrong file: it's `agent_state.py` (`VariableMetadata.__init__` → `datetime.now()`), not `variable_utils.py`. It does reach the model — 32 hits in this branch's own `321ec38_1` trace:

```
## todos
- Created: 2026-07-26 02:02:31     ← real clock; task data is 2022/2023
```

Deferring rather than fixing here, for two reasons:

1. **The field is dead weight either way.** `add_variable` rebuilds the metadata on every write, so `created_at` is *last-write* time — and the same write moves the variable to the end of the ordering. Position and timestamp encode one event. The right change is deletion, not freezing.
2. **It edits model-facing prompt text on 95/101 tasks**, which invalidates the 90.7%/65% numbers as evidence for the merged code. Follow-up issue with its own run.

One correction: I couldn't reproduce the `6474048_1` failure — no trace in either repo has that reflection text (only July-2025 runs, no expired-card reasoning). The trace evidence above stands on its own; I'd drop that citation.

## ⚠️ Clock freeze in cuga's coroutines — premise right, example wrong

Agreed the patch is module-global:

```python
_cuga_dtmod.datetime = _CugaDatetime      # module attr, not namespace
_cuga_tmod.time = lambda: _cuga_epoch
```

Default `local` executor runs in-process on the same loop, so an `await` in user code does expose the frozen clock process-wide. (Only in benchmark mode — `is_benchmark_mode()` gate.)

But `_is_stale()` isn't reachable from it — `BaseAuthManager` lives in the registry, which `eval.sh` starts as a **separate process** (`run_registry.sh`, reached over HTTP). And making `native` + `fake_datetime` an error would disable the only config this PR was evaluated in.

Proposing: note the global-mutation caveat under *Known gaps*, leave the freeze as is.

---

So: fixing the `"N/A"` gate, adding *Known gaps* bullets for the other two plus follow-up issues. Let me know if you'd rather see any of them fixed in this PR instead — happy to, it just means a re-run before the numbers mean anything.

Sergey-Zeltyn (approved):
I agree with your last suggestions

PR #548, opened 2026-07-27
fix(cuga-lite): preserve stdout on error and hint assignment-RHS tool names

### LINKED ISSUE #547: [Bug]: Local executor drops stdout on exception and suppresses tool-name hints for assignment RHS
### What happened
Two related local-executor failures compound when the agent runs `find_tools` and later lines in the **same** script error out (observed on AppWorld task `baeb104_1`):

**1) Stdout is discarded when execution raises**
In `local/local_executor.py`, stdout is captured with `redirect_stdout(StringIO)` while `_async_main()` runs. `f.getvalue()` is only read after a successful return. If a later line in the concatenated script raises (e.g. `NameError`), `code_executor.py` catches the exception and returns only `format_error(...)` — any `print()` output from an earlier successful `find_tools(...)` in that same script is never returned. The agent explicitly printed discovery results but never saw the real Gmail tool name because a subsequent line failed.

Because `_async_main` ends with `return locals()`, a mid-script exception also skips `context_locals.update(result_locals)`, so partial in-script assignments are lost too.

**2) Tool-name correction is suppressed for assignment RHS**
`_unknown_tool_correction` (with difflib cutoff 0.6) would have suggested `gmail_send_email_emails_post` for a fabricated `gmail_send_message_send_message_post`, but `_missing_name_usage` only treats the missing name as a fabricated tool when it appears as a **Call**. When the agent writes `send_email = gmail_send_message_send_message_post` (assignment RHS, not a call), `called=False` and the hint is suppressed. Wrong tool names used as `await tool(...)` still get "Did you mean:" — assignment-shaped mistakes do not.

Together: discovery output hidden (1) and no difflib recovery for the common alias pattern (2), so the agent retries and eventually gives up.

### How to reproduce
1. Run CugaLite in **local** executor mode (`CodeExecutor.eval_with_tools_async`, `mode='local'`).
2. Provide a callable `find_tools` in `_locals` that returns markdown including a real tool name (e.g. `gmail_send_email_emails_post`).
3. Execute a single script that:
   - calls `find_tools("send email with attachment", "gmail")` and `print`s the result, then
   - references a fabricated name on the next lines, e.g. `send_email = gmail_send_message_send_message_post` followed by another line that raises `NameError`.
4. Observe the returned string: only `Error during execution: ...` traceback — **no** printed find_tools output.
5. For bug 2 alone: run code `send_email = gmail_send_message_send_message_post` with `available_tools` containing `gmail_send_email_emails_post` — no `[tool-name correction]` hint. Contrast with `result = await gmail_send_message_send_message_post()` which does get "Did you mean: gmail_send_email_emails_post".

Relevant code:
- `src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/local/local_executor.py` (`execute`, `format_error`, `_unknown_tool_correction`, `_missing_name_usage`)
- `src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/code_executor.py` (exception path ~236–241)

### Environment
- Repo: cuga-agent (local executor / CugaLite code path)
- Observed during AppWorld benchmark runs (e.g. chain `baeb104_1.json`)
- Python 3.x, `asyncio` local sandbox (not E2B-only)

Example trajectory (paraphrased from `baeb104_1`):
- Steps 21 and 35: agent ran `find_tools(..., "gmail")` with a `print` in the same script; a later line raised `NameError`; harness returned error only — agent never saw discovery output.
- Gmail alias used assignment form (`send_email = gmail_send_message_send_message_post`); no difflib hint. file_system-style errors that used `await wrong_name(...)` did receive "Did you mean:".

Related: #312 (undefined tools at invoke time), #546 (validate find_tools output upstream). Complements PR #416 guardrail intent — assignment RHS for fabricated tool names is an uncovered case.

### Checklist

sami-marreed:
## Bug fix

Fixes #547

### Summary
- Preserve redirected stdout when local execution raises, and surface it in `format_error` as `Output before error:` so earlier `find_tools` / `print` output is not dropped.
- Treat fabricated tool names on assignment RHS (`send_email = gmail_send_message_…`) as tool usage for `[tool-name correction]` / difflib hints; keep bare NameErrors for plain variable refs like `print(formatted_total)`.

### Testing
  - `uv run pytest src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/tests/test_code_executor.py -v` (28 passed)

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved any stdout produced before a later execution error, so earlier diagnostics aren’t lost.
  * Improved error/timeout messaging to include “Output before error” when prior output was captured.
  * Enhanced tool-name correction so fabricated tool-like names in assignment/alias-style cases produce more accurate “Did you mean” suggestions.
* **Tests**
  * Added new unit tests covering output preservation across failures, tool discovery output behavior, and assignment-RHS tool-shaped usage detection.


offerakrabi:
# PR Review: #548 — fix(cuga-lite): preserve stdout on error and hint assignment-RHS tool names

Only small findings, first one just a variable name change and the second one could be deffered, not sure it really merits another test

## Summary

**Verdict: merge with fixes.**
The two bugs are real and well-scoped, the approach is minimal and correct, and the new tests exercise real code paths through `CodeExecutor.eval_with_tools_async`, not mocks. One small fix is needed before merge (MINOR-1).


**Notes / out of scope:**
- Pre-existing tests in the same file (`test_basic_execution_local`, `test_execution_with_variables`, etc., lines 24–164) are missing `@pytest.mark.unit`. That is not introduced by this PR and is not filed here.

## Issues

### 🟡 Minor

**[MINOR-1] Test uses the old variable name `called` for a return value renamed to `used`**
`src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/tests/test_code_executor.py:245-256`

**What's wrong:** `_missing_name_usage` was renamed in this PR to return `(used, defined)` instead of `(called, defined)`, reflecting the broadened semantics. `test_missing_name_usage_detects_assignment_rhs` still unpacks both calls into `called, defined` and asserts `called is True`. The test passes because the value is correct, but the local name contradicts the function's contract — readers checking the test to understand the API see the old "call" framing, not the new "used" framing.

**What it can cause:** A future reader verifying what `_missing_name_usage` means by looking at its test will learn that the first return value indicates a *call*, which is now false — it also fires for plain assignment-RHS references that are not calls. The mismatch makes it harder to reason about whether a new edge case (e.g. `x: Tool = fabricated_name`) should set `used` or `called`.

---

### 🔵 Nice to Have

**[NICE-TO-HAVE-1] The `TimeoutError + captured_stdout` path has no test**
`src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/local/local_executor.py:129-137`

**What's wrong:** `format_error` has a separate branch for `asyncio.TimeoutError` that also checks `getattr(error, "captured_stdout", None)` and prepends `Output before error:`. The stdout-preservation test (`test_stdout_preserved_when_later_line_raises`) only exercises the `NameError` path. The timeout branch is untested — if the `captured_stdout` lookup or prepend logic in that branch were accidentally removed or broken, no test would catch it.

**What it can cause:** Timeout regressions are silent: a task that prints `find_tools` output and then times out will lose that output, and the only symptom is the agent not seeing the discovery output — which looks identical to the original bug in #547.

---

sami-marreed:
Addressed [MINOR-1](https://github.com/cuga-project/cuga-agent/pull/548#issuecomment-5101519706): renamed `called` → `used` in `test_missing_name_usage_detects_assignment_rhs` so it matches `_missing_name_usage`'s `(used, defined)` contract (`59695e48`).

Deferring NICE-TO-HAVE-1 (timeout + `captured_stdout` coverage) as suggested.

offerakrabi (approved):

Issue #486, opened 2026-07-14
[Performance]: optimize CUGA loading across SDK and server (~30s)


sami-marreed:
### Issue type

[Performance] – speed, memory, or efficiency improvement

### What you want and why

CUGA loading is noticeably slow across multiple entry points (SDK and server). Cold start / initialization currently takes on the order of **~30 seconds**, which makes local development, SDK usage, and server bring-up feel unresponsive.

We want faster time-to-ready so that:

- SDK consumers can construct and use a CUGA agent without a long wait
- The server becomes ready to serve requests much sooner after process start
- Developers iterating locally are not blocked by a ~30s load tax on every restart

### How it could work

Investigate and reduce end-to-end load time across the shared initialization path used by the SDK and server. Likely areas to profile and tighten:

1. **Measure first** — instrument / profile cold start for both SDK and server to identify the top contributors (imports, config, model/client setup, registry, skills, tools, etc.).
2. **Defer expensive work** — lazy-load heavy modules, tools, skills, or clients until first use instead of at import/startup.
3. **Share and cache warm state** — avoid repeating the same setup when SDK and server share code paths; cache immutable startup artifacts where safe.
4. **Trim startup surface** — reduce eager imports and optional dependency loading on the critical path.
5. **Set a target** — aim for a much lower cold-start budget (e.g. seconds, not tens of seconds) with a simple benchmark so regressions are caught.

Success looks like: SDK and server reach a usable ready state far under ~30s on a typical developer machine, with the remaining cost explained and intentional.

### Links or extra context

- Affects multiple surfaces: **SDK** and **server** (shared loading path suspected).
- Observed load time: **~30 seconds**.
- No specific prior issue linked; please check for overlapping startup/import optimization work during triage.

### Checklist

- [x] I searched existing issues and this doesn’t look like a duplicate

offerakrabi:
@sami-marreed @nir-mashkif please review

## Plan: optimize CUGA startup (< 10s ready)

### Goal
Reduce cold-start ready time for SDK and server from ~30s to **under 10s**, with behavior kept **identical**.

### Ready definitions
- **SDK ready** = able to run first `invoke()` (import + `CugaAgent` construction; LLM/network time measured separately, excluded from the startup number).
- **Server ready** = able to receive the first chat message (browser opens).

### Approach: measure first, then optimize
**Phase 1 — Standalone startup-profiling tool** (run on demand, no CI gate) under `src/system_tests/profiling/startup/`:
[...]

PR #549, opened 2026-07-27
feat: validate shortlister tool names with retry (#546)

### LINKED ISSUE #546: [Feature]: Validate shortlister/find_tools output against real tool names with retry
### What you want and why
Shortlister and `find_tools` can return tool names the model invented that are not in the real candidate set. Downstream the agent then tries to call non-existent tools (related failure mode: #312). We need validation that every returned name exists in the known tool list, plus retry logic when validation fails, so discovery output is trustworthy before it reaches the agent loop.

### How it could work
1. After shortlister / `find_tools` structured output is parsed, intersect returned names with the actual available tool names for that app/query context.
2. If any name is unknown (hallucinated), reject the result and retry the LLM shortlist/find_tools call (bounded retries, e.g. 1–2) with feedback listing invalid names and the allowed set (or a compact reminder to only use provided names).
3. If retries still fail, fall back safely (e.g. empty match set, or best-effort filter dropping invalid names) and log clearly — never forward invented tool names as real discoveries.
4. Cover both cuga-lite `find_tools` / `PromptUtils` shortlisting paths and the API shortlister used for the same discovery schema where applicable.

### Links or extra context
- Parent epic: #353
- Related: #312 (invoke-time undefined tools — complementary; this issue is discovery-output validation)
- Touchpoints: `prompt_utils.py` shortlist/find_tools, `helpers/find_tools.py`, shortlister prompts / `bind_tools` shortlist path

sami-marreed:
## Feature

Closes #546

### Summary
Validate CUGA Lite shortlister / `find_tools` output against the real candidate tool set. On unknown (hallucinated) names, retry the LLM up to twice with feedback, then drop remaining invalid names. `find_tools` surfaces a clear note about filtered names; `shortlist_tool_names` (bind-tools cap) filters and logs. Classic API shortlister is out of scope.

### Testing
  - `uv run pytest tests/unit/test_shortlister_name_validation.py -m unit`
  - `uv run pytest tests/unit/test_langfuse_tracing.py -k shortlist_passes_explicit_run_config`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tool discovery and shortlisting by retrying invalid tool names and filtering unrecognized or duplicate results.
  * Added messaging when invalid tool names are removed from results.
  * Prevented knowledge-engine initialization errors from interrupting startup; failed initialization is cleaned up and reported with details.
  * Clearly reports knowledge availability as ready, disabled, or failed during startup.
* **Tests**
  * Added coverage for tool validation, retries, filtering, output messaging, startup failure isolation, and lazy-import scenarios.


offerakrabi:
# PR Review: #549 — feat: validate shortlister tool names with retry (#546)

## Summary

**Verdict: merge with fixes.** The core mechanism is correct and solves a real reliability problem — hallucinated tool names can no longer propagate to the agent. The retry loop with feedback and the final filter-and-note approach are clean. The change is well-scoped to CUGA Lite only, as intended.


## Issues

### 🟡 Minor

**[MINOR-1] `find_tools` "all invalid" branch is untested**
`src/cuga/backend/cuga_graph/nodes/cuga_lite/prompt_utils.py:523-526`

**What's wrong:** When `_ainvoke_shortlister_with_name_validation` exhausts all retries and every returned name is still invalid, `validated_details` is empty and `filtered_invalid_names` is non-empty. The code at line 523-525 returns `"No matching tools found for your query.\n\n{filtered_note}"`. This is a new code path added by the PR (the old code had a simple unconditional `"No matching tools found..."` return), and it has no test coverage. The only test for `find_tools` (`test_find_tools_mentions_filtered_invalid_names_after_retries`) always includes `tool_a` in the final attempt's valid results, so `validated_details` is never empty.

**What it can cause:** If the LLM consistently returns only hallucinated names across all three attempts, the agent receives `"No matching tools found for your query.\n\n**Note:** Filtered out 1 unrecognized tool name..."`. If a future refactor changes this branch (e.g., removes the `if filtered_note:` guard), the regression would be invisible because no test exercises it.

---

**[MINOR-2] Retry feedback only lists the *last* attempt's invalid names, not the union across all attempts**
`src/cuga/backend/cuga_graph/nodes/cuga_lite/prompt_utils.py:194-211`

**What's wrong:** The retry loop in `_ainvoke_shortlister_with_name_validation` overwrites `last_invalid` on every iteration. If attempt 0 returns `["ghost_a", "ghost_b"]` and attempt 1 returns only `["ghost_a"]`, the feedback injected for attempt 2 says "ghost_a is not a valid name" — it has forgotten about `ghost_b`. Similarly, the final `filtered_invalid_names` returned to the caller and shown in the user-visible note only reflects the final attempt's invalids, not the full set of names the LLM hallucinated across the session.

**What it can cause:** A user sees "Filtered out 1 unrecognized tool name: `ghost_a`" when in fact the LLM invented both `ghost_a` and `ghost_b` across attempts. The note is less useful for debugging. The retry feedback also misses a name, giving the LLM a weaker correction nudge on the final retry.

---

### 🔵 Nice to Have

**[NICE-TO-HAVE-1] `prompt_utils.py` is approaching the AGENTS.md 1000-line limit**
`src/cuga/backend/cuga_graph/nodes/cuga_lite/prompt_utils.py:1`

**What's wrong:** `AGENTS.md` states "No single file in this directory should exceed 1,000 lines of code if possible." The PR branch version is 826 lines (up from ~673 on main). The new validation helpers (`_partition_shortlist_details`, `_shortlist_retry_instructions`, `_format_filtered_tool_names_note`, `_ainvoke_shortlister_with_name_validation`) are 95 lines of closely related shortlister logic. There is no violation yet, but the next PR touching this file will likely cross 1000 lines.


---

src/cuga/backend/cuga_graph/nodes/cuga_lite/prompt_utils.py
```diff
@@ -106,9 +106,111 @@ class FindToolsOutput(BaseModel):
...
+                max_retries + 1,
+                last_invalid,
+            )
+            if attempt >= max_retries:
+                break
+            instructions = PromptUtils._shortlist_retry_instructions(
+                base_instructions or "",
+                last_invalid,
+            )
+        return last_valid, last_invalid
```
Sergey-Zeltyn: **Retry can discard valid tools found on an earlier attempt.**

`last_valid` is overwritten wholesale each iteration, and only the **final** attempt's valid set is returned. A legitimately discovered tool from attempt 1 is lost if a later attempt regresses:

| attempt | LLM returns | `last_valid` after |
|---|---|---|
| 1 | `tool_a` (valid), `fake_one` | `[tool_a]` |
| 2 | `fake_two` | `[]` &larr; `tool_a` lost |
| 3 | `fake_three` | `[]` |

Returns `([], [fake_three])`. Pre-PR behavior kept `tool_a` from the single call, so **the retry loop can produce strictly worse output than no retry at all**. LLM output is non-monotonic under retry, so this isn't hypothetical.

Suggest accumulating across attempts instead of replacing:

```python
accumulated: dict[str, Any] = {}
for attempt in range(max_retries + 1):
    ...
    valid, last_invalid = PromptUtils._partition_shortlist_details(...)
    for d in valid:
        accumulated.setdefault(d.name, d)
    if not last_invalid:
        break
return list(accumulated.values()), last_invalid
```

Worth a regression test too: all three current retry tests return `tool_a` on every attempt, which is exactly the shape that masks this. A case where a later attempt returns *fewer* valid names would catch it.

src/cuga/backend/cuga_graph/nodes/cuga_lite/prompt_utils.py
```diff
@@ -106,9 +106,111 @@ class FindToolsOutput(BaseModel):
...
+                    "all_tools": tools_as_dict,
+                    "instructions": instructions,
+                },
+                config=nested_langgraph_invoke_config(run_config),
+            )
+            last_valid, last_invalid = PromptUtils._partition_shortlist_details(
+                getattr(response, "result", None) or [],
+                valid_names,
+            )
+            if not last_invalid:
```
Sergey-Zeltyn: **Retrying on _any_ invalid name costs up to 3x the full payload.**

The loop retries whenever `last_invalid` is non-empty, even when the response is mostly usable. If the shortlister returns 9 valid names and 1 hallucination, we pay two extra round-trips — each re-sending `all_tools` with full `args_schema`, `_response_schemas`, and `_param_constraints` — to recover a name that gets dropped anyway.

`find_tools` is on the runtime discovery path, so this is a real latency and token cost. `test_shortlist_filters_invalid_names_after_two_retries` demonstrates it: `tool_a` is valid on attempt 1, yet 3 calls are made.

Suggest gating the retry on whether the result is actually unusable — e.g. `if not last_valid`, or an invalid-fraction threshold — rather than on the presence of any invalid name.

Sergey-Zeltyn (commented):
Two findings on the retry helper from a review pass. Both are about `_ainvoke_shortlister_with_name_validation`; the first is a correctness regression against pre-PR behavior.

Not re-raising the two open CodeRabbit findings — I verified both and they hold, but they're against the knowledge-engine portion of this PR, which is unrelated to #546.

src/cuga/backend/cuga_graph/nodes/cuga_lite/prompt_utils.py
sami-marreed: Fixed in `8e1eb836`. Valid details now accumulate across attempts (`dict` keyed by name), so a later all-invalid retry cannot drop an earlier usable tool. Also gated retries on an empty accumulated set — see sibling thread.

src/cuga/backend/cuga_graph/nodes/cuga_lite/prompt_utils.py
sami-marreed: Fixed in `8e1eb836`. Retries now run only when the accumulated valid set is still empty. A mostly-good shortlist (valid names + a hallucination) filters the bad names in one call instead of paying for two more full-payload round-trips.

src/cuga/backend/server/main.py
```diff
@@ -763,7 +763,17 @@ async def _knowledge_warmup_then_maybe_oobe_pdf():
         kb_config = KnowledgeConfig()
 
     if kb_config.enabled:
-        await initialize_knowledge_engine(app_state, kb_config)
+        try:
+            await initialize_knowledge_engine(app_state, kb_config)
+        except Exception as e:
+            logger.warning(f"Failed to initialize knowledge engine: {e}")
+            app_state.knowledge_engine = None
```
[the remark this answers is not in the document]
sami-marreed: Fixed in `8e1eb836`. Failure path now calls `aclose()` / `shutdown()` via `_rollback_partial_knowledge_engine` before clearing the reference. Added `test_partial_engine_is_torn_down_on_error`.

tests/unit/test_knowledge_init_failure_isolation.py
```diff
@@ -0,0 +1,157 @@
...
+            app_state.knowledge_engine = None
+            app_state.set_subsystem_status(
+                "knowledge",
+                "failed",
+                "Knowledge subsystem failed to initialize",
+                {"error": str(e)},
+            )
+    else:
+        app_state.knowledge_engine = None
+        app_state.set_subsystem_status("knowledge", "disabled", "Knowledge subsystem disabled")
```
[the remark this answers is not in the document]
sami-marreed: Fixed in `8e1eb836`. Extracted `run_knowledge_startup` as the production helper; lifespan and the unit tests both call it. The duplicated `_run_call_site` mirror is gone.

sami-marreed:
Addressed the minor items from the review in `8e1eb836` (also merged `main` to clear the conflict):

- **MINOR-1:** `test_find_tools_all_invalid_includes_filtered_note` covers the empty-valid + filtered-note path.
- **MINOR-2:** Invalid names are unioned across attempts for both retry feedback and the user-facing filtered note.
- **NICE-TO-HAVE-1:** Left as-is — file is still under the 1000-line soft limit; not splitting in this PR.

Also took Sergey's two retry-helper findings (accumulate valids; retry only when the shortlist is unusable) and the open CodeRabbit knowledge-init notes (teardown + production helper for the tests).

sami-marreed:
1. run both easy and med on aws gpt oss 120b
2. ask AI to validate if the PR helped

sami-marreed:
AppWorld results on this branch with **aws/gpt-oss-120b** (following up on the easy + med request):

| Run | Total | Completed | Passed | Failed | Pass rate |
|---|---:|---:|---:|---:|---:|
| `test_easy_13-08--9h50m30s954ms` | 43 | 43 | 38 | 0 | **88.4%** |
| `test_med_13-08--11h15m30s128ms` | 40 | 40 | 17 | 0 | **42.5%** |

Both suites finished. That is already better than the earlier `test_med` process-kill at 23/40 (#625).

**On the completed ≠ passed gap.** The dashboard’s Passed column is exact score `1.0` (full AppWorld TGC). Failed is exact score `0.0`. Everything in between is partial `match_rate` (`pass_percentage / 100`) — the task ran and some of its AppWorld tests passed, but not all. So this is not 23+5 silent skips:

- easy: 38 full pass, **5 partial**, 0 zeros
- med: 17 full pass, **23 partial**, 0 zeros

**Did this PR help?** These numbers cannot answer that. This change is a discovery guard (retry, then drop hallucinated shortlister names) — it should cut undefined-tool calls (#312 / #546), not move overall TGC by itself. Easy 88% / med 42% is a plausible gpt-oss-120b profile (easy nearly saturated, med still hard). Without a same-model, same-day `main` control, attributing the pass rate to name validation would be guessing.

What *would* show this PR working: traces with `Filtered out N unrecognized tool name` / retry-on-invalid, and fewer invoke-time unknown-tool failures vs `main`. If those notes are absent, gpt-oss-120b may already be emitting valid names (especially after the #639 `json_mode` fallback) and this PR is a no-op on these runs — still worth merging as the guard.

Sergey-Zeltyn:
# Review pass on current state (post-`8e1eb836`)

**Verdict: shortlister logic is now correct; blocked only by the `tests.yml` change, which is what turned CI red.**

## Prior findings — verified fixed in the diff ✅

- Valid tools now accumulate across attempts (`accumulated` + `setdefault`) — a bad retry can no longer discard an earlier usable name.
- Retries gated on an empty accumulated set — no more 3× payload for a mostly-valid shortlist.
- Invalid names unioned across attempts for both retry feedback and the filtered note (MINOR-2).
- All-invalid `find_tools` path covered by `test_find_tools_all_invalid_includes_filtered_note` (MINOR-1).
- Knowledge rollback (`aclose`/`shutdown` before clearing) and the extracted `run_knowledge_startup` production helper both land as described.

## 🔴 Blocking: the `tests.yml` block added here fails CI and omits this PR's own tests

Unit Tests is red on this branch ([run 31675193234](https://github.com/cuga-project/cuga-agent/actions/runs/31675193234/job/94368130828)), entirely inside the newly added pytest block:

1. `test_llm_advanced_params.py` (~11 failures): tests patch `ReasoningChatOpenAI` / `ChatGroq` / `ChatWatsonx` / `ReasoningChatLiteLLM` at `cuga.backend.llm.models` module scope — names the #486/#535 lazy-import refactor removed. Stale tests, never CI-green; wiring them in just exposed it.
2. `test_llm_minimax_provider.py` (3 failures): `assert '***' == 'MiniMax-M3'` — the `***` is Actions secret masking; `_get_model_name` picks up a repo-secret env var in CI. Not hermetic.
3. `tests/unit/test_shortlister_name_validation.py` — this PR's headline test file — appears **zero times** in `tests.yml`. The workflow enumerates files explicitly, so the new coverage never runs in CI while five unrelated failing files do.

**Suggested fix:** drop the five `test_llm_*` lines (separate PR to make those CI-green), keep `test_knowledge_init_failure_isolation.py`, add `test_shortlister_name_validation.py` to the same block.

## 🟡 MCP-thread rollback (open CodeRabbit item)

Verified real but marginal: the daemon `knowledge-mcp` thread starts at `main.py:755-757` and only `_warm` scheduling follows it, so the failure window is tiny, and `run_http` has no stop handle. Suggest a follow-up issue rather than growing this already-out-of-scope portion.

## On the observed firing rate (1/43 easy, 0/40 med on gpt-oss-120b)

This is not anomalous — but check how "fired" was counted.

- **Statistics:** at the observed ~1.2% rate, P(zero firings in 40 med tasks) ≈ 62%. 1-vs-0 at these suite sizes carries no signal; distinguishing 1% from 0% needs ~250 tasks per suite.
- **Channel:** the memorable hallucinations in #312 (`phone_resolve_contact_by_name` NameErrors) come from the **code agent** at invoke time. This guard sits on the **shortlister's structured output** at discovery time — #546 calls it complementary to #312. Structured output over an explicit candidate list rarely hallucinates; freeform code-gen does.
- **Model:** #312 was observed on GPT-4.1; this run is gpt-oss-120b after the #639 `json_mode` fallback — already predicted above to be a near-no-op combination.
- **Measurement caveat:** since `8e1eb836` there are three distinct signals, and they don't fire together:
  - `"Shortlister returned unrecognized tool names (attempt …)"` — logs on **any** invalid name. This is the complete signal; grep this one.
  - An actual **retry** — now happens only when the *entire* shortlist is invalid; partial hallucinations filter in one shot with no retry.
  - The `"Filtered out N unrecognized…"` note — `find_tools` output only. The `shortlist_tool_names` (bind-tools) path logs `"dropping unrecognized names after retries"` and shows nothing in traces.

  If the count came from retries or trace notes, partial hallucinations on the bind-tools path were invisible. Re-grep both run logs for `unrecognized tool names` to get the true rate. The metric that would truly validate the PR — invoke-time undefined-tool failures vs `main` — mostly lives in #312's channel, which this PR intentionally doesn't touch.

src/cuga/backend/server/main.py
```diff
@@ -188,6 +188,42 @@ def _knowledge_citations_enabled_for_app_state(app_state: "AppState" | None) ->
...
+                "failed",
+                "Knowledge subsystem failed to initialize",
+                {"error": str(e)},
+            )
+    else:
+        app_state.knowledge_engine = None
+        logger.info("Knowledge features disabled (knowledge.enabled=false)")
+        app_state.set_subsystem_status("knowledge", "disabled", "Knowledge subsystem disabled")
+
+
```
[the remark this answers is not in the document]
sami-marreed: Deferring this from #549.

`run_http` blocks in `mcp.run(...)` with no stop handle, so teardown would mean adding MCP lifecycle that this PR does not otherwise touch. The daemon thread starts at the HTTP MCP block and the only subsequent work is `_warm` scheduling — the failure window after start is tiny.

Follow-up is the right place; not growing the already-out-of-scope knowledge portion here.

sami-marreed:
`tests.yml` updated in 31e8a6d4:

- Dropped the five `test_llm_*` files (stale lazy-import patches from #486/#535, plus MiniMax secret masking).
- Kept `test_knowledge_init_failure_isolation.py`.
- Added `test_shortlister_name_validation.py` to the explicit list.

Local run: 13 passed on those two files.

MCP-thread rollback: same call as above — `run_http` has no stop handle, failure window after start is tiny. Not growing this PR for it.

Sergey-Zeltyn (approved):

PR #551, opened 2026-07-28
[work in progress]: palette as a skill + CUGA demo for deck building 

anupamamurthi:


laredo (commented):
I would prefer that all demos be handled by a separate README and code files, when we start permeating the code and docs with references to demo components we will likely confuse agents (and humans)

sami-marreed:
Closing after open-PR triage.

This isn’t the right repo for a palette skill + deck-building demo. That work should live in **cuga-skills** (skills / demos that aren’t core agent runtime), not cuga-agent.

Thanks for the WIP — please continue the palette / deck work over there.

PR #558, opened 2026-07-28
fix(cuga_graph): handle harmony control tokens when a harmony model is detected

### LINKED ISSUE #553: [Bug]: Harmony control tokens leak into the final answer via the reasoning-only fallback
### What happened

When a harmony-format model (e.g. gpt-oss) finalizes a turn with empty visible content, the reasoning-only fallback in `create_call_model_node` (`src/cuga/backend/cuga_graph/nodes/cuga_agent_core/graph/shared_nodes.py`) surfaces the raw reasoning text — including `<|channel|>`-style control tokens — as the user-facing answer. The fallback should never surface text containing harmony control tokens; motivation: users receive raw protocol tokens as their answer.

### How to reproduce

Run a cuga-lite turn where the model returns empty `content` and reasoning like `"...call getdata<|start|>assistant<|channel|>final<|message|>42"` — that string is returned to the user verbatim, tokens and all.

### Environment

CUGA main (fbb9f185), any harmony-format model (gpt-oss family).

### Checklist

### LINKED ISSUE #557: [Bug]: Stray <|...|> control tokens are not stripped from the final answer
### What happened

gpt-oss (harmony-format) models can emit stray `<|...|>` control tokens inside the final answer text, and CUGA delivers them to the user unfiltered — there is no sanitization anywhere on the final-answer path (`FinalAnswerNode.apply_citation_resolution` in `src/cuga/backend/cuga_graph/nodes/answer/final_answer.py` is the last mutation point of `final_answer` and does not strip them); motivation: users should never see raw protocol tokens in an answer. Complementary to #553, which guards a different emission point (the reasoning-only fallback).

### How to reproduce

Any turn where the model's final text ends with a control token, e.g. `"The total is 42<|return|>"` — the answer reaches the user with `<|return|>` still in it.

### Environment

CUGA main (fbb9f185), any harmony-format model (gpt-oss family).

### Checklist

segevshlomovIBM:
## Bug fix

Fixes #557
Fixes #553 (supersedes #554)

### Summary

gpt-oss speaks the harmony protocol, and its framing tokens (`<|channel|>`, `<|return|>`, …) can reach the user. Per @sami-marreed's review this is now **one PR** covering both emission points, gated on detecting a harmony model.

`utils/harmony.py` owns the concern — vocabulary, gate, detection, stripping — and two call sites use it:

| Emission point | Behaviour |
|---|---|
| `shared_nodes.call_model` | won't promote token-bearing **reasoning** into an answer when visible content is empty (was #554) |
| `final_answer` | strips framing from the delivered answer and its chat-history copy |

**Vocabulary from the library, not by hand.** `harmony_special_tokens()` loads `openai-harmony`'s `HARMONY_GPT_OSS` encoding — 1091 tokens vs the 8 originally hardcoded — so it tracks upstream. Any `<|...|>`-shaped run is a candidate; only members of that set are touched, so `<|custom|>` / `<|im_end|>` survive. Falls back to the eight framing tokens if the wheel is missing.

**Detected, not blanket-applied.** `advanced_features.strip_harmony_control_tokens`: `"auto"` (default) acts only when the final-answer model is gpt-oss — every other provider's output is byte-identical; `true`/`false` force it. The gate is checked before the encoding loads, so non-harmony runs pay nothing.

```
'The total is 42<|return|>'              -> 'The total is 42'
'keep <|custom|> and <|im_end|>'         -> unchanged
'<|message|>    def f():\n        pass'  -> indentation preserved
gpt-4o (any text)                        -> byte-identical
```

**Why it leaks at all** (@sami-marreed asked): providers normalize the visible `content` channel, but CUGA reads `reasoning_content` off the raw response dict via `ReasoningChatOpenAI`, because LangChain drops it — so the un-normalized reasoning channel is the actual source.

**Not included:** using harmony to *recover* the answer rather than discard it (in `…<|channel|>final<|message|>42` the answer is `42`). That's a behaviour change; happy to open a follow-up.

### Testing
- [x] Verified fix locally; tests pass — 190 across `test_final_answer_citations.py` and `cuga_agent_core/tests/`, covering both emission points, the gate (auto per model, forced on/off, non-harmony untouched), library-sourced vocabulary, and indentation preservation. Ruff clean.


src/cuga/backend/cuga_graph/nodes/answer/final_answer.py
```diff
@@ -102,6 +102,10 @@ def apply_citation_resolution(state) -> None:
             )
 
             text = state.final_answer or ""
+            if "<|" in text:
+                # harmony-format models (gpt-oss) can leak <|...|> control tokens
+                text = _re.sub(r"<\|[^>]*?\|>", "", text).strip()
+                state.final_answer = text
```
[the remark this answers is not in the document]
segevshlomovIBM: Fixed: extracted a shared _strip_control_tokens helper and applied it to the chat-history copy in _generate_final_answer; added test_chat_history_copy_is_sanitized covering the chat-enabled branch.

tests/unit/test_final_answer_citations.py
```diff
@@ -50,6 +50,22 @@ def test_hallucinated_marker_stripped_even_without_ledger():
...
+    assert state.final_answer == "The total is 42"
+
+
+def test_control_tokens_stripped_from_cited_answer():
+    _seed_ledger()
+    state = _state("answer [s1] done<|channel|>")
+    FinalAnswerNode.apply_citation_resolution(state)
+    assert "<|" not in state.final_answer
+    assert state.final_answer == "answer [1] done"
+    assert state.sources[0]["filename"] == "f.pdf"
```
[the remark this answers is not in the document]
segevshlomovIBM: Already covered: this file declares module-level `pytestmark = pytest.mark.unit` (line 7), which applies to the new tests.

haroldship:
@segevshlomovIBM I'm worried that a regex based filter will remove legitimate parts of the final answer for software developers, ai researchers, data scientists, etc. I don't know if these are cuga customers, but this being an open source project, can we be sure?

What about passing in stop sequences to the LLM?  Like openai-harmony `stop_tokens()` etc. 
Also, is this code robust to other api's like watsonx, groq, anthropic, etc.?

haroldship:
ChatOpenAI, ReasoningChatOpenAI, ChatGroq, ChatWatsonx all support stop sequences

segevshlomovIBM:
Good points — addressed the false-positive risk in the latest commit: the filter now matches only the **closed harmony protocol vocabulary** (`<|start|>, <|end|>, <|message|>, <|channel|>, <|constrain|>, <|return|>, <|call|>, <|endoftext|>`) instead of a generic `<|...|>` wildcard, and non-matching text is returned untouched (added `test_legitimate_special_token_text_is_preserved` — an answer discussing `<|custom|>` or `<|im_end|>` passes through verbatim). Those eight tokens are gpt-oss protocol framing with no legitimate un-escaped occurrence in a final answer, so the residual false-positive surface is an answer that literally quotes harmony's own tokens.

On stop sequences: I considered them, but they solve a different problem. A stop sequence **truncates at the first match**, and the leak we observed arrives mid-payload — e.g. `"...call getdata<|start|>assistant<|channel|>final<|message|>42"` — where stopping at `<|start|>` would discard the actual answer (`42`) that follows the tokens, while stripping preserves it. The tokens also arrive via the *reasoning* channel already embedded in the returned payload (a serving-stack artifact on WatsonX gpt-oss), which generation-time stops can't reach. Happy to additionally wire harmony `stop_tokens()` into the client layer as a belt-and-braces follow-up if you'd like.

On robustness across APIs: the strip lives at the final-answer egress (`apply_citation_resolution`), so it's provider-agnostic; with the narrowed vocabulary it's a no-op for providers that never emit harmony framing (Anthropic, Groq-native, etc.).

haroldship (dismissed):
Thanks for making the change. Looks good now.

sami-marreed:
I recommend using https://github.com/openai/harmony over parsing every possible tag, but with careful detection or feature flag

BTW where did you get this problem since;

```
If you are not using gpt-oss directly but through an API or a provider like HuggingFace, Ollama, or vLLM, you will not have to be concerned about this as your inference solution will handle the formatting
```

segevshlomovIBM:
Good question — the quote holds for the **content** channel, but we go around it.

`models.py` defines `ReasoningChatOpenAI` specifically to read `reasoning_content` off the **raw response dict**, because LangChain drops it. So the provider normalizes `content`; we deliberately surface the un-normalized reasoning channel — and #553/#554's leak is exactly there, in the fallback that promotes reasoning to a user-facing answer when content is empty.

Being straight about this PR though: the reasoning case (#554) is the one we actually observed. A stray token in the *final answer proper* — what this PR guards — I can't point to a captured trace for. So I'm happy to gate it behind harmony-model detection / a flag as you suggest, rather than applying it to every provider's output.

On `openai/harmony`: agreed it's the right library, but I'd argue not for stripping — it's an encoder/parser for whole conversations, heavy for "does this string contain stray framing" (8 literals). Where it genuinely wins is *recovery*: in `...getdata<|start|>assistant<|channel|>final<|message|>42` the real answer is `42`, and today we discard it. Parsing the `final` channel would return it. I'd like to do that as a follow-up — better behaviour than either suppressing or stripping.

Want me to add the detection gate here, or hold this PR until the harmony-parsing approach is agreed?

segevshlomovIBM:
@sami-marreed — added the detection/flag you asked for.

`advanced_features.strip_harmony_control_tokens`:
- **`"auto"` (default)** — strips only when the final-answer model is a gpt-oss/harmony model. Every other provider's output now passes through **byte-identical**; the filter can't touch it.
- **`true` / `false`** — force it on or off (e.g. a provider that leaks framing under a model name we don't recognise, or to disable outright).

Verified end-to-end:
```
gpt-4o                -> 'The total is 42<|return|>'   (untouched)
openai/gpt-oss-120b   -> 'The total is 42'             (stripped)
```

Four tests cover the gate itself (auto resolves per model, forced values win, and a non-harmony run leaves tokens untouched), and the existing strip tests now pin the flag so they assert the filter rather than whichever model the ambient settings name. 16 passed, ruff clean.

On `openai/harmony` itself: I'd still argue against it *for stripping* — it's an encoder/parser for whole conversations, heavy for checking 8 literals at the egress. Where it genuinely wins is **recovery**: in `...getdata<|start|>assistant<|channel|>final<|message|>42` the real answer is `42`, and both this PR and #554 currently discard it. Parsing the `final` channel would return it instead. I'd like to do that as a separate PR rather than widen this one — happy to open the issue if you agree it's worth doing.

src/cuga/backend/cuga_graph/nodes/answer/final_answer.py
```diff
@@ -22,6 +23,50 @@
...
+        return False
+    return "gpt-oss" in model_name.lower()
+
+
+def _strip_control_tokens(text: str) -> str:
+    if "<|" not in (text or "") or not _CONTROL_TOKEN_RE.search(text):
+        return text
+    if not _harmony_stripping_enabled():
+        return text
+    return _CONTROL_TOKEN_RE.sub("", text).strip()
```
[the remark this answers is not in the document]
segevshlomovIBM: Fixed — you're right, and the failure case is real:

```
'<|message|>    def foo():\n        return 1'
  before -> 'def foo():\n        return 1'      ← first line's indent lost
  after  -> '    def foo():\n        return 1'  ← preserved
```

Dropped the trailing `.strip()` so the filter removes the eight tokens and nothing else, matching the PR's stated contract. Added `test_indentation_after_a_token_is_preserved` as the regression test you asked for. 17 passed, ruff clean.

segevshlomovIBM:
On the outside-diff `[agent_spawn]` finding (flagged Critical) — that one doesn't reproduce; skipping it:

```
$ grep -n '^\[agent_spawn\]' src/cuga/settings.toml
83:[agent_spawn]                     # exactly one occurrence

$ python -c "import tomllib; tomllib.load(open('src/cuga/settings.toml','rb'))"
  parses OK
```

There's a single declaration, the file parses cleanly, and this PR's `settings.toml` diff only adds `strip_harmony_control_tokens` under `[advanced_features]` (line ~68) — it doesn't touch line 83 or any table header. If TOML genuinely had a duplicate table the whole app would fail to boot, which it doesn't.

The inline whitespace finding was valid and is fixed in the commit above.

segevshlomovIBM:
@sami-marreed — adopted your proposal; the tag list is gone.

`_harmony_special_tokens()` now loads the `HARMONY_GPT_OSS` encoding from `openai-harmony` and uses its `special_tokens_set` (**1091 tokens**, vs the 8 I'd hardcoded), so the vocabulary tracks upstream rather than drifting. Any `<|...|>`-shaped run is a candidate; only members of that set are removed:

```
'The total is 42<|return|>'              -> 'The total is 42'
'a<|channel|>b<|message|>c'              -> 'abc'
'keep <|custom|> and <|im_end|>'         -> unchanged   (not harmony specials)
'<|message|>    def f():\n        pass'  -> '    def f():\n        pass'   (indent kept)
```

Added `openai-harmony>=0.0.4` as a dependency (lockfile +25 lines, one package, nothing else moved). It's loaded **lazily and cached behind the model gate**, so a non-gpt-oss run never builds the encoding — no startup or per-answer cost where it doesn't apply. If the wheel is ever missing it falls back to the eight framing tokens, degrading rather than disabling.

The detection/flag from your earlier note is unchanged: `strip_harmony_control_tokens="auto"` strips only for harmony models, `true`/`false` force it. 18 tests pass, including one pinning that the vocabulary is library-sourced rather than hand-written. I've updated #557 with the revised approach.

One thing I'd still like your view on, separately from this PR: the library also lets us **recover** rather than just strip. In `...<|channel|>final<|message|>42` the answer is `42`, and today both this PR and #554 throw it away. Parsing the `final` channel would return it. Worth a follow-up issue?

segevshlomovIBM:
## Completeness pass: every surface that delivers model text

Following up on *"make sure we address all concerns and cover all cases"* — I audited each path where model output can reach a user rather than just the two known emission points.

| Surface | Covered by | Status |
|---|---|---|
| `state.final_answer`, all terminal paths (chat agent, task analyzer, CugaLite, supervisor, HITL fallback, generated answer) | `apply_citation_resolution` | ✅ |
| Visible chat message | explicit strip on the chat copy | ✅ |
| Reasoning surfaced as an answer | `contains_harmony_tokens` gate in `shared_nodes` | ✅ |
| Server SSE (`/api/...`), A2A runners, SDK `result.answer` | all read the post-node answer — covered transitively | ✅ |
| Token-level streaming | not a path: the graph streams at node granularity (`stream_mode="updates"`/`"values"`), so no partial framing can reach the client ahead of the strip | ✅ n/a |
| **Trajectory / analytics step** | — | ❌ → fixed |
| `state.messages` | intentionally raw — model context, not output | ✅ by design |

The one real gap was the trajectory step. Every other branch of `node_handler` collects its step *after* `apply_citation_resolution`, but `_generate_final_answer` collected *before* it, so persisted trajectories and analytics annotations were the last surface still carrying raw framing. Now sanitized the same way the chat copy is, with no reordering of the existing calls.

`test_tracked_trajectory_step_is_stripped_too` covers it, and fails on the previous commit.

Full run: 32 passed across the citation and shared-call-model suites; lint and format clean.

src/cuga/backend/cuga_graph/utils/harmony.py
```diff
@@ -0,0 +1,128 @@
...
+        return True
+    if normalized in ("false", "0", "no", "off"):
+        return False
+    # auto: harmony framing originates from the gpt-oss family only
+    try:
+        model_name = str(getattr(settings.agent.final_answer.model, "model_name", "") or "")
+    except Exception:
+        return False
+    return "gpt-oss" in model_name.lower()
+
```
sami-marreed: **Important:** `"auto"` inspects the wrong model.

This keys off `settings.agent.final_answer.model.model_name`, but CugaLite — the default `lite_mode = true` path — is built from `settings.agent.code.model` or published `llm_config.model` (`graph.py`). `configurable["llm"]` can also override the runtime model. On the CugaLite passthrough, `FinalAnswerAgent` is not re-run; `apply_citation_resolution` is the only filter, and it uses this same gate.

A published agent whose UI model is gpt-oss on top of `settings.openai.toml` / `settings.rits.toml` (final-answer still gpt-4.1 / Gemma) will leave `"The total is 42<|return|>"` in `state.final_answer`, the Answer SSE event, A2A, and the SDK. Force-on (`true`) works; `"auto"` does not.

Treat any of `agent.code.model`, `agent.final_answer.model`, and runtime `llm_config` / `configurable["llm"]` as harmony if the name contains `gpt-oss` (reuse `model_runtime_profile._normalized_model_key`). Or drop name detection and enable whenever the text matches the vocabulary — the `"<|" in text` fast-path already keeps non-harmony runs cheap.

src/cuga/backend/cuga_graph/nodes/cuga_agent_core/graph/shared_nodes.py
```diff
@@ -267,7 +268,12 @@ async def call_model(state: Any, config: RunnableConfig = None) -> Command:
         # ponytail: reasoning-only models may finalize with empty visible content
         final_answer = content
         if not (final_answer or "").strip() and reasoning:
-            final_answer = (reasoning or "").strip()
+            candidate = (reasoning or "").strip()
+            # Never surface raw harmony protocol framing as an answer. Detection
+            # is gated + vocabulary-driven (see utils.harmony), so reasoning that
+            # merely mentions other special tokens is still usable.
+            if not contains_harmony_tokens(candidate):
+                final_answer = candidate
```
sami-marreed, line 276: **Important:** visible `content` is assigned to `final_answer` with no strip, and the new guard only covers the reasoning fallback.

`get_event_message` (`agent_loop.py`) yields `CugaLiteSubgraph` / `CugaLiteCallback` `final_answer` as a `CodeAgent` SSE event *before* `CugaLiteCallback` → `FinalAnswerAgent` runs `apply_citation_resolution`. Default lite mode is the gpt-oss execution path, so users can see `<|return|>` / `<|channel|>` in the stream even when the terminal Answer payload is clean. The “SSE is transitive” audit misses this hop.

Sanitize (or refuse) here when assigning `final_answer` from visible `content`, not only from the reasoning fallback. `FinalAnswerNode` remains defense in depth.

src/cuga/backend/cuga_graph/nodes/cuga_agent_core/tests/graph/test_shared_call_model.py
```diff
@@ -29,6 +29,19 @@
...
+@pytest.fixture(autouse=True)
+def _force_harmony_handling(monkeypatch):
+    monkeypatch.setattr(
+        "cuga.backend.cuga_graph.nodes.cuga_agent_core.graph.shared_nodes.contains_harmony_tokens",
+        lambda text: (
+            "<|" in (text or "")
+            and any(t in text for t in ("<|start|>", "<|end|>", "<|message|>", "<|channel|>", "<|return|>"))
+        ),
+    )
+
```
sami-marreed: **Important:** this autouse patch means `call_model` tests never exercise the real detector or gate.

It replaces `contains_harmony_tokens` with a 5-token substring check that ignores `harmony_handling_enabled()` and the `openai-harmony` vocabulary (`<|call|>` / `<|constrain|>` are missing). There is also no test that `load_harmony_encoding` is not called when the gate is closed — a stated requirement.

A regression that loads the encoding on gpt-4o, or that fails to refuse `<|call|>`, would not fail this file. Gate tests in `test_final_answer_citations.py` do not cover this import site.

Keep a force-on fixture if needed, but add one test that uses the real `contains_harmony_tokens` with `strip_harmony_control_tokens="auto"` and a non-gpt-oss `final_answer` model, plus a mock asserting `load_harmony_encoding` is not called.

pyproject.toml
```diff
@@ -32,6 +32,9 @@ dependencies = [
     "typer-slim>=0.15.3", # explicit: docling pulls typer meta-package which needs typer-slim
     # <4: fastmcp 4.0.0 is currently beta. Lift to >=4.0.0 once a stable release exists.
     "fastmcp>=3.2.0,<4",
+    # Authoritative harmony special-token vocabulary for the final-answer filter
+    # (see final_answer._harmony_special_tokens); avoids hand-maintaining the list.
```
sami-marreed: **Minor:** this comment still points at `final_answer._harmony_special_tokens`; that helper now lives in `utils.harmony` (`harmony_special_tokens`). The next reader will grep the wrong module.

src/cuga/backend/cuga_graph/utils/harmony.py
```diff
@@ -0,0 +1,128 @@
...
+    never pay for building the encoding.
+    """
+    try:
+        from openai_harmony import HarmonyEncodingName, load_harmony_encoding
+
+        encoding = load_harmony_encoding(HarmonyEncodingName.HARMONY_GPT_OSS)
+        return frozenset(encoding.special_tokens_set)
+    except Exception as e:  # pragma: no cover - defensive
+        logger.debug("openai-harmony unavailable; using built-in token list: {}", e)
+        return _FALLBACK_CONTROL_TOKENS
```
sami-marreed, line 76: **Minor:** the 8-token fallback is untested. `test_vocabulary_comes_from_openai_harmony` requires the wheel, and CI always has the dep, so a broken fallback would only show up in a stripped install.

A unit test that patches `load_harmony_encoding` to raise (or `openai_harmony` import to fail) and asserts `_FALLBACK_CONTROL_TOKENS` is used would lock this in.

src/cuga/backend/cuga_graph/nodes/answer/final_answer.py
```diff
@@ -225,13 +229,26 @@ async def _generate_final_answer(state: AgentState, agent: FinalAnswerAgent, nam
...
+        # of node_handler tracks an already-stripped answer, so without this the
+        # trajectory is the one surface still carrying raw framing. The raw text
+        # stays in state.messages, which is model context rather than output.
+        tracker.collect_step(
+            Step(
+                name=name,
+                data=final_answer_output.model_copy(
+                    update={"final_answer": strip_harmony_tokens(final_answer_output.final_answer)}
+                ).model_dump_json(),
+            )
```
sami-marreed: **Minor:** the same string is stripped three times (chat copy, trajectory, then `apply_citation_resolution`). Harmless with the cache; a local `sanitized = strip_harmony_tokens(...)` would make the “this copy is delivered first” comment easier to follow.

src/cuga/settings.toml
```diff
@@ -65,6 +65,10 @@ cuga_lite_nl_auto_continue = false  # When model returns NL with no code, LLM-cl
 path_segment_index = 1  # Which path segment to use for operation naming (1 = first, 2 = second, 3 = third)
 force_autonomous_mode = false
 tool_call_timeout = 30  # Timeout in seconds for tool/API calls (sandbox operations). If exceeded, raises TimeoutError
+# Strip harmony protocol control tokens (<|channel|>, <|return|>, …) from the final answer.
+# "auto" = only when the final-answer model is a gpt-oss/harmony model (other providers untouched);
+# true/false force it on/off. Env: DYNACONF_ADVANCED_FEATURES__STRIP_HARMONY_CONTROL_TOKENS
+strip_harmony_control_tokens = "auto"
```
sami-marreed, line 72: **Minor:** other `advanced_features` keys have a dynaconf `Validator` with a default (e.g. `cuga_lite_nl_auto_continue` in `config.py`). Missing overlay files still work via `getattr(..., "auto")`, but a typo in the env var fails closed (`Exception` → `False` in `harmony_handling_enabled`) with only a silent skip.

Add `Validator("advanced_features.strip_harmony_control_tokens", default="auto")`.

sami-marreed (changes requested):
## Review summary

The filter, vocabulary source, gate-off byte-identity, whitespace behavior, and the two planned call sites are implemented and tested. Default gpt-oss TOML configs will ship clean final answers. Two leak paths remain on the lite-mode / UI-selected-model route this bug exists to fix — those should land before merge.

**Ready to merge?** With fixes

### Strengths
- One module owns vocabulary, gating, detection, and stripping; `openai-harmony` loads only after the gate.
- Candidate `<|...|>` runs are removed only if they are in the official set; surrounding whitespace is preserved.
- `shared_nodes` refuses token-bearing reasoning (#553); `FinalAnswerNode` strips delivered text, chat-history copy, and trajectory (#557).
- CI is green. Lazy-import rules are respected.

### Important (should fix)
1. **Auto-gate inspects the wrong model** — `"auto"` keys off `agent.final_answer.model`, not the model that actually emitted the text (`agent.code.model` / published `llm_config` / `configurable["llm"]`). See inline on `harmony.py`.
2. **CugaLite streams the unsanitized answer** — `call_model` assigns `final_answer = content` with no strip; `get_event_message` yields that as a `CodeAgent` SSE event before `FinalAnswerAgent` sanitizes. See inline on `shared_nodes.py`.
3. **`call_model` tests never exercise the real detector or gate** — autouse patch replaces `contains_harmony_tokens` with a 5-token substring check. See inline on `test_shared_call_model.py`.

### Minor
- Stale `final_answer._harmony_special_tokens` comment in `pyproject.toml`.
- No dynaconf `Validator` for `strip_harmony_control_tokens` (typos fail closed with a silent skip).
- 8-token fallback path in `harmony_special_tokens()` is untested.
- `_generate_final_answer` strips the same string three times; a local `sanitized = ...` would make the “delivered first” comment easier to follow.

### Out of scope (agreed)
Recovering the `final` channel via harmony parsing (`…<|channel|>final<|message|>42` → `42`) is correctly deferred. Document the strip-vs-refuse UX split if that follow-up is opened.

src/cuga/backend/cuga_graph/utils/harmony.py
segevshlomovIBM: Correct, and the fix is to stop inspecting model names entirely.

You're right that `auto` read the wrong key — CugaLite resolves `agent.code.model` or a published `llm_config`, so the common gpt-oss setup was never handled. Rather than enumerate the four sources, I took your second option: detection now follows the **text**. A name can be wrong (proxy, `rits/` prefix, `configurable["llm"]`); the framing can't. The `"<|" not in text` fast path keeps non-harmony runs free and the vocabulary check still lets `<|custom|>` through.

The setting survives as a kill switch only. It also now fails **open** on unreadable settings — a safety filter shouldn't disable itself over a typo'd env var (that's your `settings.toml` point too).

src/cuga/backend/cuga_graph/nodes/cuga_agent_core/graph/shared_nodes.py
segevshlomovIBM: Good catch — this was a real leak and my 'SSE is transitive' audit was wrong. I traced the terminal path and never checked the intermediate emit.

Rather than add a strip here, handling moved to `adapter.normalize_response` — the one place both graphs decode a reply. `content` is sanitized there, so `final_answer`, `state.messages`, the streamed `CodeAgent` event, the chat copy and the trajectory step are all clean at source, including this hop.

`reasoning` is deliberately left raw so the guard below can still tell framing from a real answer when visible content is empty.

`test_harmony_decode_boundary.py` covers both adapters and fails on the previous commit.

src/cuga/backend/cuga_graph/nodes/cuga_agent_core/tests/graph/test_shared_call_model.py
segevshlomovIBM: Agreed — the fixture was hiding the thing under test. Removed, along with the matching force-on fixture in `test_final_answer_citations.py`; both suites now run the real detector and vocabulary, and they pass without patching.

Added the two you asked for: one that the encoding is never built when the feature is off, and one that simulates a missing `openai-harmony` wheel and asserts the fallback list is used (your other minor).

pyproject.toml
segevshlomovIBM: Fixed — now points at `cuga_graph.utils.harmony.harmony_special_tokens`.

src/cuga/backend/cuga_graph/utils/harmony.py
segevshlomovIBM: Added `test_vocabulary_falls_back_when_openai_harmony_is_missing`: patches the import to raise, asserts `_FALLBACK_CONTROL_TOKENS` is used and that it still contains the framing that actually leaks (`<|return|>`, `<|channel|>`, `<|call|>`, `<|constrain|>`).

src/cuga/backend/cuga_graph/nodes/answer/final_answer.py
segevshlomovIBM: Resolved by the restructure rather than a local variable — the chat and trajectory strips are gone. Both copies come from `content`, which is sanitized at the decode boundary, so there's one strip left in `apply_citation_resolution` as defence in depth for answers assembled elsewhere.

src/cuga/settings.toml
segevshlomovIBM: Added `Validator("advanced_features.strip_harmony_control_tokens", default="auto")`. And the fail-closed behaviour you flagged is gone: unreadable settings now return True, so a typo can't silently switch off the filter.

segevshlomovIBM:
## All seven addressed — the two Important ones shared a root cause

Both came from the same mistake: sanitizing at **display surfaces**, and gating on a **model name**. Fixed at the source rather than patched.

### Handling moved to the decode boundary

`adapter.normalize_response` is the one point both graphs decode a model reply through (`shared_nodes.py:203`), and there are exactly two implementations. `content` is sanitized there, so everything downstream is clean at source: `final_answer`, `state.messages`, the streamed `CodeAgent` event, the chat copy, the trajectory step — **including the SSE hop you found**.

That hop is the argument for the move. I audited the surfaces, declared it complete, and missed one. Per-surface stripping doesn't converge — a new surface is a new leak. `reasoning` stays raw so the `shared_nodes` guard can still tell framing from a real answer when visible content is empty.

### The gate no longer looks at model names

You were right that `auto` read the wrong key. I took your second option rather than enumerating the four sources: **detection follows the text**. A name can be wrong — proxy, `rits/` prefix, published `llm_config`, `configurable["llm"]` — the framing can't. The `"<|" not in text` fast path keeps non-harmony runs free, and the vocabulary check still lets `<|custom|>` through.

The setting remains as a kill switch, and now fails **open** on unreadable settings: a filter whose job is keeping protocol framing away from users shouldn't switch itself off over a typo'd env var.

### Tests

Both autouse fixtures that masked the real detector are gone — the suites now exercise the real vocabulary and gate, and pass unpatched. Added:

- `test_harmony_decode_boundary.py` — both adapters; **fails on the previous commit**
- fallback vocabulary when the `openai-harmony` wheel is missing
- the encoding is never built when the feature is off
- gate is a kill switch, not a model check; unreadable settings leave the filter on

Registered in `tests.yml`. 400 passed across the citation, core, lite and supervisor suites; lint and format clean.

### Net effect

The diff is smaller than before: three display-surface strips and the model-name resolution are deleted, replaced by two lines at the boundary.

sami-marreed (approved):
Re-reviewed after the decode-boundary follow-up.

The two Important findings are addressed: detection follows the text (not `agent.final_answer.model`), and `normalize_response` sanitizes `content` at the shared decode point so the CugaLite SSE hop is covered. Minors (vocabulary source, fallback test, kill-switch Validator, redundant per-surface strips) are in as well. CI is green.

Optional follow-ups, not blocking: restore a strip on the FinalAnswerAgent chat/trajectory copies (that path does not go through the adapter), and the agreed recover-vs-strip issue for `…<|channel|>final<|message|>42`.

segevshlomovIBM:
## Found a case where this PR made things worse — fixed in c9e23b8

While checking the boundary change against realistic gpt-oss shapes I hit a problem worth surfacing rather than quietly shipping.

Plain token removal is only correct for **loose** framing. On **channel-structured** output it welds the channel names onto the text — and promotes the model's private `analysis` channel into the answer:

| input | before | after |
|---|---|---|
| `The total is 42<\|return\|>` | `The total is 42` ✅ | `The total is 42` ✅ |
| `<\|channel\|>final<\|message\|>The total is 42<\|return\|>` | `finalThe total is 42` ❌ | `The total is 42` ✅ |
| `<\|channel\|>analysis<\|message\|>Let me think<\|end\|><\|channel\|>final<\|message\|>42` | `analysisLet me thinkfinal42` ❌ | `42` ✅ |

So a filter whose job is keeping protocol framing away from users could instead have leaked the chain-of-thought.

**Fix:** when the text is channel-structured, keep only the body after the last `<|message|>` — the protocol's answer channel — then strip. Loose trailing tokens behave as before. Two new tests cover it and fail on the previous commit.

This isn't full harmony parsing. That's still worth doing (it would also retire the `contains_harmony_tokens` heuristic in `shared_nodes`), but this closes the leak without adopting the parser.

### One thing I can't verify — could you help?

**On my endpoint the filter never fires.** Instrumented against a real run: `strip_harmony_tokens` was called twice and changed the text **zero** times, because watsonx already normalizes `content` before we see it.

That means I have no local reproduction of the leak you reported, and can't confirm the fix resolves what you actually saw. Your report pointed at a published agent on `settings.openai.toml` / `settings.rits.toml` — if you can share that config (or just paste a raw `content` value that leaked), I'll verify against the real shape rather than my synthetic ones.

Worth knowing which shape yours is: the trailing-token form was always handled correctly; the channel-structured form is the one that was broken until this commit.

PR #562, opened 2026-07-28
feat(cuga_lite): record direct LangChain tool calls in ToolCallTracker

### LINKED ISSUE #561: [Feature]: Auto-record direct LangChain tool calls when track_tool_calls is enabled
### Issue type

[Feature]

### What you want and why

Tool calls made through the direct-LangChain tools path (`CugaAgent(tools=[...])` → `DirectLangChainToolsProvider`) are never recorded: registry/combined provider tools auto-record into `ToolCallTracker`, but the `prepare_node` wrapper for direct tools does not, so `invoke(track_tool_calls=True)` returns an empty `tool_calls` trace unless the user hand-decorates every tool with `@tracked_tool`; motivation: evaluation/observability consumers should get `{tool_name, arguments, result}` traces for direct tools without wrapping every tool themselves.

### How it could work

Running a task with LangChain `StructuredTool`s and `track_tool_calls=True` yields `result.tool_calls == []` today (verified: a tracking session around the exact prepare_node wrap chain records 0 calls). Fix: in `prepare_node`, wrap tools originating from `DirectLangChainToolsProvider` with a recording wrapper (applied after `make_tool_awaitable`, so sync tools record in the event-loop context where the tracker's contextvars are visible). `ToolCallTracker.record_call` is a no-op without an active session, so default behavior is unchanged; registry/combined tools keep their own recorders (no double-recording).

### Checklist

segevshlomovIBM:
## Feature

Closes #561

### Summary

Running a task with LangChain `StructuredTool`s (`CugaAgent(tools=[...])`) and `invoke(track_tool_calls=True)` yields an empty `tool_calls` trace: registry/combined provider tools auto-record into `ToolCallTracker`, but the `prepare_node` wrapper for direct tools never records, so consumers must hand-decorate every tool with `@tracked_tool` (verified empirically: a tracking session around the exact prepare_node wrap chain records 0 calls on main). This adds `make_recording_awaitable` next to `tracked_tool` and applies it in `prepare_node` only to tools originating from `DirectLangChainToolsProvider` — applied after `make_tool_awaitable`, so sync tools record in the event-loop context where the tracker's contextvars are visible. `record_call` is a no-op without an active tracking session, so default behavior is unchanged, and registry/combined tools keep their own recorders (no double-recording).

### Testing
- [x] Tested locally; tests pass — new `test_direct_tool_recording.py` (3 tests: async+sync tools recorded in order with `{tool_name, arguments, result}`, errors recorded and re-raised, behavior unchanged without a session) fails on main (feature absent), passes with the change; `cuga_lite` + `cuga_agent_core` test dirs green (520 passed) except two pre-existing `test_workspace_fs.py` failures that also fail on unpatched main; ruff clean.

## Summary by CodeRabbit

* **New Features**
  * Added direct tool-call recording for LangChain-style tools, capturing tool name, app context, arguments, results, duration, and errors.
  * Introduced per-code-block tool call counting for improved execution visibility.

* **Bug Fixes**
  * Improved detection of direct tools wrapped by tool-guarding layers.
  * Prevented duplicate recording for tools already configured for tracking while preserving error reporting and cancellation handling.


src/cuga/backend/cuga_graph/nodes/cuga_lite/tests/test_direct_tool_recording.py
```diff
@@ -0,0 +1,77 @@
...
+from cuga.backend.cuga_graph.nodes.cuga_lite.tracking.tracker import (
+    ToolCallTracker,
+    make_recording_awaitable,
+)
+
+
+def _wrap_like_prepare_node(tool_func, tool_name):
+    """The exact chain prepare_node applies to a direct tool."""
+    wrapped = make_arg_warning_callable(tool_func, None, enable=True)
+    return make_recording_awaitable(make_tool_awaitable(wrapped), tool_name, app_name="runtime_tools")
```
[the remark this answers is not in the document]
segevshlomovIBM: Deliberately keeping this at the wrap-chain level: a full prepare_tools_and_apps integration needs the registry/prompt/policy plumbing stubbed end-to-end (the existing adapter tests explicitly scope that out), which is out of proportion for this change. The provider-identification branch is 3 lines of isinstance+set logic, and the live path is exercised by sdk_core's test_invoke_with_tool_tracking, whose tool_calls now populate through this wiring. Happy to add the heavier harness in a follow-up if maintainers want it.

src/cuga/backend/cuga_graph/nodes/cuga_lite/tests/test_direct_tool_recording.py
```diff
@@ -0,0 +1,77 @@
...
+async def _get_data(x: int) -> dict:
+    return {"value": x * 2}
+
+
+def _get_data_sync(x: int) -> dict:
+    return {"value": x * 3}
+
+
+@pytest.mark.asyncio
+async def test_direct_tool_calls_recorded_in_order_with_args_and_results():
```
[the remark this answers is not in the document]
segevshlomovIBM: Fixed: added @pytest.mark.unit to all tests.

src/cuga/backend/cuga_graph/nodes/cuga_lite/tracking/tracker.py
```diff
@@ -116,6 +116,46 @@ def get_current_calls() -> List[Dict[str, Any]]:
...
+            result = await awaitable_func(*args, **kwargs)
+            return result
+        except Exception as e:
+            error_msg = str(e)
+            raise
+        finally:
+            duration_ms = (time.time() - start_time) * 1000
+            ToolCallTracker.record_call(
+                tool_name=tool_name,
+                arguments=kwargs if kwargs else {f"arg{i}": a for i, a in enumerate(args)},
```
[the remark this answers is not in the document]
segevshlomovIBM: Fixed: positional and keyword arguments are now merged in the recorded trace; test extended with a positional-call assertion.

src/cuga/backend/cuga_graph/nodes/cuga_lite/tests/test_direct_tool_recording.py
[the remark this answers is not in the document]
segevshlomovIBM: Good call — strengthening the SDK test exposed a real wiring gap: CugaAgent always wraps the base provider in ToolGuardingToolProvider, so the isinstance check never matched in the real flow and direct tools still weren't recorded. Fixed by unwrapping the decorator chain before detection, and test_invoke_with_tool_tracking now requires the recorded add_numbers call including its arguments field (verified live: fails without the unwrap, passes with it).

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/prepare_node.py
```diff
@@ -362,7 +380,12 @@ async def prepare_tools_and_apps(state: Any, config: Optional[RunnableConfig] =
                     getattr(tool, "args_schema", None),
                     enable=_warn_args,
                 )
-                adapter._tools_context[tool.name] = make_tool_awaitable(tool_func)
+                awaitable_tool = make_tool_awaitable(tool_func)
+                if tool.name in _direct_tool_names:
```
haroldship: Tools that already use `@tracked_tool` will now show up twice in the trace: once from the decorator, once from this wrapper. I confirmed it on the branch: one call to a decorated async tool produces two entries.

`@tracked_tool` is the documented way to do this today and it's exported from cuga, so anyone who followed that advice gets duplicates.

Suggest skipping the wrapper when a tool is already decorated. Have tracked_tool mark the functions it wraps, and check for the mark here:

```python
# tracker.py — on both wrappers inside tracked_tool
async_wrapper._cuga_tracked = True
sync_wrapper._cuga_tracked = True
```

```python
# prepare_node.py
if tool.name in _direct_tool_names and not getattr(tool_func, "_cuga_tracked", False):
```

One detail that bears on which rule you choose: decorated sync tools don't duplicate, because they run on a background thread where the decorator can't see the tracking session, so it records nothing. Today a decorated async tool is recorded twice and a decorated sync tool once, and this wrapper is the only thing that gets the sync case right. Both cases are worth a test.

src/cuga/backend/cuga_graph/nodes/cuga_lite/tracking/tracker.py
```diff
@@ -116,6 +116,46 @@ def get_current_calls() -> List[Dict[str, Any]]:
...
+        except Exception as e:
+            error_msg = str(e)
+            raise
+        finally:
+            duration_ms = (time.time() - start_time) * 1000
+            ToolCallTracker.record_call(
+                tool_name=tool_name,
+                arguments={**{f"arg{i}": a for i, a in enumerate(args)}, **kwargs},
+                result=result,
+                app_name=app_name,
```
haroldship, line 204: There's already a helper for this: `merge_tool_call_args` in tracking/arguments.py, and both the combined and toolguard providers use it. It maps positional args to their real parameter names, and it handles the case where generated code passes a single dict, which happens often.

With `arg{i}` naming, a call like `tool({"product_id": 1})` records as `{"arg0": {"product_id": 1}}`, while a registry tool records `{"product_id": 1}` for the same call. So the argument names in tool_calls end up depending on which provider the tool came from, which makes the trace harder to consume.

`prepare_node` has `tool.args_schema` right at the wrap site, so the parameter names are available to pass in. The `{"arg0": 7}` assertion in the new test would then read `{"x": 7}`, which is what you'd want it to say anyway.

src/cuga/sdk_core/tests/test_sdk_integration.py
```diff
@@ -184,6 +184,11 @@ async def test_invoke_with_tool_tracking(self):
         assert "15" in result.answer
         # Tool calls should be tracked when enabled
         assert isinstance(result.tool_calls, list)
+        # Direct tools are auto-recorded: the add_numbers call must appear with
+        # its arguments (proves the DirectLangChainToolsProvider wiring)
+        add_calls = [c for c in result.tool_calls if c.get("name") == "add_numbers"]
+        assert add_calls, f"expected a recorded add_numbers call, got: {result.tool_calls}"
```
haroldship: This assertion requires the model to actually call `add_numbers` for "What is 10 + 5?", which it may just answer on its own. sdk_core/tests/ runs in CI, so that's an intermittent failure waiting to happen. The existing "15" in `result.answer` check passes either way, so it won't catch the difference.

A test at the `prepare_node` level would cover this without depending on the model. test_prepare_node_weak_schema_tools.py already has the mocked-adapter setup to build on, and the same test could cover the ToolGuard unwrap and the direct-tools-only filter, which nothing exercises today without a live run.

haroldship (approved):
The diagnosis is right — direct tools were the one path with no recorder, and applying it after make_tool_awaitable is the correct placement, since sync tools run on a background thread that can't see the tracking session. Worth noting the side benefit: sync tools decorated with @tracked_tool were silently not recorded before, and now they are.

Approving. The duplicate-recording case in the first comment is the one I'd want handled before merge; the other two are quick follow-ups.

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/prepare_node.py
segevshlomovIBM: Confirmed and fixed — I reproduced exactly what you saw (decorated async → 2 records, decorated sync → 1).

Took your suggestion with the refinement your last paragraph implies: `tracked_tool` now marks both wrappers with `_cuga_tracked`, but the skip rule is `_cuga_tracked and iscoroutinefunction(tool_func)` rather than the mark alone. As you noted, the decorator's sync path records nothing (it runs in `make_tool_awaitable`'s executor thread, where the tracker's contextvars aren't visible), so skipping on the mark alone would have turned decorated-sync from 1 record into 0. Now: decorated-async → 1 (decorator's, keeps its `app_name`), decorated-sync → 1 (wrapper's), undecorated → 1 (wrapper's).

Both cases are pinned by tests (`test_decorated_async_tool_recorded_once_not_twice`, `test_decorated_sync_tool_still_recorded_once`) — the async one fails on the previous commit and passes now.

Worth noting the underlying issue outlives this PR: `@tracked_tool` on a sync tool silently records nothing today, with or without my change. Fixing that properly means propagating contextvars into the executor in `make_tool_awaitable` (`contextvars.copy_context()`), which would then let the skip rule collapse back to the mark alone. Happy to do that here or as a follow-up — it touches a helper every tool goes through, so I left it out of this PR.

src/cuga/backend/cuga_graph/nodes/cuga_lite/tracking/tracker.py
segevshlomovIBM: Good catch — switched to `merge_tool_call_args`, with `param_names` derived from `tool.args_schema` at the `prepare_node` wrap site (same `model_fields`/`__fields__` derivation `toolguard` uses). Direct-tool traces now name arguments the same way registry-tool traces do, including the single-positional-dict case you mentioned.

The new prepare_node test asserts exactly the shape you predicted: `context["echo"](7)` records `{"x": 7}` (fails on the previous commit, which produced `{"arg0": 7}`). The wrapper-level test still asserts `{"arg0": 7}` — it calls `make_recording_awaitable` directly with no schema, so it now documents the no-param-names fallback; I added a comment pointing at the schema-backed test.

src/cuga/sdk_core/tests/test_sdk_integration.py
segevshlomovIBM: Agreed — reverted. `test_sdk_integration.py` is byte-identical to main again, so no flaky model-dependent assertion and no marker question.

Replaced it with `test_prepare_node_direct_tool_recording.py`, built on the `test_prepare_node_weak_schema_tools.py` mocked-adapter setup you pointed at. Five tests, no live model:
- `test_direct_provider_detected_through_toolguard_decorator` — the ToolGuard unwrap (nothing covered this)
- `test_non_direct_provider_tools_are_not_wrapped` — the direct-tools-only filter
- `test_positional_args_recorded_under_real_parameter_names` — schema-derived arg names
- `test_decorated_async_tool_recorded_once_not_twice` / `test_decorated_sync_tool_still_recorded_once` — both duplicate cases

Two of the five fail on the previous commit and pass now; the other three are regression guards for wiring that had no coverage.

src/cuga/sdk_core/tests/test_sdk_integration.py
```diff
@@ -184,6 +184,11 @@ async def test_invoke_with_tool_tracking(self):
...
+        assert add_calls[0].get("arguments"), "recorded call must include arguments"
```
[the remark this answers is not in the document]
segevshlomovIBM: Moot now — that test is reverted to its original form (byte-identical to main), so it's no longer a changed test and needs no new marker. The coverage moved to `test_prepare_node_direct_tool_recording.py`, which carries a module-level `pytestmark = pytest.mark.unit`. Context: the assertion I'd added there depended on the model actually calling the tool, which would have flaked in CI.

segevshlomovIBM:
Both CodeRabbit findings on the recorder were valid and are now fixed in `83f428e3` — each pinned by a test that fails without the change.

**Per-block counting.** Direct tools bypass the registry/combined call paths, so they never reached `BlockToolCallCounter` and were missing from the count the executor reports as timeout evidence. `make_recording_awaitable` now increments it at entry. `test_direct_tool_counts_toward_block_tool_call_evidence` asserts two direct calls produce `current_count() == 2`.

**Cancellation.** `CancelledError` derives from `BaseException`, so it skipped `except Exception` and the `finally` block recorded a cancelled/timed-out call as a **success** (`result=None, error=None`) — exactly the silent-wrong-record class this recorder exists to avoid. Added an explicit handler that sets the error and re-raises. `test_cancelled_call_is_not_recorded_as_success` drives a hanging tool through `asyncio.wait_for` and asserts the record carries an error.

Verified fail-before/pass-after: with the tracker change stashed, both new tests fail; with it, 10 passed across the two recording test files.

sami-marreed (approved):

PR #565, opened 2026-07-28
fix(context): summarization fixes for dead setting, silent truncation, Gemma window

### LINKED ISSUE #563: fix(context_summarization): dead summarization_model setting, silent truncation fallback, stale Gemma context size
Three related problems in the context-summarization subsystem, found while investigating why summarization never engages in benchmark runs.

## 1. Doc/code divergence: `summarization_model` is dead config

`settings.toml` documents a dedicated summarizer model:

```toml
[context_summarization]
summarization_model = "gpt-4o-mini"  # Model to use for generating summaries (fast and cost-effective)
```

The setting is **never read anywhere in `src/`** (`grep -rn summarization_model src/cuga --include='*.py'` → no hits). The actual flow is `shared_nodes.py` → `apply_context_summarization(..., active_model, ...)` — summaries are generated by the **main agent model**.

The implemented behavior is arguably the better design (no cross-provider credential dependency; the payload — up to 70% of the main model's window — always fits the summarizer, which a dedicated 128k gpt-4o-mini could not guarantee for large-window main models). Recommendation: **delete the setting** (or, if a dedicated model is really wanted, wire it up with per-provider resolution plus a payload-fits-in-summarizer-window guard). Either way, doc and code should agree.

## 2. Silent hard truncation on summarizer failure

`context_summarizer.py` (~line 329): on any summarization exception the fallback is

```python
return messages[-keep_n:], {"error": str(e), "fallback": "kept recent messages only"}
```

i.e. **silent hard truncation** — the maximal state-loss mode (dropped IDs, tokens, partial results), logged only as a regular exception line. A failed summarization that silently discards all but the last N messages deserves a loud, structured signal: a WARNING/ERROR log with counts (messages dropped, tokens dropped) plus an activity-tracker event, so affected tasks can be detected in analysis instead of surfacing as unexplained behavior changes.

## 3. Stale Gemma context size vs. actual deployment (blocks the 70% trigger)

`token_counter.py` `MODEL_CONTEXT_SIZES` registers Gemma 4 at its native window:

```python
"gemma-4-31B-it": 262144,   # (+3 name variants)
```

but the RITS deployment actually serves `max_model_len = 131072` (verified via `GET .../google-gemma-4-31b-it-a100/v1/models` on 2026-07-28; it was 32k earlier in July — deployments drift). Consequence: the summarization trigger fires at `0.70 × 262144 ≈ 183k`, **beyond the 131k cliff where the endpoint 400s**, so summarization can never engage on Gemma — which, with its small window, is exactly the model class where it is supposed to help. (Before these entries were added, Gemma fell through to `DEFAULT_CONTEXT_SIZE = 131072`, which was accidentally correct for the current deployment.)

Fix: set the four Gemma 4 entries to `131072`. Longer-term, a static name→window table cannot track deployment drift; consider populating `model.profile.max_input_tokens` from the endpoint's `/v1/models` `max_model_len` at startup where available.

Sergey-Zeltyn:
## Bug fix

Fixes #563

### Summary

Three related context-summarization problems, found while investigating why summarization never engages in benchmark runs.

**Root cause**

1. `settings.toml` documented a dedicated summarizer model (`summarization_model = "gpt-4o-mini"`) that no code ever reads — summaries actually run on the main agent model, so doc and code disagreed.
2. On summarizer failure, the fallback silently kept only the last N messages (maximal state loss: dropped IDs, variables, partial results) with a single warning log; failure metrics were never stored, so the tracker/eval analysis could not detect affected tasks.
3. `MODEL_CONTEXT_SIZES` registered Gemma 4 at its native window (262,144), but the RITS vLLM deployment serves `max_model_len=131072` (probed via `GET .../v1/models`, 2026-07-28). The 70% summarization trigger computed from the native window (~183k) sat beyond the deployment's 131k context cliff, so summarization could never engage on Gemma.

**Solution**

1. Removed the dead setting; the `[context_summarization]` section comment now documents that summaries are generated by the main agent model and why (provider-safe, payload always fits the summarizing model's window).
2. Hard truncation is now loud at all three layers: ERROR logs with dropped/kept message and token counts in `context_summarizer.py` and `agent_state.py`; failure metrics enriched (`hard_truncation`, `messages_dropped`, `messages_kept`, `message_count_before`, `token_count_before`) and stored on state; `context_management_utils.py` emits a `ContextSummarizationFailure` tracker step so affected tasks are discoverable in trajectories.
3. Gemma 4 entries set to the deployment window `131072` (probe-dated comment in code), making the 70% trigger (~92k) reachable on real trajectories.

### Testing
- Extended `test_model_error_handling` to pin the enriched failure-metrics shape
- 198 token/context unit tests, 49 summarizer/utils tests, 11 shared-call-model tests pass
- `ruff check` / `ruff format --check` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved conversation summarization fallback on errors by preserving the most recent messages via hard truncation.
  * Enhanced failure reporting with dropped vs. kept message counts and pre-truncation context/token details.
  * Ensures summarization failure metrics are consistently recorded for monitoring/analytics.
* **Configuration**
  * Updated Gemma context limits to match supported deployment capacity.
  * Conversation summaries now use the primary agent model (no separate summarization model setting).
* **Tests**
  * Added unit coverage for hard-truncation behavior and failure metric tracking.


src/cuga/backend/cuga_graph/utils/token_counter.py
```diff
@@ -201,10 +201,14 @@
     # ============================================================================
     # Google Gemma 4 Models
     # ============================================================================
-    "gemma-4-31B-it": 262144,
-    "gemma-4-31b-it": 262144,
-    "google/gemma-4-31B-it": 262144,
-    "google/gemma-4-31b-it": 262144,
+    # Deployment cap, not the 262,144 native window: the RITS vLLM deployment serves
+    # max_model_len=131072 (probed via GET .../v1/models, 2026-07-28). Registering the
```
sami-marreed, line 205: actual gemma is 260k

Sergey-Zeltyn: > actual gemma is 260k

I suggest to:
- reverse this change and make default 262K
- Create an issue on the case when provider parameter is different from model default. Maybe we can introduce an optional parameter that addresses such cases.

@sami-marreed , do you agree?

sami-marreed (approved):

PR #535, opened 2026-07-22
perf(startup): lazy imports — 66% faster SDK cold-start, sub-5ms import cuga (#486)

offerakrabi:
## Feature

Partially fixes #486

### Summary

Reduces CUGA cold-start time by deferring heavy imports off the critical path. Three categories of change:

**Measurement tooling** (`src/system_tests/profiling/startup/`)
- Subprocess-isolated SDK timer (`measure_sdk_startup.py`) for true cold-start measurement
- Server readiness poller (`measure_server_startup.py`) — polls `/health/readiness` on an ephemeral port
- Ranked import hotspot finder (`import_breakdown.py`) using `-X importtime`
- N-run benchmark harness (`run_startup_bench.sh`) with median/min/max aggregation

**Import deferrals** (`sdk.py`, `server/main.py`, `openlit_init.py`, T9–T13)
- Heavy module-top imports moved into the function bodies that use them
- `openlit` (observability) no longer loads when disabled (the default)
- Graph, policy, and provider imports in the SDK no longer run on every `import`
- Policy + knowledge lifespan init parallelised (T13)

**Lazy entry-point + scope-based providers** (`__init__.py`, `models.py`, T15a/T15b)
- `cuga/__init__.py` replaced with PEP 562 `__getattr__` map — `import cuga` is now ~4 ms
- All LangChain provider libraries (Groq, Watsonx, Google GenAI, IBM) moved inside their `if platform ==` branch in `models.py` so only the configured provider is ever loaded

**Results**

| Metric | Before | After | Δ |
|--------|--------|-------|---|
| `import cuga` | 4.77 s | **~0.004 s** | −99.9% |
| SDK `import_s` | 4.77 s | **~1.64 s** | −66% |
| SDK `ready_s` | 4.84 s | **~3.58 s** | −26% |
| Server `server_ready_s` | ~9.3 s | ~9.5 s | ≈ flat |

Server ready-time and prebuilt-agent paths are left for a follow-up (issue success criteria covered SDK + server).

**Post-merge fixes** (after merging `main`)
- Restored missing lazy `HumanMessage` import in `sdk.py` `invoke()` (merge conflict artefact)
- Updated test patch targets broken by T15b lazy factory (`test_llm_http_timeout`, `test_llm_minimax_provider`, `test_llm_override` Watsonx)
- Added / tightened lazy-import rules in `AGENTS.md` so future contributors preserve the gains

### Testing
- [x] Tested locally; tests pass

## Summary by CodeRabbit

* **Performance**
[...]

PR #568, opened 2026-07-29
perf(memory): add memory profiling harness and defer BaseChatModel import in sdk.py (#486)

### LINKED ISSUE #486: [Performance]: optimize CUGA loading across SDK and server (~30s)

offerakrabi:
## Feature

Closes #486

### Summary

Follow-up to PR #535 (startup time). That work measured **time**; this branch adds a **memory profiling harness** so future work can measure and track memory usage.

**Memory profiling harness** (`src/system_tests/profiling/memory/`):
- `memlib.py` — shared RSS/USS/PSS sampling library with correct macOS/Linux `ru_maxrss` normalisation
- `measure_sdk_memory.py` — SDK checkpoints 0–5, isolated and walk modes
- `measure_server_memory.py` — single-process server cold/converged/growth
- `measure_tree_memory.py` — full `cuga start demo` process-tree footprint (USS sum)
- `import_memory_breakdown.py` — per-import RSS profiler (memory analogue of `-X importtime`)
- `tracemalloc_report.py` — Python-heap attribution, top-25 by file/line
- `memray_profile.sh` — optional native flamegraph driver (skips gracefully if memray absent)
- `aggregate_memory_results.py` + `run_memory_bench.sh` — N-run median/min/max bench
- `README.md` — full usage guide with caveats (USS not PSS on macOS, never sum RSS across processes, tracemalloc excludes native, etc.)
- `profiling = ["memray>=1.11"]` optional extra added to `pyproject.toml` (Linux only)

No production code is changed. The harness is a developer tool for diagnosing memory usage — run it manually when investigating memory regressions or optimisation opportunities.

### Testing

## Summary by CodeRabbit

* **New Features**
  * Added an optional Linux profiling toolkit for measuring SDK, server, and process-tree memory usage.
  * Added reports for imports, Python heap allocation, native memory, and Memray flamegraphs.
  * Added tools for repeated benchmarks, checkpoint comparisons, result aggregation, and summaries.
* **Documentation**
  * Added comprehensive setup, usage, output, and troubleshooting guidance for memory profiling.
* **Chores**
  * Added configuration to exclude generated profiling artifacts from source control.


src/system_tests/profiling/memory/aggregate_memory_results.py
```diff
@@ -0,0 +1,239 @@
...
+    all_runs: list[list[dict]] = []
+    for fpath in args.files:
+        try:
+            records = _load_records(Path(fpath))
+            all_runs.append(records)
+        except Exception as exc:
+            print(f"Warning: could not load {fpath}: {exc}", file=sys.stderr)
+
+    if not args.keep_first and len(all_runs) > 1:
+        all_runs = all_runs[1:]
```
sami-marreed: **Important:** this drops one *file*, not run `#1`.

`run_memory_bench.sh` appends files as `sdk-1`, `server-1`, `sdk-2`, `server-2`, … so `all_runs[1:]` removes only the first SDK file. Server run 1 stays in the median, and `n_runs` counts files rather than runs. The README “N−1 runs” story is then false.

Group by the run index in `run-<script>-<i>-<ts>.json` and drop every file with `i == 1` (unless `--keep-first`).

src/system_tests/profiling/memory/run_memory_bench.sh
```diff
@@ -0,0 +1,113 @@
...
+echo "  CUGA Memory Benchmark  (runs=${RUNS})" >&2
+echo "========================================" >&2
+
+# ---------------------------------------------------------------------------
+# T3/T5/T6 measurement scripts (may not all exist yet)
+# ---------------------------------------------------------------------------
+MEASURE_SCRIPTS=(
+    "measure_sdk_memory.py"
+    "measure_server_memory.py"
+    "measure_agent_memory.py"
```
sami-marreed: **Important:** this script does not exist, so T6 is silently skipped.

README (and the T3/T5/T6 comment above) say the bench runs `measure_tree_memory.py`. This list names `measure_agent_memory.py`, which is missing — the `[[ ! -f ]]` guard prints a warning and continues. Tree footprint is never part of the default bench.

Replace with `measure_tree_memory.py`. Do not pass `--isolated` unless that script accepts it.

src/system_tests/profiling/memory/run_memory_bench.sh
```diff
@@ -0,0 +1,113 @@
...
+        if [[ ! -f "${script}" ]]; then
+            echo "Warning: ${script} not found — skipping" >&2
+            continue
+        fi
+
+        echo "--- Run ${i}/${RUNS}: ${script} ---" >&2
+        ts="$(date -u +%Y-%m-%dT%H-%M-%SZ)"
+        out_file="results/run-${script%.py}-${i}-${ts}.json"
+
+        if ! uv run python "${script}" --isolated > "${out_file}" 2>/dev/null; then
```
sami-marreed: **Important:** `2>/dev/null` hides worker stderr, so a hang or import failure has no diagnostic.

Match the startup bench: keep stderr visible. Also add a timeout on the Python workers (see `measure_sdk_memory.py` / `import_memory_breakdown.py` / `tracemalloc_report.py` — none of those `subprocess.run` calls set `timeout=`). A blocked import or invoke stalls this loop with no output.

src/system_tests/profiling/memory/import_memory_breakdown.py
```diff
@@ -0,0 +1,349 @@
...
+        return getattr(self._real, name)
+
+    def exec_module(self, module):
+        gc.collect()
+        before_rss = psutil.Process().memory_info().rss
+        self._real.exec_module(module)
+        gc.collect()
+        after_rss = psutil.Process().memory_info().rss
+        delta_mb = max(0.0, (after_rss - before_rss) / 1024 / 1024)
+        self._tracker[self._name] = self._tracker.get(self._name, 0.0) + delta_mb
```
sami-marreed, line 66: **Important:** this records *inclusive* RSS, not self RSS.

Nested imports run inside the parent’s `exec_module` window, so `tracker[name]` already contains every child. `_compute_cumulative` (lines 151–172) then adds children again; `total_profiled_mb` can exceed process RSS. Filtering below `MIN_SELF_MB` *before* roll-up also breaks parent chains when a cheap intermediate package is dropped.

This is the memory analogue of `-X importtime` — wrong numbers here send the next lazy-import work to the wrong modules.

Fix: depth-aware exclusive deltas; compute cumulative on the unfiltered map; sum exclusive values for `total_profiled_mb`.

src/system_tests/profiling/memory/memray_profile.sh
```diff
@@ -0,0 +1,186 @@
...
+
+    # Inline worker script: import + construct CugaAgent; no real LLM calls.
+    SDK_SCRIPT=$(cat <<'PYEOF'
+import asyncio
+from cuga.sdk import CugaAgent
+from langchain_core.language_models.fake_chat_models import GenericFakeChatModel
+from langchain_core.messages import AIMessage
+
+model = GenericFakeChatModel(messages=iter([AIMessage(content="ok")]))
+agent = CugaAgent(tools=[], llm=model)
```
sami-marreed: **Important:** `CugaAgent.__init__` takes `model=`, not `llm=`. This TypeErrors before any allocation is recorded.

Also `from cuga.sdk import CugaAgent` does not match `measure_sdk_memory.py` (`from cuga import CugaAgent`), so even a fixed worker would not match the SDK checkpoint.

Below, `wait "$MEMRAY_PID" || true` (line 168) suppresses server/memray failures: if uvicorn dies during import, a partial `.bin` still produces a report and the script exits 0. Propagate a nonzero status when the target exits early.

src/system_tests/profiling/memory/measure_sdk_memory.py
```diff
@@ -0,0 +1,306 @@
...
+import memlib
+from cuga import CugaAgent
+from langchain_core.tools import tool
+
+@tool
+def echo(x: str) -> str:
+    \"\"\"Echo x back.\"\"\"
+    return x
+
+agent = CugaAgent(tools=[echo])
```
sami-marreed: **Important:** isolated `constructed` is a different workload than walk mode.

Here: `CugaAgent(tools=[echo])` (real default model path). Walk mode (lines 160–162) builds `CugaAgent(tools=[echo], model=fake_model)` and imports `GenericFakeChatModel` / `AIMessage` before sampling.

The aggregator groups by `(surface, checkpoint)`, so mixing `--isolated` and `--walk` files corrupts `sdk/constructed`. Align the two workers (same constructor args, same pre-imports).

src/system_tests/profiling/memory/tracemalloc_report.py
```diff
@@ -0,0 +1,263 @@
...
+    "\n"
+    "agent = CugaAgent(tools=[echo], model=fake_model)\n"
+    "snap_constructed, rss_constructed = _snap()\n"
+    "\n"
+    "# ── checkpoint: converged ───────────────────────────────────────────────\n"
+    "import asyncio\n"
+    "try:\n"
+    "    asyncio.run(agent.invoke('hello'))\n"
+    "except Exception:\n"
+    "    pass\n"
```
sami-marreed: **Important:** swallowing the invoke failure still emits a `converged` checkpoint.

If `agent.invoke` raises immediately, near-zero growth is indistinguishable from “invocation allocates nothing.” Record `invoke_error` in the JSON (and/or fail the worker).

Also, `_run_worker` at line 202 does `splitlines()[-1]` with no empty-stdout guard — `import_memory_breakdown.py` and `measure_sdk_memory.py` already handle that; copy the same check here.

src/system_tests/profiling/memory/measure_sdk_memory.py
```diff
@@ -0,0 +1,306 @@
...
+    Returns the parsed JSON record from the last stdout line.
+    Raises RuntimeError if the subprocess exits non-zero.
+    """
+    import os
+
+    env = os.environ.copy()
+    existing = env.get("PYTHONPATH", "")
+    env["PYTHONPATH"] = f"{memlib_dir}:{existing}" if existing else memlib_dir
+
+    result = subprocess.run(
```
sami-marreed: **Important:** no `timeout=` on worker subprocesses (same gap in `import_memory_breakdown.py` and `tracemalloc_report.py`).

If an import blocks on network, a model download, or an interactive prompt, `run_memory_bench.sh` hangs with no output. Add a shared timeout (e.g. 300s) at each `subprocess.run` site.

There are also no `@pytest.mark.unit` tests for `memlib` or the aggregator. Commit `27834cec` mentioned five fixture JSON files that were never committed; CI green only means these scripts are not collected. Tests for discard-by-run-index, `ru_maxrss` units, and “never sum RSS as the total” would have caught the aggregator and import-profiler bugs.

src/system_tests/profiling/memory/measure_server_memory.py
```diff
@@ -0,0 +1,423 @@
...
+
+# ---------------------------------------------------------------------------
+# Port helpers (copied from measure_server_startup.py — no cross-dir import)
+# ---------------------------------------------------------------------------
+
+
+def _find_free_port() -> int:
+    """Bind to port 0 and let the OS pick a free ephemeral port."""
+    with socket.socket() as s:
+        s.bind(("", 0))
```
sami-marreed: **Minor:** GHAS/CodeQL flags this as bind-all (CWE-605). It is not a real listen-all server — the socket is closed immediately and uvicorn is started with `--host 127.0.0.1`. Same pattern as `measure_server_startup.py`.

Still switch to `s.bind(("127.0.0.1", 0))` to clear the alert (as #535 did for the actual listen address).

src/system_tests/profiling/.gitignore
```diff
@@ -30,3 +30,10 @@ __pycache__/
 startup/results/*.json
 !startup/results/.gitkeep
 startup/findings.md
+
+# memory profiling
+memory/results/*.json
+!memory/results/.gitkeep
+memory/findings.md
+memory/*.bin
+memory/memray-*.html
```
sami-marreed, line 41: **Minor:** memray writes `memory/results/memray-*.{bin,html}` (`memray_profile.sh`). These two patterns only cover files directly under `memory/`, so results artifacts show up in `git status` and can be committed accidentally.

Add `memory/results/*.bin` and `memory/results/*.html` (or ignore `memory/results/*` except `.gitkeep`).

src/system_tests/profiling/memory/measure_tree_memory.py
```diff
@@ -0,0 +1,362 @@
...
+                {
+                    "name": r["name"],
+                    "label": r["label"],
+                    "pid": r["pid"],
+                    "rss_mb": round(r["rss_mb"] or 0.0, 2),
+                    "uss_mb": round(r["uss_mb"] or 0.0, 2),
+                }
+                for r in enriched
+            ],
+            "total_uss_mb": round(effective_total_mb, 2),
```
sami-marreed: **Minor:** this key is misnamed. On Linux `tree_sample` reports PSS; on the fallback path the value is an RSS sum (`metric_used` becomes `rss_sum_fallback`). A consumer that reads `total_uss_mb` without also reading `metric_used` records the wrong metric.

Rename to `total_mb` and keep `metric_used`.

sami-marreed (changes requested):
## Review

**Ready to merge: with fixes.**

The profiling architecture is sound (correct USS/PSS handling, subprocess isolation, process-group teardown, Linux-only `memray` extra). CI is green. The PR does **not** contain the production memory fix the title and body advertise, and the default bench path produces wrong “discard run #1” stats and never measures the process tree.

### Critical

**Claimed `sdk.py` BaseChatModel deferral is not in the mergeable diff.** Triple-dot vs `main` is 14 files (harness + `pyproject.toml` + `uv.lock` + gitignore) — zero lines of `sdk.py`. The fix existed in `e36d8a53` (import behind `TYPE_CHECKING`, quoted annotations). Merge `c86c9825` conflicted on `sdk.py` and took `main`’s version. HEAD still has `from langchain_core.language_models import BaseChatModel` at module top. The ~186 MB / −1,284 modules result in the PR body will not ship.

Restore `e36d8a53` on current `sdk.py`, or retitle/rebody this PR as harness-only and land the deferral separately. Do not merge while the title still claims that change.

### Important (inline)

1. Aggregator “discard run #1” drops one **file**, not one **run**.
2. Bench lists `measure_agent_memory.py` (does not exist) instead of `measure_tree_memory.py`.
3. Import profiler records inclusive RSS, then rolls it up again.
4. Memray SDK worker uses nonexistent `llm=` kwarg; failed server targets can still exit 0.
5. Isolated vs walk `constructed` checkpoints are different workloads.
6. No tests for `memlib` / aggregator (T4 fixtures were never committed).
7. Worker subprocesses have no timeout; bench hides stderr with `2>/dev/null`.
8. `tracemalloc_report.py` swallows invoke failures.

### Minor (inline)

- GHAS bind-all on `_find_free_port` is a false positive (socket closed immediately; uvicorn binds `127.0.0.1`) — still switch to `127.0.0.1` to clear the alert.
- Gitignore misses memray artifacts under `memory/results/`.
- `total_uss_mb` can hold PSS or RSS-sum.

### Strengths

`memlib.py` gets the hard parts right: macOS-vs-Linux `ru_maxrss` normalisation, USS/PSS selection, and an explicit `rss_mb_sum` that is **not** used as the tree total. Scripts follow the startup-harness stdout-last-line JSON contract. README caveats (never sum RSS, USS is a macOS lower bound, tracemalloc is Python-heap only) are accurate.

src/system_tests/profiling/memory/measure_tree_memory.py
andy-viera: Fixed: grace now comes only from the subscription selected for the effective retention policy.

sami-marreed (approved):

PR #578, opened 2026-08-03
fix(policy): keep NL conflict fallback matches above threshold

### LINKED ISSUE #577: [Bug]: Flaky LLM policy e2e — test_healthcare_family_claims_e2e_with_tools fails with Policy should match
### What happened

CI Unit Tests intermittently fail on:

`src/cuga/backend/cuga_graph/policy/tests/test_e2e_healthcare_family_claims.py::test_healthcare_family_claims_e2e_with_tools`

Assertion:

```text
AssertionError: Policy should match
assert None
 + where None = {}.get('policy_matched')
```

Expected: the healthcare family-claims graph run records `policy_matched` in result metadata.

Actual: `policy_matched` is missing/`None`. Logs show NL trigger conflict resolution failing because the LLM returned empty/invalid JSON:

```text
Error in LLM conflict resolution: Invalid json output:
...
langchain_core.exceptions.OutputParserException: Invalid json output:
... OUTPUT_PARSING_FAILURE
Falling back to first policy: 'Family Healthcare Plan Navigation'
```

This is unrelated to recent set/JSON serialization work (e.g. PR #575): the same failure has appeared on `main` and on unrelated PR runs. It blocks the Unit Tests job and forces re-runs.

Related prior flake pattern: #301 (another LLM-backed policy e2e).

### How to reproduce

1. Run the Unit Tests workflow policy/SDK integration step (or locally with the same LLM-backed policy suite):

```bash
uv run pytest src/cuga/backend/cuga_graph/policy/tests/test_e2e_healthcare_family_claims.py::test_healthcare_family_claims_e2e_with_tools -q
```

2. Re-run until the LLM conflict-resolution call returns empty/invalid JSON (intermittent).

3. Observe assertion failure at `test_e2e_healthcare_family_claims.py:397` even when the agent falls back to the first policy.

Example failing CI jobs:
- https://github.com/cuga-project/cuga-agent/actions/runs/30806860877/job/91664131704 (PR #575)
- https://github.com/cuga-project/cuga-agent/actions/runs/30805985229/job/91661353655 (PR #575)
- Earlier `main` Unit Tests failure on 2026-08-02 (`test_healthcare_family_claims_e2e_with_tools`)

### Environment

- OS: GitHub Actions `ubuntu` runners (CI)
- Python: 3.12 (Unit Tests job)
- Repo: `cuga-project/cuga-agent`
- Suite: policy e2e with live/LLM-backed conflict resolution (`_resolve_nl_trigger_conflicts` in `policy/agent.py`)
- Marker context: runs in Unit Tests workflow “Run policy and SDK integration tests” step

From the failing job:

```text
ERROR | cuga.backend.cuga_graph.policy.agent:_resolve_nl_trigger_conflicts:672 -
❌ Error in LLM conflict resolution: Invalid json output:
For troubleshooting, visit: https://docs.langchain.com/oss/python/langchain/errors/OUTPUT_PARSING_FAILURE
...
Falling back to first policy: 'Family Healthcare Plan Navigation'
...
FAILED ...::test_healthcare_family_claims_e2e_with_tools - AssertionError: Policy should match
============ 1 failed, 56 passed, 23 warnings ============
```

Likely hardening directions (not prescribing a fix):
- Treat fallback policy selection as a successful match for metadata (`policy_matched`), or
- Retry/structured-output the conflict-resolution LLM call, or
- Mark/quarantine this LLM-backed case so Unit Tests are not gated on non-deterministic JSON parse failures.

### Checklist

sami-marreed:
## Bug fix

Fixes #577

### Summary
NL conflict-resolution fallback after LLM JSON parse failures used hardcoded confidence `0.5`. Policies with threshold `0.7` (e.g. healthcare family-claims playbook) were then discarded by `_evaluate_natural_language_policies`, so CI logged “Falling back to first policy” but `policy_matched` stayed missing.

- Fall back with confidence that clears the selected policy’s NL threshold
- Retry structured-output once before falling back
- Add unit coverage for fallback + retry paths

### Testing
  - `uv run pytest src/cuga/backend/cuga_graph/policy/tests/test_nl_trigger_conflict_fallback.py test_keyword_operator.py -q`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved policy conflict resolution with reliable structured-response retries.
  * Fallback decisions now use configured natural-language trigger thresholds for more accurate confidence handling.
  * Preserved valid policy matches during fallback evaluation, including when multiple candidates are available.
  * Improved handling when no language model is available or a response cannot be parsed.

* **Tests**
  * Added coverage for threshold calculations, candidate filtering, fallback selection, retry behavior, and successful structured responses.

### DIFF
--- a/src/cuga/backend/cuga_graph/policy/agent.py
+++ b/src/cuga/backend/cuga_graph/policy/agent.py
@@ -6,9 +6,11 @@ from typing import Any, Dict, List, Optional
 import numpy as np
 from langchain_core.language_models import BaseChatModel
 from langchain_core.messages import HumanMessage, SystemMessage
+from langchain_core.prompts import ChatPromptTemplate
 from loguru import logger
 from pydantic import BaseModel, Field
 
+from cuga.backend.cuga_graph.nodes.shared.base_agent import BaseAgent
 from cuga.backend.cuga_graph.policy.models import (
     AlwaysTrigger,
     AppTrigger,
@@ -626,40 +628,22 @@ Provide:
                 policy_types=policy_types,
             )
 
-            messages = [
-                SystemMessage(content=system_prompt),
-                HumanMessage(content=user_prompt),
-            ]
-
             logger.debug("  - Sending request to LLM for conflict resolution")
             logger.debug(f"    User prompt length: {len(user_prompt)} chars")
             logger.debug(f"    Number of policies in prompt: {len(policies_with_nl_triggers)}")
 
-            # Use structured output for reliable JSON parsing.
-            # method="json_schema" (matching OutputFormatter in enactment.py) is enforced
-            # server-side by IBM Watsonx. The default (function_calling) returns None on
-            # gpt-oss-120b when it emits no tool call, which raised AttributeError below and
-            # silently dropped the policy match.
-            # Retry once on parse failures — intermittent empty/invalid JSON is a known flake.
-            structured_llm = self.llm.with_structured_output(PolicyConflictResolution, method="json_schema")
-            result: Optional[PolicyConflictResolution] = None
-            last_error: Optional[Exception] = None
-            for attempt in range(2):
-                try:
-                    result = await structured_llm.ainvoke(messages)
-                    break
-                except Exception as e:
-                    last_error = e
-                    if attempt == 0:
-                        logger.warning(
-                            f"⚠️  LLM conflict resolution attempt {attempt + 1} failed "
-                            f"({type(e).__name__}: {e}); retrying once"
-                        )
-                        continue
-                    raise
-
-            if result is None:
-                raise last_error or RuntimeError("LLM conflict resolution returned no result")
+            # Reuse BaseAgent structured-output chain: json_schema with json_mode
+            # fallback on missing parsed/refusal, plus with_retry(stop_after_attempt=3).
+            prompt_template = ChatPromptTemplate.from_messages(
+                [
+                    ("system", system_prompt),
+                    ("human", "{user_prompt}"),
+                ]
+            )
+            chain = BaseAgent.get_chain(prompt_template, self.llm, PolicyConflictResolution)
+            result = await chain.ainvoke({"user_prompt": user_prompt})
+            if isinstance(result, dict):
+                result = PolicyConflictResolution(**result)
 
             logger.debug("  - Received structured LLM response:")
             logger.debug(f"    - matched_policy_index: {result.matched_policy_index}")
--- a/src/cuga/backend/cuga_graph/policy/agent.py
+++ b/src/cuga/backend/cuga_graph/policy/agent.py
@@ -518,16 +518,20 @@ Which policy (if any) best matches the context above?"""
 Which policy (if any) best matches the user's intent?"""
 
     @staticmethod
-    def _fallback_confidence(nl_triggers: List[NaturalLanguageTrigger]) -> float:
-        """Confidence for non-LLM fallback selection.
+    def _min_nl_threshold(nl_triggers: List[NaturalLanguageTrigger]) -> float:
+        """Minimum NL threshold across triggers (most strict gate)."""
+        return min(t.threshold for t in nl_triggers)
 
-        Must be >= the policy's NL threshold, otherwise
-        ``_evaluate_natural_language_policies`` discards the fallback match.
+    @staticmethod
+    def _fallback_confidence(nl_triggers: List[NaturalLanguageTrigger]) -> float:
+        """Gate-clearing confidence for non-LLM fallback selection.
+
+        Not an estimated match-quality score: exactly the policy's minimum NL
+        threshold so ``_evaluate_natural_language_policies`` keeps the match.
+        Lower thresholds therefore yield lower fallback confidence and can lose
+        to higher-confidence keyword matches in ``match_policy`` — intentional.
         """
-        thresholds = [t.threshold for t in nl_triggers if hasattr(t, "threshold")]
-        if thresholds:
-            return min(thresholds)
-        return 0.5
+        return PolicyAgent._min_nl_threshold(nl_triggers)
 
     async def _resolve_nl_trigger_conflicts(
         self,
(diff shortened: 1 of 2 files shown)


offerakrabi:
## PR Review: #578 — fix(policy): keep NL conflict fallback matches above threshold

I'm not sure I agree with 1 , the suggested fix is cleaner but in my opinion the current implementation is functionally correct.

 **Merge with fixes.** The confidence fix is correct and the `BaseAgent.get_chain` reuse is sound. One structural concern to resolve before merge.
Approve once #1 is addressed (either move the import or accept it with a tracking note).

|---|---|---|---|---|
| 1 | Minor | `policy/` is first module outside `nodes/` to import `BaseAgent` | `policy/agent.py:13` | Sole cross-boundary import of `nodes/shared/base_agent`; if that file is split per the 1000-line rule, `PolicyAgent`'s import breaks at runtime with no static warning. |
| 2 | Nice to have | `@pytest.mark.unit` label inconsistent with real storage use | `tests/test_nl_trigger_conflict_fallback.py:44,105` | Tests call `PolicyStorage.initialize_async()` → real SQLite + sqlite-vec; all other tests in this directory do the same but are unmarked or `@pytest.mark.asyncio` only — the `unit` marker overstates isolation. |
| 3 | Nice to have | No-LLM early-exit fallback path not covered | `policy/agent.py:567-582` | The `if not self.llm or not query_text:` branch also uses the fixed `_fallback_confidence` but has no test; a regression there requires manual inspection. |

## Findings

### 1 — `policy/` is first module outside `nodes/` to import `BaseAgent` (Minor)
`src/cuga/backend/cuga_graph/policy/agent.py:13`

`PolicyAgent` now imports `BaseAgent` from `cuga.backend.cuga_graph.nodes.shared.base_agent`.
This is the only file in the entire codebase outside the `nodes/` subtree that does so —
38 other consumers are all inside `nodes/`. `AGENTS.md`'s under-1000-line rule actively
encourages splitting `base_agent.py` (currently 230 lines, but the rule applies to the
directory's files broadly). If `get_chain` is extracted to a sub-module in a future
refactor, this import silently breaks the next time `_resolve_nl_trigger_conflicts` is
called in production.

The dependency itself is reasonable — `get_chain` is the right abstraction to reuse. The
risk is that `nodes/shared/` has no stable public API contract (its `__init__.py` is empty).
Exposing `get_chain` via a path outside `nodes/` (e.g. a thin `cuga_graph.utils` module) or
adding it to `nodes/shared/__init__.py` would make the dependency explicit and refactor-safe.

---

## Questions

1. Is the intent to permanently couple `policy/` to `nodes/shared/`, or should `get_chain` be promoted to a stable shared location outside the `nodes/` subtree?

## Notes

- The existing `test_keyword_operator.py` and `test_similarity_integration.py` in the same directory also use real `PolicyStorage` without the `unit` marker — the label inconsistency on the new tests is a cosmetic issue, not a functional one.

src/cuga/backend/cuga_graph/policy/tests/test_nl_trigger_conflict_fallback.py
```diff
@@ -0,0 +1,150 @@
...
+def _playbook() -> Playbook:
+    return Playbook(
+        id="playbook_family_claims",
+        name="Family Healthcare Plan Navigation",
+        description="Guide for navigating family healthcare plans and claims",
+        triggers=[
+            NaturalLanguageTrigger(
+                value=["get my daughter's claims", "family member claims"],
+                target="intent",
+                threshold=0.7,
```
haroldship, line 28: This needs a test case for a threshold less than 0.5, since the method allows it.
The method will return a value less than 0.5, with possible downstream effects. The
old code always returned 0.5, so this is lower than before. What happens when there's
also a keyword match? match_policy picks between the two on confidence, so a lower one
is less likely to win. Is that intended?

Also I don't think the 0.5 default can ever happen - threshold always has a value, so
the list is never empty. I'd just remove it.

src/cuga/backend/cuga_graph/policy/agent.py
```diff
@@ -559,10 +573,13 @@ async def _resolve_nl_trigger_conflicts(
             logger.warning(f"    - Query text available: {query_text is not None}")
             if policies_with_nl_triggers:
                 policy, triggers = policies_with_nl_triggers[0]
```
haroldship: not clear to me why we take the first entry - is the list in descending order of some
kind? Worth a check and a comment, perhaps in the docstring above where it's described
as a list of policy tuples, and it's supposed to return the "best" policy.

I looked - it is sorted, but only when embeddings are configured and the search doesn't
fail. Otherwise it's whatever order came back from storage. So first isn't always best,
and the comment should say that.

Separately, I don't think this branch can run. The caller already returns None when
there's no llm and when there's no query text. Can we drop it?

src/cuga/backend/cuga_graph/policy/tests/test_nl_trigger_conflict_fallback.py
```diff
@@ -0,0 +1,150 @@
...
+    chain = MagicMock()
+    if side_effect is not None:
+        chain.ainvoke = AsyncMock(side_effect=side_effect)
+    else:
+        chain.ainvoke = AsyncMock(return_value=return_value)
+    return chain
+
+
+@pytest.mark.unit
+@pytest.mark.asyncio
```
haroldship, line 67: Also you should have tests with multiple triggers on one policy. The min() only does
something when a single policy has more than one NL trigger - two policies with one
trigger each won't reach it.

For example: two triggers with different thresholds both above 0.5, one with thresholds
above and below 0.5, and one with 2 different "targets".

A test with multiple policies is worth having too, but for the ordering question above,
not for this.

haroldship (changes requested):
The fix works - returning min(threshold) does get the fallback past the threshold
check, so #577 is fixed.

What bothers me is that the confidence isn't really a confidence any more, it's just
the number that makes the check pass. And it only works because the same min() is
computed in two places. If someone changes one and not the other, the bug comes back
and nothing catches it. Please pull it into one helper like CodeRabbit suggested.

Bigger question: when the LLM call throws we now return a match, but when the LLM
works fine and says no match, or gives low confidence, we return nothing. So a parse
error is more likely to trigger a policy than a working model is. Is that what we
want? If yes, please say so in a comment.

Things I'd like changed:
- remove the 0.5 default and the hasattr check, they can't be reached
- remove or document the "no llm / no query text" branch, the caller makes it dead
- tests with more than one trigger on a policy, and with a threshold under 0.5

On the get_chain change: structured output is provider dependent now, and for models
with GCP or Claude in the name get_chain returns a chain with no parser. That would
throw and land in the new fallback, so we'd always return the first policy and it
would look like a normal match. Nothing we ship hits that today, but the comment above
the call says the retry and json_mode fallback always apply and they don't. Worth
fixing the comment or passing wx_json_mode explicitly.

One more - both tests mock out get_chain, so the retry in the description isn't
actually tested, and it isn't new in this PR anyway. Maybe reword that.

Approve once those are in.

haroldship (approved):
Thanks for the fixes. All look ok except one minor thing:

The test `test_conflict_resolution_fallback_filters_by_target` passes even if the target filtering is removed - min(0.55, 0.95) is 0.55 either way, so the agent_response trigger never affects the result. Swap the thresholds (intent 0.95, agent_response 0.55) and assert 0.95, then it actually fails when the filter breaks.

Approving, please address this later in another PR.

PR #587, opened 2026-08-04
fix(registry): stop unwrapping arguments['params'] in ApiRegistry

### LINKED ISSUE #582: [Bug]: ApiRegistry silently unwraps arguments['params'] and drops sibling args
### What happened

`ApiRegistry.call_function` replaces the entire arguments dict with `arguments['params']` whenever a key named `params` is present. Sibling arguments are silently dropped before the call is forwarded to FastMCP.

Validated on `origin/main` @ `2538f1c2`.

### How to reproduce

1. Call a registry tool with a legitimate argument named `params` plus other fields:

```python
arguments = {
    "params": {"filter": "active"},
    "limit": 10,
    "user_id": "u-1",
}
```

2. Observe what FastMCP receives.

**Expected:** all keys forwarded as-is (`params`, `limit`, `user_id`).

**Actual:** only `{"filter": "active"}` is forwarded — `limit` and `user_id` are dropped.

Code path:

```python
# api_registry.py (~line 306)
args = arguments['params'] if 'params' in arguments else arguments
```

### Environment

- OS: n/a (code-path bug)
- CUGA: `origin/main` @ `2538f1c2`
- Component: tools registry / MCP call path

`src/cuga/backend/tools_env/registry/registry/api_registry.py`

### Checklist

sami-marreed:
## Bug fix

Fixes #582

### Summary
`ApiRegistry.call_function` treated a key named `params` as a wrapper and replaced the entire arguments dict before forwarding to FastMCP. Sibling fields were silently dropped. Arguments are now forwarded as-is.

### Testing
- Added unit test: `test_call_function_forwards_params_key_without_unwrapping_siblings`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed argument forwarding to ensure tools receive complete argument mappings without unintended extraction or discarding of fields.

* **Tests**
  * Added regression test to verify proper argument handling in tool calls.


sami-marreed:
Re: CodeRabbit nitpick on adding a downstream `MCPManager.call_tool` argument-shape test — skipping here.

#582 is at the `ApiRegistry.call_function` boundary: it must forward the full `arguments` mapping into `mcp_client.call_tool(...)`. The unit test asserts that contract.

How `MCPManager` later wraps args for traditional servers (`{"params": args, "headers": headers}`) vs FastMCP (passthrough) is existing manager behavior and a different layer. Covering that belongs with MCPManager tests if we want it, not as part of this registry unwrap fix.

offerakrabi (approved):

Issue #262, opened 2026-05-24
[Epic] Platform Health & Technical Debt


sami-marreed:
## What you want and why

Track cross-cutting **platform health** and **technical debt** work that improves maintainability without adding product features — UI consolidation, dead-code removal, and graph/modularity refactors.

## Scope

- Frontend legacy cleanup and workspace consolidation
- Backend dead-code removal and quarantine
- Shared graph modularity refactors (CugaLite + CugaSupervisor)

## Out of scope

- New product features → relevant product epics
- Bug fixes → file as bugs under component epics

## Issues

- #46 — Clean up old UI
- #78 — Cleanup unused backend code
- #205 — Modularize CugaLite and CugaSupervisor graphs
- #364 — Retire duplicate WorkspacePanel after ChatLanding workspace consolidation

sami-marreed:
Created to house platform health / tech-debt orphans: #46, #78, #205 (moved from #258).

sami-marreed:
Added under this epic: #367 — Visualize agent todos with Carbon ProgressIndicator in chat UI.

PR #592, opened 2026-08-05
chore(ci): add Dependabot configuration and limit workflow token permissions

### LINKED ISSUE #591: [Security]: Add Dependabot configuration and limit token permissions in two workflows
Part of #262.

### What you want and why

There are two small gaps in how this repository handles security updates.

**1. Dependabot has no configuration file.**

There is no `.github/dependabot.yml`. Dependabot still lists security alerts on the Security tab, but without this file it never opens a pull request to fix them. Someone has to notice the alerts and update the lockfiles by hand.

That is what PR #589 did. It closed 101 alerts in one manual sweep. It was good work, but nothing stops the same backlog building up again, and the next person has to do the same thing from scratch.

**2. Two workflows do not limit what their GitHub token can do.**

Every workflow run gets an automatic GitHub token. If a workflow does not say what that token is allowed to do, it is given broad write access to the repository whether it needs it or not. Two workflows do not say:

- `deploy-image.yml` — code scanning alert 59, open since 26 February
- `smoke-pip-install.yml` — code scanning alert 160

Neither job writes anything to the repository, so neither needs write access.

### How it could work

**Add `.github/dependabot.yml`** covering the three places dependencies come from:

- Python packages, from `pyproject.toml` and `uv.lock`
- Frontend packages, from the pnpm workspace at `src/frontend_workspaces`
- The versions of the actions used in `.github/workflows`

Check weekly, and group the updates so that one pull request covers many packages. All 45 currently open alerts come from the first two places, and without grouping that would arrive as 45 separate pull requests.

Keep security fixes in their own group, separate from routine version bumps, so an urgent fix is not held up behind an unrelated upgrade. Leave major version upgrades to arrive on their own, because they are the ones most likely to break something and they need a closer look.

**Add a `permissions` block to the two workflows**, setting them to read-only:

```yaml
permissions:
  contents: read
```

`deploy-image.yml` pushes to the IBM Cloud Registry using its own credentials, not the GitHub token, so read-only is enough. `smoke-pip-install.yml` only builds and tests. This is the same thing `tests.yml` and `external-fork-tests.yml` already do.

### Links or extra context

- PR #589 — the manual sweep this is meant to stop repeating
- Issue #337 — earlier request for a consistent way of handling Dependabot alerts
- https://github.com/cuga-project/cuga-agent/security/code-scanning/59
- https://github.com/cuga-project/cuga-agent/security/code-scanning/160
- https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference

Neither change alters how the product behaves. Both are repository configuration.

- [ ] I searched existing issues and this doesn't look like a duplicate

haroldship:
## Chore

Closes #591

### Summary

Repository configuration only. Nothing changes how the product behaves.

**1. Adds `.github/dependabot.yml`**

There was no Dependabot configuration. Dependabot still lists security alerts on the Security tab,
but without this file it never opens a pull request to fix them — someone has to notice and update
the lockfiles by hand, which is what #589 did last: 101 alerts closed in one manual sweep. Nothing
stopped that backlog rebuilding.

The file covers the three places dependencies come from: Python (`pyproject.toml` and `uv.lock`),
the frontend pnpm workspace at `src/frontend_workspaces`, and action versions in
`.github/workflows`. Checks run weekly on Mondays, grouped so one pull request covers many
packages — all 45 alerts open today come from the first two, and ungrouped they would arrive as 45
separate pull requests.

Three choices worth flagging for review:

- Security fixes are kept in a group of their own, separate from routine version bumps, so an
  urgent fix is not held up behind an unrelated upgrade.
- Major version upgrades are deliberately left ungrouped, so each arrives on its own. They are the
  most likely to break something and need a closer look.
- Routine updates split two ways per ecosystem — `python-tooling` / `python-runtime` by package
  pattern, `npm-dev` / `npm-prod` by dependency type. A group merges as a unit, so one bad bump
  blocks everything else in the batch. Two smaller groups mean broken build tooling cannot hold a
  runtime fix hostage, and a red batch says which half it came from. This came out of review.

`docs/examples` is left out on purpose — standalone illustrations rather than shipped code, so a
bump there is not fixing a risk to anyone using CUGA.

**2. Limits the token permissions in two workflows**

A workflow that does not declare `permissions` gets broad write access to the repository whether
it needs it or not. Code scanning flagged both: `deploy-image.yml` (alert 59, open since 26
February) and `smoke-pip-install.yml` (alert 160). Both now declare `permissions: contents: read`.
Neither job writes to the repository, and `deploy-image.yml` pushes to the IBM Cloud Registry with
its own credentials. This is the pattern `tests.yml` and `external-fork-tests.yml` already use.

**3. Removes two files that carried nothing**

Both came out of review: the root `package-lock.json` (arrived with #425, no `package.json` beside
it, nothing installs from it) and an empty generated `.ai/why/` template.

### Type of Changes

- [x] Chore / repository configuration

### Testing

- [x] All three files parse as valid YAML, and GitHub's own `dependabot.yml` validation passes
- [x] Pre-commit passed, including the secret scan
- [ ] Dependabot picks up the configuration after merge (only observable once it is on `main`)
- [ ] Code scanning alerts 59 and 160 close on the next scan

### Checklist

- [x] Commits follow Conventional Commits and are signed off
- [x] Linked to an issue, which is linked to epic #262
- [x] No product code touched

### Note for the reviewer

The grouping strategy has now been through a round of review and the routine groups were split as
a result. What is left open is `open-pull-requests-limit`, still at 5. With two routine groups per
ecosystem it is tighter than it was, so it may want raising once we see a real Monday. One-line
change either way.


sami-marreed:
### Review notes (dual pass)

Two independent reviews ([A](https://github.com/cuga-project/cuga-agent/pull/592) checklist) agreed on the same blockers before merge.

**OK as-is**
- `permissions: contents: read` on `deploy-image.yml` and `smoke-pip-install.yml` — neither workflow writes to the repo with `GITHUB_TOKEN`. Safe least-privilege tighten.

**Please fix before merge**
1. **Root `package-lock.json` not covered by Dependabot** — a lockfile exists at repo root (e.g. includes `pptxgenjs`), but `dependabot.yml` only watches `/src/frontend_workspaces`. Either add an `npm` entry for `"/"` or remove/document that the root lockfile is stale/intentional.
2. **GitHub Actions major updates are grouped** — the `github-actions` group has no `update-types` filter, so majors get batched. That contradicts the stated intent to keep majors ungrouped (same pattern as the other ecosystems). Add `update-types` (e.g. patch/minor only) or a separate security/major path.

**CI**
- Unit Tests failures seen earlier looked unrelated to this config-only diff (no Python changes). Re-check green before merge; don’t treat that flake as a product blocker for this PR.

Happy to re-review once the two Dependabot gaps are addressed.

.ai/why/chore-dependabot-config-and-workflow-permissions.md
```diff
@@ -0,0 +1,10 @@
+# Why: chore-dependabot-config-and-workflow-permissions
```
anupamamurthi: Is this file required? Can it be removed from this PR?

.github/dependabot.yml
```diff
@@ -0,0 +1,63 @@
...
+#
+# Without this file, Dependabot still reports security alerts on the Security
+# tab, but it never opens a pull request to fix them. Someone has to notice the
+# alerts and update the lockfiles by hand.
+#
+# Updates are grouped so that one pull request covers many packages. Ungrouped,
+# a backlog of 45 alerts would arrive as 45 separate pull requests.
+version: 2
+
+updates:
```
anupamamurthi, line 19: dependencies in docs/examples/*/pyproject.toml are uncovered - I guess that is fine for now but wanted to point out just in case.  I am not sure if alerts get raised for modules in docs/examples/ or if they are ignored. (probably better to ignore them as those are just examples)?

anupamamurthi:
AI Review Suggestions (I don't think the current PR should be blocked but maybe a follow-up or something to keep in mind) and the issue that is discussed is real!


Suggestion: split the routine groups along a fault line.

Optimize for blast radius, not PR count. A group merges as a unit — one bad bump blocks the other 29, and you can't partially merge or strip a package out of a re-resolved lockfile. Worse, "Bump the python-minor group with 30 updates" tells you nothing when CI goes red; you've traded 30 self-labeling signals for one anonymous one. And the odds of red scale with size (~2% per-package risk × 30 packages ≈ 45% chance of a red Monday).

Grouping is still right — 45 ungrouped PRs would just rot. But group size is a dial, not a switch.

uv — split by pattern (deterministic; avoids relying on how Dependabot classifies [dependency-groups]):


      python-security:
        applies-to: security-updates
        patterns: ["*"]
      python-tooling:
        applies-to: version-updates
        patterns: ["pytest*", "ruff*", "mypy*", "types-*", "coverage*", "pre-commit*"]
        update-types: ["minor", "patch"]
      python-runtime:
        applies-to: version-updates
        patterns: ["*"]
        exclude-patterns: ["pytest*", "ruff*", "mypy*", "types-*", "coverage*", "pre-commit*"]
        update-types: ["minor", "patch"]
npm — split by dependency-type (unambiguous here, no glob maintenance):


      npm-security:
        applies-to: security-updates
        patterns: ["*"]
      npm-dev:
        applies-to: version-updates
        dependency-type: "development"
        update-types: ["minor", "patch"]
      npm-prod:
        applies-to: version-updates
        dependency-type: "production"
        update-types: ["minor", "patch"]
Majors stay ungrouped as you have them. The npm split also separates shipped code from build tooling across the five workspace packages — different review bars, and one PR can't express that.

Cost: one extra PR per ecosystem per week. Gain: expected share of updates stuck drops ~45% → ~28%, a red batch names its own category, and broken tooling stops holding runtime fixes hostage.

.github/dependabot.yml
```diff
@@ -0,0 +1,63 @@
...
+      # One pull request for routine minor and patch upgrades. Major upgrades
+      # are left on their own because they are more likely to break something.
+      python-minor:
+        applies-to: version-updates
+        update-types: ["minor", "patch"]
+
+  # Frontend packages. This is the pnpm workspace root; Dependabot picks up the
+  # nested package.json files under it from pnpm-lock.yaml.
+  - package-ecosystem: "npm"
+    directory: "/src/frontend_workspaces"
```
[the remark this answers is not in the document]
haroldship: Removed the lockfile in d017b102.

An `npm` entry for `/` would not have worked: there is no `package.json` at the root, and Dependabot needs a manifest to resolve a lockfile against. The file is orphaned — it came in with PR #425, no workflow or script installs from it, and the one place `pptxgenjs` is actually needed is the presentation e2e test, which runs `npm install pptxgenjs --prefix <tmpdir>` into a throwaway directory.

So the stray file was the problem, not the gap in coverage.

.github/dependabot.yml
```diff
@@ -0,0 +1,63 @@
...
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    open-pull-requests-limit: 3
+    commit-message:
+      prefix: "chore(ci)"
+    groups:
+      github-actions:
+        patterns: ["*"]
```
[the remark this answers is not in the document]
haroldship: Fixed in a52de50e, close to the suggested shape.

The routine group is now explicit about `applies-to: version-updates` and carries `update-types: ["minor", "patch"]`, and there is a separate `github-actions-security` group for security updates. Major action upgrades now arrive on their own, which is what the Python and npm entries already did. The inconsistency was an oversight, not a decision.

.ai/why/chore-dependabot-config-and-workflow-permissions.md
haroldship: No, it is not required, and it is gone as of d7dcc47d.

It was a generated template and both of its sections were still empty, so it carried no information at all. The reasoning it was meant to hold is in the comments at the top of `dependabot.yml` and in the commit messages, which is where it is more likely to be read.

.github/dependabot.yml
haroldship: Correct, they are not covered, and that is deliberate. Your instinct is the one I went with.

To separate the two halves of your question: alerts can still be raised for those files, because the dependency graph scans the whole repository on its own and does not read this configuration. What `dependabot.yml` controls is only whether a pull request gets opened. So we keep the visibility on the Security tab and skip the weekly pull request.

The reason to skip it is that `docs/examples/*` are standalone illustrations. Nothing installs them as part of CUGA, so a bump there is not fixing a risk to anyone using the product. Only `docs/examples/demo_apps/setup` has a lock file, so the other three would give Dependabot little to work with in any case.

I have added a note at the top of the file recording this, so the next person reads it as a decision rather than an oversight. If we later decide the examples should be kept current, adding directories to the `uv` entry is a small change.

haroldship:
### Review addressed

@sami-marreed @anupamamurthi — all four points are handled. Three commits, all configuration.

**1. Root `package-lock.json` not covered** (Sami, CodeRabbit)

Removed the file (d017b102) rather than adding an `npm` entry for `/`.

Adding the entry would not have worked. There is no `package.json` at the root, and Dependabot needs a manifest to resolve a lockfile against. The lockfile is orphaned: it came in with PR #425, nothing installs from it, and the only place `pptxgenjs` is needed is the presentation e2e test, which runs `npm install pptxgenjs --prefix <tmpdir>` into a throwaway directory. The stray file was the problem, not the gap in coverage.

**2. GitHub Actions majors were grouped** (Sami, CodeRabbit)

Fixed (a52de50e). The routine group now carries `update-types: ["minor", "patch"]`, and there is a separate `github-actions-security` group. Majors arrive on their own, matching what the other two ecosystems already did. That was an oversight, not a decision.

**3. Group size** (Anu)

Taken as written (a52de50e), not deferred. The blast-radius argument is right: a group merges as a unit, so one bad bump blocks the rest and you cannot strip a package out of a re-resolved lockfile. Routine updates are now split two ways per ecosystem:

| Ecosystem | Split by | Groups |
|---|---|---|
| `uv` | package pattern | `python-tooling`, `python-runtime` |
| `npm` | dependency type | `npm-dev`, `npm-prod` |

Security fixes keep their own group in each, and majors stay ungrouped. Cost is one extra pull request per ecosystem per week; the gain is that broken tooling no longer holds a runtime fix hostage, and a red batch says which half it came from.

`open-pull-requests-limit` stays at 5 for now. With two routine groups instead of one it is tighter than it was, so it may need raising once we see a real Monday. It is still a one-line change.

**4. Empty rationale document** (Anu, CodeRabbit)

Removed (d7dcc47d). It was a generated template with both sections still empty. The reasoning is in the file comments and the commit messages instead.

**Also**

Recorded in the file that `docs/examples/*` are out of scope on purpose — they are standalone illustrations, not shipped code. Alerts can still surface for them on the Security tab, since the dependency graph scans the repository independently of this file; what we are skipping is only the weekly pull request. Details in the thread reply.

**CI**

Sami — agreed the earlier Unit Tests failure was unrelated to a configuration-only diff. Checks are running again now. GitHub's own `.github/dependabot.yml` validation check passes on the new config, so the schema is good; the rest only becomes observable once this is on `main`.

Re-requesting review from you both.

haroldship:
@sami-marreed — I went back through the history on the root lockfile, because deleting a file on a hunch is not good enough. The record is clearer than I thought, and it turns out you have seen this file before.

On `feat/199-spawn-sub-agents-at-runtime`:

- `7488371e` (15 June, Iftach) added `package.json` and `package-lock.json` at the root. The manifest was five lines: one dependency, `pptxgenjs`.
- `ef4e32a4` (13 July, you) deleted `package.json` as part of hardening the agent spawn isolation. The lockfile was left behind.
- `2538f1c2` (4 August) squash-merged that branch to `main` as #425, carrying the orphan along.

So the lockfile has never had a manifest beside it on `main`, not for a single commit. What the Dependabot gap actually surfaced was a leftover from a cleanup you had already done.

What I checked before removing it:

- No root `package.json` has ever been tracked on `main`, and `package.json` is not in `.gitignore`, so nothing is hiding untracked.
- No `node_modules` at the repo root.
- No workflow, Dockerfile, script, or doc installs from it. It is not in `.secrets.baseline`, and packaging is `setuptools` with `where = ["src"]`, so it was never shipped in the wheel either.
- The one real `pptxgenjs` user is the presentation e2e test, and it is fully isolated: `npm install pptxgenjs --prefix <tmp_path>`, then `NODE_PATH=<workdir>/node_modules` with `cwd=<workdir>`. It never looks at the repo root.
- The sandbox executors run in `cuga_workspace/`, not the repo root.

And the direct check: with the lockfile restored into an empty directory, `npm ci` fails with `ENOENT ... Could not read package.json`. The file could not be acted on even if something tried.

That is why I removed it rather than adding an `npm` entry for `/`. Dependabot needs a manifest to resolve against, and there has not been one since July.

anupamamurthi (approved):

PR #595, opened 2026-08-06
fix(trajectory): record code not content for Assistant_code step

### LINKED ISSUE #586: [Bug]: Assistant_code trajectory step logs raw LLM content instead of extracted code
### What happened

When code is extracted from an assistant response, `Assistant_code` is written with the full raw LLM `content` instead of the extracted `code`. That makes `Assistant_code` a duplicate of `Raw_Assistant_Response`.

**Fix requirement:** keep **both** steps — do not collapse to one.
- `Raw_Assistant_Response` → full raw LLM `content` (unchanged purpose)
- `Assistant_code` → extracted `code` only (when code was extracted)
- `Assistant_nl` → raw `content` when there is no code (existing branch)

Validated on `origin/main` @ `2538f1c2`.

### How to reproduce

1. Run a CugaLite turn where the model returns prose + a python fence.
2. Inspect trajectory steps.

Example model output:

```text
Sure, here is the code:
```python
print(await find_tools("orders"))
```
```

Code path:

```python
self._tracker.collect_step(Step(name="Raw_Assistant_Response", data=content))
if code:
    self._tracker.collect_step(Step(name="Assistant_code", data=content))  # same content!
```

**Expected** for a code turn:

```json
{"name": "Raw_Assistant_Response", "data": "Sure, here is the code:\n```python\nprint(await find_tools(\"orders\"))\n```"},
{"name": "Assistant_code", "data": "print(await find_tools(\"orders\"))"}
```

**Actual:** both steps have the same raw `content` string.

**Non-goal:** removing `Raw_Assistant_Response` or only logging extracted code.

### Environment

- OS: n/a (code-path bug)
- CUGA: `origin/main` @ `2538f1c2`
- Component: CugaLite graph adapter / activity tracker

`src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/graph_adapter.py` — `on_response_processed`

### Checklist

offerakrabi:
## Bug fix

Fixes #586

### Summary

`AgentGraphAdapter.on_response_processed` was passing `content` (the full normalized model response) as the payload for the `Assistant_code` trajectory step instead of `code` (the extracted Python body). This meant downstream trajectory consumers always saw the same value in both `Raw_Assistant_Response` and `Assistant_code` when a response contained prose plus a code block.

**Root cause:** A one-word typo on line 182 of `adapter/graph_adapter.py` — `data=content` instead of `data=code`.

**Fix:** Change `data=content` → `data=code` for the `Assistant_code` step. No other production changes.

The two existing adapter unit tests for `on_response_processed` only called `assert_called()` (no data assertions), so the bug was invisible to them. They have been replaced with tests that:
- Use distinct `content` (prose + fenced block) vs `code` (Python body) values
- Assert exact ordered `call_args_list` with correct data on each step
- Carry the `@pytest.mark.unit` marker

### Testing

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved response tracking so code responses record only the extracted code, rather than the full response text.
  - Natural-language responses continue to record their complete content.
  - Raw assistant responses remain available for tracking and review.

- **Tests**
  - Added coverage to verify tracking behavior for both code and natural-language responses.

### DIFF
--- a/src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/graph_adapter.py
+++ b/src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/graph_adapter.py
@@ -179,7 +179,7 @@ class AgentGraphAdapter(CoreGraphAdapter):
         try:
             self._tracker.collect_step(step=Step(name="Raw_Assistant_Response", data=content))
             if code:
-                self._tracker.collect_step(step=Step(name="Assistant_code", data=content))
+                self._tracker.collect_step(step=Step(name="Assistant_code", data=code))
             else:
                 self._tracker.collect_step(step=Step(name="Assistant_nl", data=content))
         except Exception as exc:
--- a/src/cuga/backend/cuga_graph/nodes/cuga_lite/tests/test_agent_graph_adapter.py
+++ b/src/cuga/backend/cuga_graph/nodes/cuga_lite/tests/test_agent_graph_adapter.py
@@ -182,7 +182,8 @@ def test_normalize_response_extracts_reasoning():
 # ── 6. on_response_processed hook ────────────────────────────────────────
 
 
-def test_on_response_processed_calls_tracker_for_code():
+@pytest.mark.unit
+def test_on_response_processed_code_branch_records_code_not_content():
     AgentGraphAdapter = _get_adapter_class()
     tracker = _make_tracker()
     adapter = AgentGraphAdapter(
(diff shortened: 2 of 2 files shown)


src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/graph_adapter.py
```diff
@@ -179,7 +179,7 @@ def on_response_processed(self, state: Any, code: Optional[str], content: str) -
         try:
             self._tracker.collect_step(step=Step(name="Raw_Assistant_Response", data=content))
             if code:
-                self._tracker.collect_step(step=Step(name="Assistant_code", data=content))
+                self._tracker.collect_step(step=Step(name="Assistant_code", data=code))
```
sami-marreed, line 182: I wrote in the issue to support recording both, are they recorded both now?

sami-marreed (approved):

PR #602, opened 2026-08-06
feat(events & channels): event-driven agents — channels, triggers, and NL→Flow arming

anupamamurthi:
## Summary

CUGA answers when a person asks. This adds the other two ways to wake an agent — **the clock**, and
**the world changing** — as a second service beside CUGA rather than inside it. Agents can watch
inboxes, repos, folders and schedules, converse on Slack/Discord/Telegram/web, and deliver answers
back into the thread they were armed from.

**The one rule:** every message from every channel lands on CUGA's `POST /run`, and *CUGA* — not the
eventing service — decides whether it is ordinary chat or an attempt to arm something. A slash verb
(`/automate`, `/watch`, `/schedule`, …) or a thread with an arming dialogue already open goes to the
eventing service. Everything else is ordinary chat and never touches it.

Don't deploy the second service and CUGA is unchanged: `EVENTS_API_URL` unset means a slash verb is
just text.

> **This is PR 1 of 2.** The tooling (`Makefile`, `scripts/`), the Code Engine deploy kit, the
> example rosters and most of `events_docs/` are in a stacked follow-up so this one could be
> reviewed at all. Nothing here depends on them — the CI gate was re-run green with all 73 removed.
> Consequence for reviewers: `make …` targets arrive in PR 2; use `uv run pytest` (commands below).

## What a reviewer should actually read

146 files, but almost all of it is new code sitting *beside* CUGA. **Five files of CUGA core are
touched, and only 18 lines are removed across the whole PR:**

| File | Change |
|---|---|
| `backend/server/main.py` | +466 −1 |
| `supervisor_utils/supervisor_config.py` | +32 −16 |
| `cli/main.py` | +28 −1 |
| `tools_env/registry/config/mcp_servers_cuga_apps.yaml` | +38 −0 |
| `sdk_core/tests/test_supervisor_policy_default.py` | +124 −0 *(new test)* |

Everything else is `src/cuga/backend/events/` (47 modules), its tests, and the frontend surface.

Coupling runs one way and is enforced: **CUGA imports nothing from `cuga.backend.events`** — a test
(`test_core_decoupled.py`) fails if that regresses. The events package imports exactly one thing from
CUGA (`resolve_secret`). The only link out of core is an HTTP POST guarded by one env var.

### What that core code actually is, and why it has to live here

Those 466 lines in `main.py` are five distinct things. None of them know what an event is.

**1. `POST /run` — the worker seam.** The bulk of the addition, and the one piece worth reading
closely. `/stream` is built for the web UI: it holds a connection open and emits every intermediate
step as SSE. A service-to-service caller wants the opposite — one request, one JSON answer. `/run`
drives the **same `event_stream` generator** `/stream` does, as an in-process call rather than an
HTTP call to itself, so knowledge and attachments, conversation history, citations, policies and
HITL all behave identically *by construction*. Only the output adapter differs: intermediate frames
are dropped and the terminal Answer becomes the response body. Why it lives in core: this is how
anything executes an agent. The eventing service can't run a graph — it has no agent, no tools, no
model config. It only ever asks CUGA to.

**2. `GET /run/agents` — the roster seam.** One endpoint answering "what is loaded here?" Why it
exists: **the roster belongs to whoever executes**, and in a split deployment that is this server.
Without it the events side had to guess, and it guessed "one agent" — so a webhook pinned to a real
sub-agent (`?agent=incident_triage`) was rejected as unknown before it ever reached the supervisor.
Deliberately *not* `/api/agents`: that one belongs to the dashboard, sits behind the manage-access
cookie and returns UI card data. This is the machine seam, behind the same shared secret as `/run`.

**3. The optional preloaded supervisor.** With `CUGA_SUPERVISOR_ROSTER=<path>` set, `/run` builds a
`CugaSupervisor` from the roster once and routes every call through it. Why here: the roster used to
be loaded only by the events layer's in-process runtime. That is fine while the two share a process,
but once the worker moved across an HTTP hop the executing side is vanilla CUGA — no events layer,
no runtime, no roster — so every call quietly ran as CUGA's lone default agent. Callers do not pass
an agent and shouldn't have to: they address the supervisor and it routes internally. Unset the
variable and none of this runs.

**4. The slash forwarder (~55 lines).** `_SLASH_VERBS`, `_forwards_to_events`,
`_forward_slash_to_events`. This is the part reviewers usually ask about, because it looks like
eventing logic in core. It isn't — it is a *guard*. If `/automate every 5 minutes tell me the IBM
price` reaches the plain agent, the agent tries to **implement** the schedule: a loop with sleeps
that quietly does nothing useful. That is the silent-failure trap this whole feature exists to
close, so core has to recognise the intent at the door and forward one POST. What it costs: one
regex, one in-memory set, one `httpx` call. No events import, no shared database, no bot tokens.
The in-memory set (`_events_open_threads`, so a bare "yes" continues the dialogue) is deliberately
*not* persisted — core must never read the events store; it is a routing hint, not state, and losing
it on restart costs a user one retype. Guarded twice over: `EVENTS_API_URL` unset short-circuits to
`False` before the regex ever runs, and if the service is configured but down, the exception is
caught and returns an honest sentence naming the URL. Chat never calls out at all.

**5. One field on `/api/ui/config`.** `events_api_url`, so the SPA can find the events service. The
SPA otherwise assumes a single origin, so in a split topology its Studio calls land here — where
`/api/events/*` doesn't exist — and silently fall through to the SPA catch-all. Empty (the default)
means "same origin", so nothing changes when unset.

**And in `cli/main.py` (+28):** the registry needs FILE mode whenever a roster of sub-agents is in
play; `MCP_SERVERS_FILE="none"` means managed-config-db mode, which serves only the demo app. That
value was being assigned unconditionally, overwriting an explicitly exported one — which is why a
`cuga start demo` served only `digital_sales` while every roster agent answered "the available
toolset does not include…" despite the environment being set correctly. Now it is a `setdefault`
when a roster is wanted. Also here: the sandbox execution timeout gets a `setdefault` of 120s
because slow external APIs (arXiv at ~5.5s per call, plus retries) blow the 30s default — an
explicit env var still wins.

**One restoration worth disclosing:** the GC of ephemeral stream-event rows in `lifespan` was lost
in this branch's merge with `main` and is restored here. Nothing failed loudly; the rows just
accumulated forever. The same merge dropped the events background-loop launcher, which now correctly
lives in the eventing service's own lifespan — this server starts no event loops at all.

## `supervisor_config.py` — what changed, and what deliberately didn't

**Policy auto-loading is unchanged for every existing caller.** `load_supervisor_config` gains one
keyword-only parameter that defaults to `None`:

```python
async def load_supervisor_config(yaml_path: str, *, auto_load_policies: bool | None = None)
```

Precedence is: a per-agent `auto_load_policies:` key in the YAML wins; otherwise the caller's
default; otherwise `None`, which lets `CugaAgent` fall back to `settings.policy.auto_load_policies`
exactly as it always has. `CugaSupervisor.from_yaml()` and `cuga_graph/graph.py` pass nothing, so
they are untouched.

Only the headless caller opts out, and it does so at the call site rather than imposing it on
everyone — `server/main.py`, preloading the events supervisor:

```python
cfg = await load_supervisor_config(path, auto_load_policies=False)
```

The reason is specific to that path: everything this supervisor runs is a scheduled tick, a webhook
or a channel event. Nobody is present to answer an approval interrupt, and one would hang the run
until the caller times out.

An earlier cut of this branch hardcoded `False` inside `load_supervisor_config`, which silently
disabled policy auto-loading for every downstream supervisor user regardless of their settings, with
nothing raised and nothing logged. That is fixed, and `sdk_core/tests/test_supervisor_policy_default.py`
pins the contract — including that the parameter stays keyword-only with a `None` default, so no
existing positional call can break.

**The one behaviour change that remains:** `CombinedToolProvider` is now actually scoped to the named
apps/servers, where the filter was previously a no-op that logged intent and loaded everything.
Hyphenated names map to underscores (`cuga-finance` → `cuga_finance`; the hyphenated form composed
invalid Python identifiers downstream, so `cuga-finance_get_price` parsed as subtraction). An agent
that names nothing still gets all tools, so that path is unchanged — but a supervisor YAML that names
apps and relied on getting everything anyway will now get only what it named.

## What's in it

**Channels — 4.** Web/Studio, Slack (Events API), Discord (Gateway websocket), Telegram
(long-polling). Discord and Telegram are outbound connections, so they need no public URL. Answers
return to the originating thread; the delivery address rides on `thread_id`.

**Triggers — 42, from one registry.** One table generates the classifier phrases, the flow builders,
the docs and the tests. **15 are `backend=direct`** (Slack 8 · Discord 3 · Box 2 · Telegram 1 ·
webhook 1) and need no external infrastructure; **27 are Activepieces-backed push** (GitHub 14 ·
Gmail 4 · Calendar 3 · Pinterest 3 · Box 1 · RSS 1 · YouTube 1). Plus a native in-process scheduler
for cron and poll, and a key-checked inbound webhook.

**NL→Flow with a human gate.** An utterance compiles to a typed `FlowSpec` through a deterministic
pipeline (classify → slot extraction → registry validation) with an LLM fallback for what the
resolver can't call honestly — both arm through the same tool. The contract: *an utterance resolves
to exactly the right flow, or to a question — never silently to the wrong flow.* Nothing is armed
until a human sees a card showing the cadence and **the exact prompt the agent will be handed on
every fire**; arming is then deterministic from the approved text.

**Stateful poll — the delta gate.** A cron always reports; a poll speaks only when something changed,
across four tiers (always · threshold · identity · fuzzy). The *fingerprint* is stored, never the
content, so it scales.

**Web delivery.** Slack/Discord/Telegram can be pushed into; a browser cannot. A flow armed in a web
chat delivers to a durable per-thread mailbox that the chat drains, so a fire lands in the thread it
was armed from and survives a closed tab. Without it a web-armed flow ran, logged, and told nobody.

**Operations.** A live dashboard (watchers with pause/resume/delete/run-now, runs with the tools each
actually invoked, an inline dry-run composer), per-run disk logs, one trace id across both processes,
and a startup capability report that names what is live and what still needs infrastructure — each
with its one-line fix.

**Multi-tenancy.** Everything is scoped to a principal (tenant, instance, user); an identity map
resolves a channel-native id to a CUGA user under a shared bot; credentials are either a shared
service account or per-user OAuth, and the agent never sees a raw token.

**Storage.** `EVENTS_DB` takes a PostgreSQL URL, and local dev runs the same engine as the
deployment. Local used to be a durable SQLite file while the deployment ran SQLite on an ephemeral
disk — so the fragile path was the only one nobody exercised, and a pod replacement silently deleted
armed flows with no restart recorded. SQLite is retained for the hermetic offline suite and a
zero-infrastructure quickstart.

## What's deliberately NOT in it

- **No connector actions.** The layer arms triggers and delivers answers. Anything the agent should
  *do*, it does through its own tools, which keeps credentials and side effects inside the agent
  instead of spread through event plumbing. The action half was built and then deleted (~2,640 LOC)
  rather than gated, because a half-present feature is worse than an absent one.
- **Activepieces is off by default.** It is one trigger backend for push on SaaS we don't own, not
  the spine. Its ~1,842 LOC stay intact behind flags; arming an AP trigger returns CONNECT NEEDED.
- **No routing hints in prompts.** Rosters used to carry a `HANDLES TRIGGERS: app/event (…)` line per
  sub-agent. They were removed: the supervisor never saw them. Its routing prompt renders each
  sub-agent as `{{ agent['description'] }}`, `description` falls back to `getattr(agent,
  "description", f"Internal agent: {name}")`, and `CugaAgent` defines none — so the hints only ever
  reached the sub-agent's *own* prompt, read after routing had already picked it, at ~50% of the
  roster's prompt text. Trigger ownership lives in the structured `integrations[].triggers` on each
  `AgentSpec`, which code actually consults; a test now fails if the prose comes back.

## Testing

The `Makefile` ships in PR 2, so run pytest directly here — this is also exactly what CI runs:

```bash
uv run ruff check && uv run ruff format --check      # the CI lint gate
uv run pytest tests/events -q                        # 393 tests across 20 files — hermetic, SQLite
EVENTS_TEST_PG_DSN=postgresql://… uv run pytest tests/events/test_db_postgres.py   # the SQL path that ships
```

**32 live harnesses** under `tests/events/live_*.py` are included deliberately, even though CI never
runs them (they need real Slack/Discord/Telegram/AP credentials). They are the evidence the feature
works against real providers, and a reviewer asking "is this proven?" should be able to read them
here rather than in a follow-up.

The API reference and per-connector setup guides are enforced by contract tests — add a route or a
trigger without documenting it and the build fails.

**Verified end to end:** chat on all 4 channels against real provider APIs; a human-typed Slack
`@mention` → in-thread reply; a real Slack `/automate` → card → `yes` → ARMED; cron + poll fire
(2/2 local, 2/2 deployed); a web-armed cron delivering into the browser mailbox; inbound webhook
across 3 modes; arm → kill the process → cold boot → the flow is still there. Against the deployed
app: `live_e2e` 29 passed / 0 failed, delivery probe PASS, trigger×sink matrix PASS with 0 errors.

**Honestly partial:** the 15 direct watchers are proven from signature → routing → match → dispatch →
delivery, but *that the provider actually sends that event type* is not — it depends on per-event app
subscriptions. Box-direct on the deployed app is unproven (needs a fresh 60-minute developer token).
Poll tier selection picks `fuzzy` on Code Engine where it picks `threshold` locally; diagnostics ship,
root cause open.

## Follow-ups, not blockers

- **PR 2** (`events-followup-tooling-docs`) restores the Makefile, `scripts/`, the Code Engine deploy
  kit, the example rosters and the rest of `events_docs/`.
- `src/cuga/frontend/dist/` bundles are committed here; de-blobbing needs a CI build step first.
- Activepieces on the cluster is separate infrastructure (app + Postgres + Redis), not a config flag.
- The deployed events service is anonymously readable and writable on a public route — see the
  security note in PR 2's docs; `EVENTS_WEBHOOK_KEY` closes the worst of it in one env var.
 

---
---

## Summary by CodeRabbit

* **New Features**
  * Added an Events service and Event Studio for managing agents, integrations, channels, triggers, subscriptions, runs, and event flows.
  * Added concierge chat with confirmation-based scheduling, polling, push triggers, webhooks, and delivery tracking.
  * Added direct Slack, Discord, Telegram, Box, and web delivery options.
  * Added dashboards, inbox updates, API reference, examples catalog, OAuth connections, identity linking, and administration tools.
  * Added native scheduling, persistent event history, tenant isolation, and improved MCP tool scoping.

* **Documentation**
  * Added setup guides for supported integrations and comprehensive API and examples documentation.

* **Tests**
  * Added extensive offline, integration, persistence, and live event coverage.


tests/events/test_events_api_contract.py
```diff
@@ -0,0 +1,1163 @@
...
+    assert r.status_code == 400 and "missing user_id" in r.json()["error"]
+
+
+def test_admin_add_user_without_a_user_store_is_501():
+    c, _ = _client(users=None)
+    assert c.post("/api/events/admin/users", json={"user_id": "bob"}).status_code == 501
+
+
+# ── the spec must stay golden ─────────────────────────────────────────────────
+def test_api_spec_is_golden():
```
sami-marreed: **1. Generated HTML in the repository**

Do we need `events_docs/api/{api,api_spec,examples}.html` checked in? They read as generated artifacts (about 313 KB in total: spec 204 KB, examples 72 KB, api 38 KB). CodeQL already flags `examples.html`, and they increase clone size.

CI requires them because this test runs `gen_api_spec.py --check` / `gen_examples.py --check` and greps `api.html`. Studio then serves the same files from `GET /api/events/docs/{page}`.

Could we generate them at docs-build or service start, gitignore the HTML, and keep `--check` against a temporary file? FastAPI `/docs` would cover most of what `api_spec.html` is doing.

.vscode/launch.json
```diff
@@ -1,93 +0,0 @@
-{
```
sami-marreed, line 1: **2. Deletion of `launch.json`**

Why was this file removed? It looks like a side effect of the 602/603 split (`6ef2208` deferred `.vscode/` so CodeRabbit would stay under 150 files), not an intentional product change.

If 602 merges on its own, the AppWorld, Demo, Windows, and eval debug configurations disappear until 603 restores them. Could we restore this file in 602 and leave any Events-only debug configuration for 603?

README.md
```diff
@@ -37,6 +37,7 @@ Building a domain-specific enterprise agent from scratch is complex and requires
 > | **Reasoning modes** (fast / balanced / accurate) | `[features] cuga_mode` in [`settings.toml`](src/cuga/settings.toml) · [`configurations/modes/`](src/cuga/configurations/modes/) |
 > | **Hybrid API + browser tasks** | `[advanced_features] mode = 'hybrid'` · Playwright + [browser extension](src/frontend_workspaces/extension/readme.md) |
 > | **Multi-agent (CugaSupervisor)** | `cuga start demo_supervisor` · `[supervisor]` in [`settings.toml`](src/cuga/settings.toml) |
+> | **Event-driven agents** (channels · triggers · standing flows) | **`make up-noap`** — Slack/Discord/Telegram/web chat, webhooks, cron/poll/push flows armed from natural language with a human confirming each one, one supervisor agent over a YAML roster. Runs as **a second service beside CUGA** (`cuga-core` + the eventing service); CUGA is unchanged when it is not deployed. Full stack incl. Activepieces + tunnels: `make up`. [Docs](events_docs/README.md) · [Setup](events_docs/SETUP.md) |
```
sami-marreed: **3. Documentation location**

Should events documentation live on docs.cuga.dev, or at least under `docs/`, rather than in a parallel `events_docs/` tree?

`docs/` already exists for examples, and docs.cuga.dev is the published site. This pull request adds a third tree. These README links point at `events_docs/README.md` and `SETUP.md`, both of which were moved to 603, so they are already broken on this branch.

The setup guides are useful operator documentation. The hand-written HTML API boards are not. Studio still tells people to set `EVENTS_ENABLED=1`, which has been removed.

scripts/gen_api_spec.py
```diff
@@ -0,0 +1,2901 @@
+#!/usr/bin/env python3
+"""Generate `events_docs/api/api_spec.html` — the golden, try-it-yourself API spec.
+
+    python scripts/gen_api_spec.py            # rewrite events_docs/api/api_spec.html
+    python scripts/gen_api_spec.py --check    # exit 1 if the file is stale (used by the test)
+
+**Why a generator.** A hand-written spec rots the moment someone adds a route, and nobody notices
```
sami-marreed: **4. Generator scripts under `scripts/`**

Why do `gen_api_spec.py` and `gen_examples.py` live as top-level scripts? They appear to have stayed in this pull request only so the contract tests can run `--check`.

`gen_examples.py` (74 lines) is reasonable. This file (about 2,901 lines / 134 KB) is an endpoint catalog plus an HTML renderer. If the HTML is not committed, could these move into `tests/events/` or sit next to the package?

src/cuga/backend/events/__init__.py
```diff
@@ -0,0 +1,26 @@
+"""The event-driven layer for CUGA — its OWN SERVICE (``cuga.backend.events.service``).
+
+A concierge turns natural language into standing flows; every trigger (channel message, app
+event, cron, run-once) re-enters through ``POST /invoke``, and the worker call goes out to CUGA's
+``POST /run`` over HTTP. This package is never mounted onto CUGA's app — that "combined" mode was
+removed, along with the ``EVENTS_ENABLED`` flag that gated it. See ``events_docs/ARCHITECTURE.md``.
```
sami-marreed: **5. Flat package — client versus service**

`src/cuga/backend/events/` is 47 flat modules. Is that the entire library? Is there a separation between the HTTP client CUGA uses and the events service itself?

There is a conceptual split, but not a package split:

- Service: `service.py` and `app.py` (about 3,075 lines of routes)
- Events-to-CUGA client: `cuga_door.py` and `runtime.HttpRuntime`
- CUGA-to-events client: the slash forwarder in `server/main.py`
- Persistence: `agent_store.py`, `db.py`, `subscriptions.py`, and related modules

`agent_store.py` itself is only 95 lines. Could we split this into `api/`, `runtime/`, `channels/`, and `store/` before this layout becomes the default on `main`?

src/cuga/backend/server/main.py
```diff
@@ -2012,6 +2019,18 @@ async def event_stream(
 app.include_router(manage_routes.router)
 app.include_router(secrets_routes.router)
 
+# ---- the eventing layer is a SEPARATE SERVICE ----------------------------------------------
```
sami-marreed, line 2090: **6. Growth of `main.py`, and the roster**

`main.py` grew by 504 lines, to 4,847. What is the roster, and why does it live in CUGA core? Can we extract this instead of continuing to grow this file?

The roster is the `CUGA_SUPERVISOR_ROSTER` YAML. In a split deployment, CUGA is the worker. Without a roster on this side, every `/run` used the default agent, and webhooks pinned to `incident_triage` were rejected. `/run/agents` is the machine-facing seam so the events service does not have to guess. That part is sound.

Events APIs already live in `events/app.py`, not here. Adding events middleware on `main` would pull the layer back into CUGA. A better approach is to extract `/run`, the roster, and the slash forwarder into something like `cuga.backend.server.run_routes`, following the same pattern as A2A.

One further issue: with a roster loaded, `/run` is not the non-streaming sibling of `/stream`. It calls `supervisor.invoke`, drops attachments, resume, and HITL, and “pins” `?agent=` by rewriting the prompt. The supervisor model can ignore that instruction. Either invoke the named sub-agent (or a real routing API), or drop the claim that this is the same graph.

src/cuga/backend/server/main.py
```diff
@@ -2433,6 +2487,454 @@ async def stream(
...
+    Reply: ``{ok, status, answer, thread_id, sources, variables, error}`` where ``status`` is
+    ``ok`` | ``error`` | ``interrupt`` (the graph paused for human input).
+
+    Knowledge bases are NOT passed here — they attach out-of-band to the ``thread_id`` (session
+    scope) or to the agent, exactly as for /stream, and are picked up by identity at run time.
+    """
+    from cuga.backend.cuga_graph.nodes.human_in_the_loop.followup_model import ActionResponse
+    from cuga.backend.cuga_graph.utils.agent_loop import StreamEvent
+
+    token = _run_token()
```
sami-marreed: **6b. `/run` is always mounted and unauthenticated by default**

`/stream` uses `Depends(require_chat_access)`. `/run` does not. The gate is only applied if a token happens to be set. Vanilla CUGA does not set `GATEWAY_TOKEN` or `CUGA_RUN_TOKEN`. Anyone who can reach the port can execute the agent, including when CUGA authentication is enabled and `/stream` is login-gated.

This is a new always-on machine endpoint. It is not behind `EVENTS_API_URL`. I did not find TestClient coverage of the real FastAPI `/run` handler.

Could we require a token whenever authentication is enabled (or always, outside an explicit development flag), and add a test that `/run` returns 401 without the header?

src/cuga/cli/main.py
```diff
@@ -1137,6 +1137,12 @@ def start(
     """
     validate_service(service)
 
+    # NB: there is no --events flag. The event-driven layer is a SEPARATE SERVICE
+    # (`uv run python -m cuga.backend.events.service`, or `make run-events`) that calls this server
+    # over HTTP. `make up` starts both. This command starts CUGA and nothing else.
+    # slow external APIs (arXiv ~5.5s/call + retries) blow the 30s sandbox default
+    os.environ.setdefault("DYNACONF_ADVANCED_FEATURES__SANDBOX_EXECUTION_TIMEOUT", "120")
```
sami-marreed: **7. CLI timeout and `_explicit_mcp`**

These changes are unclear, and they are not limited to events.

This `setdefault` of the sandbox timeout to 120 seconds runs for every `cuga start`. The default in `settings.toml` is still 30 seconds. The arXiv comment may be valid for events demos, but this is still a silent default change for the entire CLI.

`_explicit_mcp` / `_wants_roster` (below) is justified: it stops overwriting `MCP_SERVERS_FILE` with `"none"` when a roster or an exported file is already present.

Could we keep the MCP fix, give it a clearer name (for example `_roster_needs_file_mcp`), and apply the 120-second timeout only for roster or events paths — or change `settings.toml` and document it?

src/frontend_workspaces/frontend/src/api.ts
```diff
@@ -429,6 +466,248 @@ export async function deleteSecret(id: string): Promise<Response> {
...
+  project_grain: string;
+  features: Record<string, boolean>;
+}
+
+// Returns null unless the events layer is actually mounted AND enabled — callers
+// use that to hide the Studio entry point entirely. Robust against the SPA catch-all:
+// when EVENTS_ENABLED is off, /api/events/status is unrouted, so the server returns
+// index.html with HTTP 200. We therefore must NOT trust res.ok alone — we require a
+// JSON content-type AND an explicit enabled:true, or the Studio would leak into vanilla CUGA.
+export async function getEventsStatus(): Promise<EventsStatus | null> {
```
sami-marreed, line 549: **8. Is the UI disabled when events is disabled?**

At runtime, mostly yes. It is not removed from the bundle, and the comments are out of date.

`getEventsStatus` and `EventsGate` treat SPA HTML as “off”, so navigation, `/studio`, and inbox polling stay hidden if the events process is down and `EVENTS_API_URL` is unset. That works.

However:

- Combined mode and `EVENTS_ENABLED` were removed, but comments still refer to `EVENTS_ENABLED`.
- `eventsConnectUrl` (below) still uses `getApiBaseUrl()` (the CUGA origin). In the split topology this pull request is built for, “Connect” opens CUGA, which has no `/api/events/*`, and the SPA catch-all swallows the request. `apiFetch` already uses `getEventsBaseUrl()`; connect should do the same.
- `/api/events/status` always returns `enabled: true` when the process is up, and reports cron, poll, and push as Activepieces-only, even though the native scheduler is the default.
- Studio JavaScript still ships in the committed `frontend/dist` bundle (about 1.3 MB).

src/frontend_workspaces/frontend/src/api.ts
```diff
@@ -429,6 +466,248 @@ export async function deleteSecret(id: string): Promise<Response> {
...
+  return apiFetch("/api/events/connections");
+}
+
+// Where to send the user to log in for an integration (OAuth → redirects to consent).
+// Per-connector setup guides (how to connect, creds present?, ownership options, steps).
+export async function getEventsSetupGuides(): Promise<Response> {
+  return apiFetch("/api/events/setup-guides");
+}
+
+export function eventsConnectUrl(app: string, ownership?: string): string {
```
sami-marreed, line 683: **8b. Split-mode OAuth URL**

`eventsConnectUrl` builds the connect URL from `getApiBaseUrl()` (CUGA). `apiFetch` correctly uses `getEventsBaseUrl()` from `EVENTS_API_URL`. In a split deployment, “Connect” therefore opens CUGA’s origin, which has no `/api/events/*`, and the SPA catch-all swallows it.

Please resolve the events base (the cache from `getEventsBaseUrl`) and build the connect URL from that.

docs/examples/events/supervisor_agents.yaml
```diff
@@ -0,0 +1,93 @@
+# CUGA supervisor — the small, no-AP core roster.
```
sami-marreed, line 1: **9. Roster YAML at the repository root**

Should this file be removed, or moved to `docs/examples/` or `rosters/`?

It is the eight-agent “small no-AP core” roster. `tests/events/test_supervisor_roster.py` hardcodes it as `ROOT/supervisor_agents.yaml`. The 27-agent set is already planned for `rosters/` in 603. The existing convention on `main` is `docs/examples/.../supervisor_*.yaml`.

If 602 merges first, every clone will carry a root-level YAML indefinitely. Could this move before merge, with `CUGA_SUPERVISOR_ROSTER` and the test updated to the new path?

src/cuga/backend/events/native_scheduler.py
```diff
@@ -0,0 +1,192 @@
...
+            log.info("native scheduler retired expired %s", sub.id)
+            continue
+        try:
+            await fire_fn(sub)
+            fired.append(sub.id)
+        except Exception as e:  # noqa: BLE001 — one bad fire must not stall the whole loop
+            # Some transport exceptions (httpx timeouts, cancelled tasks) carry an EMPTY str(), so
+            # "native fire X failed: " told an operator nothing at all. Always name the type.
+            log.warning("native fire %s failed: %s: %s", sub.id, type(e).__name__, e or "(no detail)")
+        nxt = next_fire_after(sub, now)
```
sami-marreed: **A malformed cron expression stalls every later subscription**

`next_fire_after` sits outside the per-subscription `try`. `next_cron` raises `ValueError` if nothing matches within a year, and `_field_matches("*/0", ...)` raises `ZeroDivisionError`. The outer `run_scheduler` catch only keeps the process alive; the next 10-second tick hits the same bad row and skips the rest of `store.due()`. The native scheduler is the default (`EVENTS_SCHEDULER` is not `ap`).

Please move `next_fire_after` inside the per-subscription `try`. On parse or scan failure, pause or delete that row and continue. Reject `step == 0`. Add a `process_due` test with `cron_expr="*/0 * * * *"` placed ahead of a valid subscription.

src/cuga/backend/events/telegram_direct.py
```diff
@@ -0,0 +1,170 @@
...
+    except Exception:  # noqa: BLE001
+        pass
+    while not (stop and stop.is_set()):
+        try:
+            async with httpx.AsyncClient(timeout=35) as c:
+                params = {"timeout": 25, "allowed_updates": '["message"]'}
+                if offset is not None:
+                    params["offset"] = offset
+                r = await c.get(f"{_base()}/getUpdates", params=params)
+                updates = (r.json() or {}).get("result") or [] if r.status_code == 200 else []
```
sami-marreed: **Long-poll hot-loops on non-200 responses**

The sleep exists only on the `except` path. A 401 (revoked token), 409 (webhook still set), or 429 completes the loop with `updates=[]` and retries immediately. That can hammer Telegram and pin a CPU core. `test_events_telegram_direct.py` only covers `should_process` and mention gating.

Please sleep on any non-200 response (honor `retry_after` on 429), log the status, and avoid spinning.

src/cuga/backend/events/events_dashboard.py
```diff
@@ -0,0 +1,244 @@
...
+  <div class="foot">CUGA Events · native scheduler + Activepieces · this page reads only the
+    <code>/api/events/*</code> APIs.</div>
+
+<script>
+const $=s=>document.querySelector(s);
+let RUNFILTER={mode:null,backend:null,status:null};
+let GW=sessionStorage.getItem('gw')||'';
+
+async function j(u,o){try{const r=await fetch(u,o);return await r.json();}catch(e){return null;}}
+function esc(s){return (s||'').replace(/[&<>]/g,c=>({'&':'&amp;','<':'&lt;','>':'&gt;'}[c]));}
```
sami-marreed: **`esc()` does not escape quotes**

The user prompt is interpolated into `title="${esc(s.prompt||'')}"` (line 162). A double quote in the prompt breaks out of the attribute. The connect-callback XSS was fixed and tested; this one was not.

Please also escape `"`, `'`, and backticks, or build the DOM with `textContent` / `setAttribute`.

src/cuga/supervisor_utils/supervisor_config.py
```diff
@@ -182,22 +199,21 @@ async def _create_tool_provider(
...
+    # identifiers downstream ('cuga-finance_get_price' parses as subtraction), so map them.
+    # An agent that names NOTHING gets all tools (unchanged behavior).
     if app_names or mcp_servers:
-        logger.info(
-            f"Creating CombinedToolProvider for apps: {app_names}, MCP servers: {len(mcp_servers) if mcp_servers else 0}"
-        )
-        tool_provider = CombinedToolProvider()
+        scoped = [n.replace("-", "_") for n in app_names] or None
+        logger.info(f"Creating CombinedToolProvider scoped to: {scoped or 'ALL apps'}")
+        tool_provider = CombinedToolProvider(app_names=scoped)
```
sami-marreed: **`CombinedToolProvider` scoping is a silent behavior change for all supervisors**

Previously, `CombinedToolProvider()` loaded every registry app, and the filter was effectively a comment. `CombinedToolProvider(app_names=scoped)` now actually filters, including hyphen-to-underscore mapping. That is a reasonable bug fix, but it changes `CugaSupervisor.from_yaml` for every existing roster, not only events. An agent that listed one MCP server as documentation and relied on the rest of the toolset will now see only that server.

Please note this in the changelog or README, and add a unit test that a named app is the only one initialized. Consider a YAML flag if any in-tree roster depended on the old leak.

sami-marreed (changes requested):
## Review summary

This pull request is not ready to merge.

The split-service boundary is real and well tested. CUGA imports nothing from `cuga.backend.events` (`test_core_decoupled.py`). The combined-mode mount and `EVENTS_ENABLED` are gone. Slash verbs remain ordinary text when `EVENTS_API_URL` is unset. Policy auto-load stays keyword-only for existing callers. That part matches the pull request description.

The merge bar is the packaging questions below, plus the runtime defects called out inline.

### Questions

1. **Generated `events_docs/api` HTML.** Please do not commit these files. They add about 313 KB and exist in git only because the contract `--check` and Studio embed them. Generate them at build or start time, and gitignore the HTML.

2. **`.vscode/launch.json` deleted.** Please restore it in this pull request. This looks like a split artifact, not a product decision. Merging 602 alone breaks F5 debugging until 603.

3. **Documentation location.** `events_docs/` is the wrong home. Setup and architecture belong on docs.cuga.dev The README already links to files that were moved to 603.

4. **`scripts/gen_*.py`.** Partial. `--check` is a real reason to keep generators; `gen_api_spec.py` at about 134 KB is a document in script form. Please move or shrink it once the HTML is no longer committed.

5. **Flat `events/` package.** There are 47 modules and no client/service split. `cuga_door.py` / `HttpRuntime` versus `service.py` / `app.py` should be packaged separately (`api/`, `runtime/`, `channels/`, `store/`).

6. **Growth of `main.py` / the roster.** Please extract `/run`, the roster, and the slash forwarder (same pattern as A2A). Do not add events middleware back onto CUGA. With a roster loaded, `/run` is not the sibling of `/stream`: it prompt-pins the agent and drops resume and attachments. `/run` is also always mounted and unauthenticated unless a token happens to be set, while `/stream` uses `require_chat_access`.

7. **CLI timeout and `_explicit_mcp`.** Partial. Not overwriting `MCP_SERVERS_FILE` is justified. The sandbox `setdefault` of 120 seconds is a silent default for every `cuga start`.

8. **Studio and inbox when events is down.** Partial. `getEventsStatus` and `EventsGate` work. `eventsConnectUrl` still targets the CUGA origin. Comments still mention `EVENTS_ENABLED`. Status reports cron, poll, and push as Activepieces-only. Studio JavaScript still ships in `frontend/dist`.

9. **`supervisor_agents.yaml` at the repository root.** Please move it to `docs/examples/events/` or `rosters/` before merge. Tests hardcode the root path.

### Additional blockers

- **Native cron.** `next_fire_after` sits outside the per-subscription `try` and can stall every later subscription (`*/0` raises `ZeroDivisionError`). The native scheduler is the default.
- **Telegram long-poll.** Non-200 responses (401, 409, 429) retry immediately with no sleep.
- **Dashboard `esc()`.** Quotes are not escaped, which allows attribute XSS on user prompts.
- **`CombinedToolProvider` scoping.** Filtering is now real for all supervisors, not only events.

### Request

Please restore `launch.json`. Please do not merge generated HTML or a roster YAML at the repository root. Please extract `/run` from `main.py` and lock down its authentication. Please fix the cron and Telegram loops. Please do not land packaging that 603 is meant to clean up; if 602 merges first, that mess lands on `main`.

The connect-callback XSS reported by CodeRabbit is already fixed. There is no need to reopen it.

anupamamurthi:
Thank you, Sami. I will take a look today :)

src/cuga/backend/server/main.py
anupamamurthi: Hi Sami, Fixed it (will commit in a bit) but adding a note as we may want a follow up later.

 /run and /run/agents now require a token instead of skipping the check when none is set. No token → 401. There's an escape hatch for local dev (CUGA_RUN_ALLOW_UNAUTHENTICATED=1) that logs a warning at startup. I am just using a shared secret for now.

Anyone running locally without a token will start getting 401s.

On the bigger point — a shared secret only proves someone knows the string. No identity, no expiry, no rotation.

Looks like get_current_user already accepts a Bearer token and validates it against the issuer, so the events service could use client credentials and /run could check a JWT the same way /stream does, with the shared secret as fallback where there's no issuer.


I'd rather not add it into this PR. Happy to open a follow-up next to the identity work if that sounds right.

src/cuga/backend/events/__init__.py
anupamamurthi: I like the suggestion but trying to not rock the boat I would prefer to make this change as a follow up - more as as a refactor - given how this change touches EVENTING module alone.
Happy to create a little issue to track this.

README.md
anupamamurthi: Good point, Sami

**Done:**

Fixed the broken README links — both files are on 603, so they now point at events_docs/setup/
Removed three dead Studio doc tabs
Cleared all seven EVENTS_ENABLED references from the SPA
Split Studio into its own chunk; main bundle is 50 KB smaller

**Next, on 603 rather than here:**
Move events_docs/setup/*.md under docs/ — 602 has 14 files in that tree, 603 has 44, so moving here just collides
Keep the API HTML with the app rather than in docs. It's Studio UI served by /api/events/docs/{page}, and api.html is load-bearing — a test fails if a route is missing from it
docs.cuga.dev is a separate pipeline, so its own PR

anupamamurthi:
> ## Review summary
> This pull request is not ready to merge.
> 
> The split-service boundary is real and well tested. CUGA imports nothing from `cuga.backend.events` (`test_core_decoupled.py`). The combined-mode mount and `EVENTS_ENABLED` are gone. Slash verbs remain ordinary text when `EVENTS_API_URL` is unset. Policy auto-load stays keyword-only for existing callers. That part matches the pull request description.
> 
> The merge bar is the packaging questions below, plus the runtime defects called out inline.
> 
> ### Questions
> 1. **Generated `events_docs/api` HTML.** Please do not commit these files. They add about 313 KB and exist in git only because the contract `--check` and Studio embed them. Generate them at build or start time, and gitignore the HTML.
> 2. **`.vscode/launch.json` deleted.** Please restore it in this pull request. This looks like a split artifact, not a product decision. Merging 602 alone breaks F5 debugging until 603.
> 3. **Documentation location.** `events_docs/` is the wrong home. Setup and architecture belong on docs.cuga.dev The README already links to files that were moved to 603.
> 4. **`scripts/gen_*.py`.** Partial. `--check` is a real reason to keep generators; `gen_api_spec.py` at about 134 KB is a document in script form. Please move or shrink it once the HTML is no longer committed.
> 5. **Flat `events/` package.** There are 47 modules and no client/service split. `cuga_door.py` / `HttpRuntime` versus `service.py` / `app.py` should be packaged separately (`api/`, `runtime/`, `channels/`, `store/`).
> 6. **Growth of `main.py` / the roster.** Please extract `/run`, the roster, and the slash forwarder (same pattern as A2A). Do not add events middleware back onto CUGA. With a roster loaded, `/run` is not the sibling of `/stream`: it prompt-pins the agent and drops resume and attachments. `/run` is also always mounted and unauthenticated unless a token happens to be set, while `/stream` uses `require_chat_access`.
> 7. **CLI timeout and `_explicit_mcp`.** Partial. Not overwriting `MCP_SERVERS_FILE` is justified. The sandbox `setdefault` of 120 seconds is a silent default for every `cuga start`.
> 8. **Studio and inbox when events is down.** Partial. `getEventsStatus` and `EventsGate` work. `eventsConnectUrl` still targets the CUGA origin. Comments still mention `EVENTS_ENABLED`. Status reports cron, poll, and push as Activepieces-only. Studio JavaScript still ships in `frontend/dist`.
> 9. **`supervisor_agents.yaml` at the repository root.** Please move it to `docs/examples/events/` or `rosters/` before merge. Tests hardcode the root path.
> 
> ### Additional blockers
> * **Native cron.** `next_fire_after` sits outside the per-subscription `try` and can stall every later subscription (`*/0` raises `ZeroDivisionError`). The native scheduler is the default.
> * **Telegram long-poll.** Non-200 responses (401, 409, 429) retry immediately with no sleep.
> * **Dashboard `esc()`.** Quotes are not escaped, which allows attribute XSS on user prompts.
> * **`CombinedToolProvider` scoping.** Filtering is now real for all supervisors, not only events.
> 
> ### Request
> Please restore `launch.json`. Please do not merge generated HTML or a roster YAML at the repository root. Please extract `/run` from `main.py` and lock down its authentication. Please fix the cron and Telegram loops. Please do not land packaging that 603 is meant to clean up; if 602 merges first, that mess lands on `main`.
> 
> The connect-callback XSS reported by CodeRabbit is already fixed. There is no need to reopen it.

All four blockers fixed
Native cron — next_fire_after moved inside the per-subscription try; an unschedulable row is retired instead of stalling every row behind it. Also rejects step <= 0 at parse time rather than exploding in a tick, and refuses an unsatisfiable cron at arm time with a message instead of a stack trace.
Telegram — sleeps on any non-200 and honours retry_after on 429. Also retains handler tasks; a bare create_task could be GC'd mid-flight and lose a message silently.
Dashboard esc() — escapes ", ' and backticks now, and the row actions use data- attributes with a delegated listener instead of interpolated onclick.
CombinedToolProvider — unit test that a named app is the only one initialised, plus a README note. I checked every roster on main: each one that names apps names exactly its own, so nothing depended on the old leak and a compat flag would be dead config.

On your numbered points:

1, 4 — deleted api_spec.html and gen_api_spec.py outright rather than moving them (−5,674 lines). Worth flagging: only api_spec.html was generated. api.html is hand-written and load-bearing (a test fails if a route is missing from it), and examples.html is hand-written with one generated data block — deleting all three would have lost authored content.

2 — launch.json restored, byte-identical to main.

3 — fixed the broken links. Chasing them turned up two more dead Studio tabs: slides.html exists on no branch, and the NL→Flow page is runbook/nl-to-flow.html — different folder and spelling, so it never worked on 603 either. The tree move I'd do on 603, where 44 of the 58 files live.

6 — extracted into run_routes.py + events_bridge.py, same shape as A2A. main.py is back to 4,417 from 4,949, within 59 lines of main. Mounting is conditional too, so vanilla CUGA has no /run at all — a 404, not a 401.

6b — fails closed now. Token required, one explicit dev flag, logged loudly at boot. Five TestClient tests. Also documented it: .env.example had no token at all and nothing said how to generate one — the 401 itself now carries the command.
 
7 — kept the MCP fix, renamed to _roster_needs_file_mcp as you suggested. Removed the 120s timeout from the CLI entirely rather than scoping it: it exists because arXiv is slow in one demo, which makes it deployment config, not CLI behaviour. Documented in .env.example.

8, 8b — connect URL now resolves the events base (sync, from the cache — awaiting first would break window.open's user-gesture and get popup-blocked). All seven EVENTS_ENABLED references gone. Status reports cron/poll from the native scheduler instead of AP-only, and splits push from push_direct. Studio is a lazy chunk now, so it's out of the main bundle.

9 — roster moved to docs/examples/events/, matching the convention on main. Not rosters/, since that directory only exists on 603 and using it here would define a layout 603 owns.

Not doing in this PR:

5 (flat events/ package) — agreed in principle, but it touches every import and this PR is already the thing you're asking me not to grow.

CodeRabbit's two pytest-marker suggestions — 0 of 57 files in tests/events use markers, so adding two would make pytest -m unit return exactly those two.

Five CodeQL ReDoS alerts (arming.py ×3, classify.py ×2) I'll dismiss as false positives. Measured linear — ~2.0× per doubling out to 16k chars — because both patterns run on _one_space()-normalised text, so the repeated-whitespace input the analysis models can't occur. The concierge.py ones were a real O(n²) though (662ms → 34ms at 64k) and are fixed.

I ran each suite the way CI invokes it: lint clean repo-wide, 411 events tests, plus the unit groups, a2a, policy, sdk_core and manager API. Also started both processes and confirmed the /run hop works end to end.

Two things land on 603 and will break its build until fixed — Dockerfile.events copying supervisor_agents.yaml, and the api-spec Makefile target. I'll handle those when I re-stack it.

src/cuga/backend/server/main.py
```diff
@@ -2074,7 +2097,17 @@ async def ui_config():
     """Return UI configuration flags from settings."""
     hide_logo = settings.ui.hide_cuga_logo
     brand_name = getattr(settings.ui, "brand_name", "CUGA Agent") or "CUGA Agent"
-    return JSONResponse({"hide_cuga_logo": hide_logo, "brand_name": brand_name})
+    # SPLIT TOPOLOGY: this server may serve the UI while the eventing layer runs as its own
```
anupamamurthi, line 2165: ### `src/cuga/backend/server/main.py`  (+76 / −1)

CUGA's FastAPI app. Five touch points, all of them seams rather than logic:

1. **lifespan** — a comment recording that no event loops start here any more. The Telegram
   long-poll, Discord Gateway and native scheduler belong to the eventing service's own lifespan.
   Also restores an ephemeral-row cleanup call that the merge silently dropped.
2. **`import events_bridge`** — the only reference in this file to the eventing layer, and it is a
   sibling module, not the events package.
3. **`/api/ui/config`** — now returns `events_api_url`. In a split deployment the SPA would
   otherwise assume one origin and send its Studio calls here, where `/api/events/*` doesn't exist.
   Empty by default, which means same-origin, so nothing changes when unset.
4. **`/stream`** — edge dispatch. An explicit slash verb, or a thread with an open arming dialogue,
   is forwarded to the eventing service over HTTP; everything else goes straight to the agent as
   before. This used to call the concierge in-process.
5. **the mount** — `include_router(build_run_router(...))`, guarded by `run_api_enabled()`, the same
   shape as A2A. With no token and no roster configured, `/run` is never mounted at all.

src/cuga/cli/main.py
```diff
@@ -1343,7 +1343,10 @@ def start(
         os.environ["CUGA_DEMO_ADVANCED"] = "true"
         os.environ["CUGA_MANAGER_MODE"] = "true"
         os.environ["DYNACONF_POLICY__FILESYSTEM_SYNC"] = "false"
-        os.environ["MCP_SERVERS_FILE"] = "none"
+        # An operator who exports MCP_SERVERS_FILE means it — don't overwrite it. Unset still means
+        # "none" (managed-config-db mode, serving the demo app), which is what the demo wants.
+        if not (os.environ.get("MCP_SERVERS_FILE", "") or "").strip():
+            os.environ["MCP_SERVERS_FILE"] = "none"
```
anupamamurthi, line 1371: ### `src/cuga/cli/main.py`  (+4 / −1)

One guard in `cuga start demo`. It used to do `os.environ["MCP_SERVERS_FILE"] = "none"`
unconditionally, overwriting a path the operator had exported and leaving the registry serving only
the demo app — so every roster agent answered "the available toolset does not include…". Now it only
defaults when the variable is unset. Nothing else in the CLI changes, and it mentions nothing about
events.

src/cuga/backend/server/run_routes.py
```diff
@@ -0,0 +1,505 @@
+"""``POST /run`` and ``GET /run/agents`` — CUGA's machine seam, plus the supervisor roster.
```
anupamamurthi, line 1: ### `src/cuga/backend/server/run_routes.py`  (new, 505 lines)

`POST /run` and `GET /run/agents` — CUGA's machine seam — plus the supervisor roster. Extracted from
`main.py` rather than grown inside it, in the same shape as `a2a.runner.build_a2a_router_for_settings`:
this module owns the routes and their state, `main.py` keeps one `include_router` call.

`/run` is the non-streaming form of `/stream`: one request, one JSON answer. Without a roster it
drives the same `event_stream` generator in-process, so knowledge, history, citations and policies
behave identically. With a roster it calls `supervisor.invoke`, which takes no attachments, resume
or HITL, and `?agent=` is advisory — the docstring says so rather than letting someone find out.

Authenticated with a shared secret (`CUGA_RUN_TOKEN`, else `GATEWAY_TOKEN`) and fails closed: no
token, no route. The auth check is the first statement in the handler, before the lazy imports,
because those pull in `cuga.config`, whose `load_dotenv` can refill an absent variable mid-request
and change the gate's answer.

src/cuga/backend/server/events_bridge.py
```diff
@@ -0,0 +1,142 @@
+"""The slash-command bridge: CUGA core → the eventing service, over plain HTTP.
```
anupamamurthi, line 1: ### `src/cuga/backend/server/events_bridge.py`  (new, 142 lines)

The slash-command bridge: CUGA core → the eventing service, over plain HTTP. The only module in core
that knows the eventing layer exists, and all it knows is a URL — it detects `/automate …` and POSTs
it. No shared database, no shared objects, no bot tokens, and inert when `EVENTS_API_URL` is unset,
which is vanilla CUGA. Used by `/stream` (web chat) and `/run` (a channel forwarding an utterance).

It imports nothing from `cuga.backend.events`, and neither does any file above.
`tests/events/test_core_decoupled.py` parses all five with `ast` and fails the build if that changes.

docs/examples/events/supervisor_agents.yaml
anupamamurthi, line 1: fyi: in my follow up, I am moving this file into events/examples (so that we are ready to move this module out into it's repo, if it comes to it!)

scripts/gen_examples.py
```diff
@@ -0,0 +1,74 @@
...
+truth, src/cuga/backend/events/catalog.py. The surrounding HTML/CSS/JS is hand-maintained; only the
+data array is spliced in, so the doc board can never drift from the Studio Examples tab again.
+
+    python scripts/gen_examples.py            # rewrite in place
+    python scripts/gen_examples.py --check    # exit 1 if out of date (used by the consistency test)
+
+The catalog carries extra keys (`outcome`, `star`) the board also uses; we emit the full record so the
+board and the Studio stay byte-identical on the shared fields.
+"""
+
```
anupamamurthi: fyi: in my follow up, I am moving this file into events/scripts (so that we are ready to move this module out into it's repo, if it comes to it!)

events_docs/setup/DISCORD.md
```diff
@@ -0,0 +1,94 @@
...
+> `EVENTS_DISCORD_BACKEND=ap`. It watches one channel with `DedupeStrategy.TIMEBASED`. Prefer the
+> direct Gateway (default) — instant, no public URL, native author id.
+
+## What you'll need
+- A Discord server where you can add a bot, with **MESSAGE CONTENT INTENT** enabled.
+- Just the bot token — **no Activepieces, no public URL** for the direct backend.
+
+## Steps
+
+1. **Create the app + bot** — <https://discord.com/developers/applications> → **New Application** →
```
anupamamurthi: fyi: in my follow up, I am moving this file into events/docs (so that we are ready to move this module out into it's repo, if it comes to it!)

I will also reword the documentation and carefully move them out into the docs repo in upcoming PRs.

anupamamurthi:
Hi @sami-marreed To the best of my knowledge, I don't see issues with this PR and everything continues to work as-is. If the Eventing service is started, eventing studio will be visible and one can say things like "/automate... " 

I have manually tested a local set up (and ofcourse with PR 603 in place, I have also deployed this to code engine - pointers were shared in our slack thread) - as usual, once we get "cuga start demo" up and running (with the usual ENV variables) and that doesnt change CUGA's behavior. Feel free to try it to be sure

ofc, if eventing service is up, then one can start using "/automate" command. [`uv run python -m cuga.backend.events.service`]
The below variables are needed

EVENTS_API_URL=http://127.0.0.1:8100
CUGA_URL=http://127.0.0.1:7860
EVENTS_CORS_ORIGINS=http://localhost:7860,http://127.0.0.1:7860
GATEWAY_TOKEN=<>
EVENTS_DB=/Users/anu/.cuga/events.db

# in order to use the example I have in place
CUGA_SUPERVISOR_ROSTER=docs/examples/events/supervisor_agents.yaml
MCP_SERVERS_FILE=src/cuga/backend/tools_env/registry/config/mcp_servers_cuga_apps.yaml

https://github.com/cuga-project/cuga-agent/blob/event_support/docs/examples/events/supervisor_agents.yaml --->>> is just an example and one can use a different supervisor + subagents.
The implementation of this sits in cuga events bridge
https://github.com/cuga-project/cuga-agent/blob/event_support/src/cuga/backend/server/events_bridge.py
https://github.com/cuga-project/cuga-agent/blob/event_support/src/cuga/backend/server/run_routes.py

we can change this as we go.. pretty sure the other work we discussed (around adding new sub agents etc) will influence this set up

haroldship:
Hi @anupamamurthi — a heads-up on a secret-scanning alert that traces back to this branch. Nothing to change here, just so it isn't a surprise.

GitHub opened [alert #3](https://github.com/cuga-project/cuga-agent/security/secret-scanning/3) (AI-detected generic password) on `AP_PASSWORD=Sup3rSecret!` in `roadmap/setup.html`, introduced in `f613e643` and also added to `.env.events.example` in `7dd9531f` — both commits are in this branch's history. The tip here is clean: `roadmap/setup.html` was deleted in `70776ab2`, and `.env.events.example` moved to the follow-up PR in `6ef2208b`. The value is still live on #603 and #605, so I've left a note there too.

I looked into it and it's a documentation placeholder rather than a real credential — `AP_PASSWORD` is read from the environment everywhere in the code — so there's nothing to rotate, and I'm resolving the alert as a false positive.

One thing that might help going forward: the repo's pinned `detect-secrets` hook does flag this line, so if `pre-commit install` hasn't been run in your clone, that would catch this class of thing before it lands. Worth knowing that a bare `detect-secrets scan` from the shell often picks up the Yelp 1.5.0 build, which misses unquoted `PASSWORD=` assignments — the repo pins `ibm/detect-secrets 0.13.1+ibm.64.dss`, which catches them. `pre-commit run detect-secrets --all-files` uses the pinned one.

anupamamurthi:
Hi @sami-marreed 

1. Docs. events_docs/ is out of the PR; I will create a new PR in the docs repo soon
2. CombinedToolProvider - done
3. build.sh. Done

4. api.html / examples.html. removed
5. #433 / roster -> (AI generated note ) ---->>>> Dropped my opt-in flag since you made scoping the default. The only thing I kept that #433 doesn't cover is mcp_servers: counting as a named app — the roster declares only mcp_servers, so otherwise every specialist gets the whole registry.

I also had a hyphen→underscore mapping in _create_tool_provider and removed it. mcp_manager stores an MCP server's YAML key verbatim, so rewriting names in transit scoped an operator whose server really is my-server to a key the registry doesn't have — no error, one log warning, an agent with no tools. I renamed the roster to cuga_finance instead, so there's nothing to translate and core passes names through untouched.

Worth noting these were never two mechanisms — build_agents_from_stored_subagents ends by calling build_agents_from_list. Same builder, two sources. Collapsed it so the store is the only runtime source and the YAML is seeded at startup; the roster's agents now show in Manage as a side effect.

6. UI rebuild done

7. JWT on /run. done to the best of my knowledge

Since you reviewed — four changes, two of them security-shaped:

Core diff is three files: server/main.py (+94, −0, every seam gated), cli/main.py (3 lines above), supervisor_utils/supervisor_config.py.

I will rebase 603 so that all the deployment related changes and any additional bug fixes go there.

sami-marreed (approved):

PR #603, opened 2026-08-07
feat(events & channels): Events followup tooling, deploy kit, rosters

anupamamurthi:
## Summary

The second half of the event-driven agents work, split out so the feature PR could be reviewed at
all — CodeRabbit skips above 150 files and the combined change was 218.

Nothing here is imported by the product or the CI gate. PR 1 was verified green with all of it
removed: `ruff check` + `format --check` clean, `pytest tests/events` 393 passed.

**73 files:**

| | |
|---|---|
| 30 | `events_docs/` — ARCHITECTURE, the automation cookbook, runbook pages, plans, and the setup guides not pinned by a contract test |
| 14 | `rosters/` — drop-in example rosters (domain families, plus an AP / no-AP split for enterprise test beds) |
| 13 | `scripts/` — `make` targets, generators, live-test runners |
| 12 | `deploy/` — the Code Engine kit: build, deploy, managed-PostgreSQL provisioning, smoke test |
| 4 | `Makefile`, `.env.events.example`, `roadmap/`, `.vscode/` |

## Why these were the ones to move

The split was chosen by dependency, not by convenience — I removed candidates and re-ran the CI gate
to confirm each was genuinely inert:

- **`gen_api_spec.py` and `gen_examples.py` stayed in PR 1**, because `test_events_api_contract.py`
  shells out to both with `--check`. Deferring them breaks the build.
- **11 of the 22 `events_docs/setup/*.md` stayed in PR 1**, because
  `test_every_trigger_appears_in_its_setup_guide` requires one `<APP>.md` per registry app.
- **`events_docs/api/*` stayed in PR 1** — three contract tests read those files.
- **The `Makefile` ships here with the scripts it drives**, rather than sitting in PR 1 referencing
  targets whose scripts had moved. A missing Makefile is honest; a half-broken one is not.

## What's in it worth reviewing

**The Code Engine kit (`deploy/ce/`)** is the only part with real logic: an idempotent build/deploy
pair, managed-PostgreSQL provisioning (capturing the DSN at create, because the account redacts
credentials on read), a CA-pinned `sslmode=verify-full` connection, and post-deploy assertions that
the roster loaded and storage is durable.

**The rosters** are configuration only — alternative sub-agent line-ups you can point
`EVENTS_SUPERVISOR_ROSTER` at. Two cuts: domain families (`box_document_intelligence`,
`repository_intelligence`, `market_research_intelligence`, …) and an enterprise test bed split by AP
dependency — `no_ap_*` run with zero Activepieces, `ap_*` need it for SaaS push triggers.

**The docs** include the architecture diagrams (a master flowchart and one sequence diagram per
channel and trigger), the automation cookbook (what to actually type, per tier), the deployment
runbook, and a security note on the deployed service's auth posture that needs a decision.

----

Updates summary after rebasing on approved-602
The five themes
1. Security — the gates now fail closed (baeee92f, Sami A/B/D/E)
Every gate was written as "enforce if configured," which reads as safe and is the opposite: an unset secret disabled the check and served anyway. On the live deployment EVENTS_WEBHOOK_KEY was unset, so POST /api/events/hook/<name> ran an agent for anyone with the URL. One switch (EVENTS_ALLOW_UNAUTHENTICATED) now mirrors the existing CUGA_RUN_ALLOW_UNAUTHENTICATED. Separately, /run read body["user_id"] unconditionally — a signed-in user could act as anyone; the authenticated principal wins now.
2. Deploy kit — events/deploy/ (12 files), make ce-*, split topology cuga-core + cuga-events-svc, managed Postgres, and a preflight that refuses to deploy an incomplete secret.
3. Persistence & provenance (0b84faea, 8d722e4c) — Studio-created agents survive an instance replace; source: studio|seed keeps generated roster rows distinguishable from human-made ones, so "Manage" edits aren't overwritten on restart.
4. Openness fixes (06d65a1a, a4021621) — the MCP catalog became a default rather than an allow-list (you can bring your own server), with on-read migration for the pre-rename cuga-web → cuga_web names.
5. Docs — ADR 0009 (identity/credential grain: a channel is an auth authority, so isolation is always on and linking is an upgrade), the docs tree moved out of the repo, and the secrets/generation/rotation material.

File | Change
-- | --
backend/server/run_routes.py | +37 / −1 — the authenticated-principal fix
supervisor_utils/roster_seed.py | +114 / −11 — provenance + don't-clobber-edits

 

## Summary by CodeRabbit

* **New Features**
  * Added local and IBM Cloud deployment workflows with split services, durable PostgreSQL storage, OIDC login, smoke tests, and operational commands.
  * Added direct chat channels, authenticated webhooks, OAuth integrations, channel controls, event inbox updates, and Studio support for custom agents and MCP servers.
  * Added reusable example rosters and automated testing/reporting tools.

* **Bug Fixes**
  * Improved authentication, CORS handling, roster preservation, Slack event processing, polling persistence, flow uniqueness, and OAuth secret protection.

* **Documentation**
  * Added environment templates, deployment guidance, roster documentation, and troubleshooting instructions.


src/cuga/frontend/dist/vendors.4376dde5d8e0c2872516.js
anupamamurthi, line 1: this JS file seems to be +424,289!!!! ideas on fixing?

haroldship:
Hi @anupamamurthi — a heads-up on a secret-scanning alert that traces back to this branch, no action urgently needed.

GitHub opened [alert #3](https://github.com/cuga-project/cuga-agent/security/secret-scanning/3) (AI-detected generic password) on `AP_PASSWORD=Sup3rSecret!`, first seen in `roadmap/setup.html` in `f613e643`. The same value is still live on this branch in `.env.events.example:104`.

I looked into it and it's a documentation placeholder rather than a real credential — the surrounding comment tells the reader to invent their own, and `AP_PASSWORD` is read from the environment everywhere in the code — so there's nothing to rotate, and I'm resolving the alert as a false positive. It does keep the scanner interested though, so it may be worth swapping for a placeholder that doesn't look like a password (something like `AP_PASSWORD=<invent-a-strong-password>`, matching the `<...>` style already used for the bot tokens just above) before this merges.

One thing that might help going forward: the repo's pinned `detect-secrets` hook does flag this line, so if `pre-commit install` hasn't been run in your clone, that would catch this class of thing before it lands. Worth knowing that a bare `detect-secrets scan` from the shell often picks up the Yelp 1.5.0 build, which misses unquoted `PASSWORD=` assignments — the repo pins `ibm/detect-secrets 0.13.1+ibm.64.dss`, which catches them. `pre-commit run detect-secrets --all-files` uses the pinned one.

Happy to push the one-line change myself if that's easier.

events/deploy/4_postgres.sh
```diff
@@ -0,0 +1,198 @@
...
+d = json.load(sys.stdin)
+d = d[0] if isinstance(d, list) and d else d
+c = d.get("credentials") or {}
+# Databases for PostgreSQL exposes connection info under connection.postgres
+pg = ((c.get("connection") or {}).get("postgres") or {})
+comp = pg.get("composed") or []
+if comp:
+    dsn = comp[0]
+else:
+    # fall back to assembling it from the parts
```
anupamamurthi, line 111: I am creating a new docker image on purpose so that we keep this separate from all that is in "deployment"

events/deploy/1_build_push_image.sh
```diff
@@ -0,0 +1,73 @@
...
+  echo "ABORT: .env.ce leaked into the build context. Refusing to upload secrets."; exit 1
+fi
+CTX_SIZE=$(du -sh "$STAGE" 2>/dev/null | cut -f1)
+echo "Build context size: ${CTX_SIZE:-?}"
+
+uuid=$(uuidgen | tr '[:upper:]' '[:lower:]' | awk -F- '{print $1}')
+BUILD_NAME="cuga-events-build-${uuid}"
+
+echo "Submitting buildrun '$BUILD_NAME' -> $IMAGE_REF (uv sync is heavy; ~10-20 min) ..."
+ibmcloud ce buildrun submit \
```
anupamamurthi, line 56: The deployment targets CE for now. We can expand further as we see fit.

events/deploy/1_build_push_image.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
+# ============================================================
```
anupamamurthi, line 2: technically, if we use a different docker registry, we can
docker build && push to any OCI registry

events/deploy/3_state_store.sh
```diff
@@ -0,0 +1,141 @@
...
+else
+  echo "→ creating bucket '$BUCKET' ..."
+  ibmcloud cos bucket-create --bucket "$BUCKET" --ibm-service-instance-id "$COS_GUID" \
+    --region "$BUCKET_REGION" --class smart
+  echo "✓ bucket created"
+fi
+
+# ---- 3+4. HMAC service credentials → Code Engine secret ----
+# CAPTURE THE KEYS AT CREATION. This account redacts credentials on read: both
+# `ibmcloud resource service-key <name> --output json` and the resource-controller REST API return
```
anupamamurthi, line 81: additional tables in PG for the eventing work

Image

haroldship:
Secret-scanning alert: https://github.com/cuga-project/cuga-agent/security/secret-scanning/3

The alert's flagged location (`roadmap/setup.html`, an old doc that's since been deleted) never touched `main` — but this branch carries the same literal placeholder value in `.env.events.example`:

```
AP_PASSWORD=Sup3rSecret!                                         # ≥8 chars, mixed case + number + symbol
```

It's not a live credential (it's an example value readers are told to invent), but it's still a literal string that reads as a real password, and it'll get flagged again once this branch merges. Suggest swapping it for an instructional placeholder, matching every other secret-shaped field in that file, e.g.:

```
AP_PASSWORD=<invent-a-strong-password>                           # ≥8 chars, mixed case + number + symbol
```

Separate, unrelated issue found while checking this file: the `detect-secrets` pre-commit hook currently fails on `.env.events.example` regardless of the above, on line 84 (`SLACK_SIGNING_SECRET=`, empty/commented-out value) — a false positive that blocks the hook for any edit to this file. Worth allowlisting with `# pragma: allowlist secret` in the same pass.

src/cuga/backend/events/service.py
```diff
@@ -147,6 +147,27 @@ def create_app():
...
+    #
+    # Identity only, no password: roles are what the gates read, and the local-login path is
+    # separate. This grants nothing that the header-based principal model does not already grant —
+    # see decisions/0009, which is where wiring real auth is tracked.
+    _admins = [
+        u.strip() for u in (os.environ.get("EVENTS_ADMIN_USERS", "admin") or "").split(",") if u.strip()
+    ]
+    for _uid in _admins:
+        try:
+            if users.get(_uid, "default") is None:
```
sami-marreed, line 165: [P1] Authenticate admin requests before granting the default identity admin privileges

This bootstrap makes the default `admin` identity privileged, but `/api/events/admin/*` only checks `_is_admin()` against an identity resolved from caller-controlled scope/headers or `EVENTS_USER_ID`. Those routes do not check the gateway token or authenticate the principal. With a configured `GATEWAY_TOKEN` and `EVENTS_ALLOW_UNAUTHENTICATED=0`, I reproduced an unauthenticated `POST /api/events/admin/users` creating another admin with HTTP 200. The same authorization path protects OAuth configuration and connector credential changes. Require authentication on these routes and derive the privileged principal from verified identity before consulting its roles.

src/cuga/backend/events/concierge.py
```diff
@@ -659,13 +659,33 @@ async def find_or_create_flow(
...
+            # `_utterance`, a ContextVar the react-agent does not reliably propagate into tool
+            # execution (the same caveat documented at the poll-tier selection below) — so two
+            # different requests can hash identically, which is exactly a collision.
+            #
+            # Leaving dedup_key EMPTY (rather than deleting the column) is what actually disables
+            # this: the store's UNIQUE index is partial — `WHERE dedup_key != ''` — so an empty key
+            # cannot collide, and `upsert` can never raise DuplicateSubscription. Flip the flag on
+            # and both the lookup and the index constraint come back exactly as before.
+            if os.environ.get("EVENTS_FLOW_REUSE", "0").split(" #", 1)[0].strip() != "1":
+                dedup_key = ""
```
sami-marreed, line 680: [P1] Keep push-flow names unique when flow reuse is disabled

The AP push path later computes its flow-name suffix as `sha1(dedup_key.encode()).hexdigest()[:8]`. Clearing this value makes that suffix `da39a3ee` for every new watch. I reproduced watches for two different GitHub repositories under the same scope both receiving `push-github-new-pr-cuga-da39a3ee`. `APEngine._new_flow()` deletes the existing same-named flow before creating its replacement, so the second arm destroys the first watch's AP flow while leaving its subscription behind. Generate a unique subscription identifier before flow creation and use it for the AP flow name independently of the optional reuse key.

events/deploy/1_build_push_image.sh
```diff
@@ -0,0 +1,73 @@
...
+  --exclude '.venv*' --exclude 'venv/' \
+  --exclude 'node_modules/' \
+  --exclude '__pycache__/' --exclude '*.pyc' --exclude '.pytest_cache/' \
+  --exclude 'src/frontend_workspaces/' \
+  --exclude 'src/cuga/logging/trajectory_data/' \
+  --exclude '*.egg-info/' \
+  --exclude '*.js.map' \
+  --exclude '*.db' --exclude 'events.db' \
+  --exclude '/output/' --exclude '/results/' \
+  --exclude 'events/deploy/.env.ce' \
```
sami-marreed, line 40: [P1] Exclude the generated core secret file from the uploaded build context

`make_env_ce.sh` also generates `events/deploy/.env.ce.core`, containing gateway, LLM, and database credentials. This rsync exclusion and the following safety check only cover `.env.ce`; `.gitignore` does not filter rsync, and `.dockerignore` does not exclude `.env.ce.core`. Running this exact staging command against dummy files leaves `.env.ce.core` in the staged source that is submitted to the cloud builder. Exclude both generated secret files and their backup variants, and check that none are present before upload. This finding concerns uploaded build source, not a claim that the file is copied into the final image.

events/deploy/4_postgres.sh
```diff
@@ -0,0 +1,198 @@
...
+# ---- 3. persist the DSN + CA: Code Engine secret AND .env.ce ----
+# BOTH, deliberately. The secret is what the running apps read; .env.ce is what a from-scratch
+# deploy REBUILDS the secret from. Writing only the secret is how these two keys used to exist in
+# exactly one place in the world — so `teardown.sh WIPE_SECRET=1`, or a deploy into a fresh
+# project, destroyed them with nothing to restore from, and 2_deploy.sh then silently fell back
+# to ephemeral SQLite and exited 0.
+# `secret update` MERGES — it preserves the other keys (bot tokens, watsonx, GATEWAY_TOKEN), so
+# do NOT recreate.
+echo "→ writing EVENTS_DB into Code Engine secret '$SECRET_NAME' ..."
+ibmcloud ce secret update --name "$SECRET_NAME" --from-literal "EVENTS_DB=${DSN}" >/dev/null
```
sami-marreed, line 145: [P2] Persist the provisioned credentials before updating a potentially missing CE secret

The documented fresh-project sequence runs this script before `2_deploy.sh`, but the latter is what creates `$SECRET_NAME`. This unconditional `secret update` therefore fails when that secret does not exist. Because `set -euo pipefail` is enabled, execution stops before the DSN and CA are written to `$ENV_CE_FILE`, despite the database and service credentials already having been provisioned. The script also documents that credentials are redacted on later reads. Persist the create response locally first, then create or update the CE secret according to whether it exists.

sami-marreed (commented):
Changes are needed before merge. Reviewed commit `858aff4ebf7f924512ec6ba65410c22435a784ac`.

| Priority | Finding |
|---|---|
| P1 | Unauthenticated callers can create admins through the events admin API. |
| P1 | Distinct AP push watches receive the same flow name when reuse is disabled, replacing existing flows. |
| P1 | The generated `.env.ce.core` credentials file is included in the cloud build source. |
| P2 | Fresh PostgreSQL provisioning can fail before saving the newly created credentials. |
| P2 | Teardown targets the retired app name and leaves both deployed apps running. |
| P2 | The development PostgreSQL port is published beyond localhost with fixed development credentials. |

Details and suggested corrections are in the inline comments. Several findings match existing unresolved CodeRabbit comments; these were checked independently.

Validation: 40 focused tests passed. Two additional local reproduction tests demonstrated the admin authentication bypass and the push-flow name collision. Teardown was checked with a mocked CLI; build staging was checked with dummy credentials. No live cloud resources were changed, and the full test suite was not run.

events/deploy/teardown.sh
```diff
@@ -0,0 +1,28 @@
...
+# Leaves the registry image + the shared registry secret intact.
+# ============================================================
+set -euo pipefail
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+source "${SCRIPT_DIR}/config.sh"
+admin_guard "${1:-}"
+require_login
+ce_target
+
+if ibmcloud ce app get -n "$APP_NAME" >/dev/null 2>&1; then
```
sami-marreed, line 15: [P2] Tear down both apps created by the split deployment

`config.sh` defaults `APP_NAME` to `cuga-events`, while `2_deploy.sh` creates `CORE_APP=cuga-core` and `EVENTS_APP=cuga-events-svc`. With these defaults, teardown looks up an app that was never created, prints `App 'cuga-events' not found (already gone).`, removes the URL file, and exits 0. I reproduced this with a mocked CLI: no app deletion was issued. Both actual apps remain running and billable, including the event-processing loops. Share the two app-name defaults with the deployment script and delete both configured apps.

events/scripts/events_pg.sh
```diff
@@ -0,0 +1,76 @@
...
+    echo "✓ events postgres already running (${CONTAINER}, :${PORT})"
+  else
+    if _exists; then
+      echo "→ starting existing container ${CONTAINER} ..."
+      "$RT" start "$CONTAINER" >/dev/null
+    else
+      echo "→ creating ${CONTAINER} (postgres 16, :${PORT}) ..."
+      "$RT" run -d --name "$CONTAINER" \
+        -e POSTGRES_USER="$PGUSER" -e POSTGRES_PASSWORD="$PGPASS" -e POSTGRES_DB="$PGDB" \
+        -p "${PORT}:5432" "$IMAGE" >/dev/null
```
sami-marreed, line 56: [P2] Bind the development PostgreSQL port explicitly to loopback

`-p "${PORT}:5432"` publishes the database on all host interfaces rather than only localhost, contrary to the security assumption documented at the top of this script. The default database uses the published development credentials, so on a reachable development host another machine can connect and read or modify event state. Use `-p "127.0.0.1:${PORT}:5432"` for the default local-development path.

anupamamurthi:
Hi @sami-marreed To the best of my knowledge, we (Claude & I) fixed the issues you raised. I can explain the details if needed. 
copy pasting a blurb

```

**P1 — Unauthenticated callers can create admins** (`service.py:165`)
- All six `/api/events/admin/*` routes now require `X-Gateway-Token` before any role check, and refuse the request without it.
- The Studio can't hold that token, so admin calls from the UI go through a core proxy protected by `require_manage_access`. The proxy replaces `X-User-Id` with the logged-in user's verified identity, so the admin principal no longer comes from a header the caller controls.
- The deployment now has OIDC login turned on (IBM App ID). Verified live: an anonymous create-admin returns 401 both on the events service and through the core proxy (it returned 200 before).
- Tests call the auth dependency directly with auth on and auth off. A source-grep test had passed while the endpoint was still open.

**P1 — Push watches get the same flow name when reuse is disabled** (`concierge.py:680`)
- The flow name now comes from a separate `flow_identity`, captured before disabling reuse clears `dedup_key`. Different watches get different names. New test: `test_flow_name_uniqueness.py`.

**P1 — `.env.ce.core` is included in the build source** (`1_build_push_image.sh:40`)
- The exclude pattern is now `events/deploy/.env.ce*` plus the root `.env`. A check after staging stops the build if any secret file is still there.

**P2 — Fresh Postgres provisioning can lose the new credentials** (`4_postgres.sh:145`)
- The DSN and CA are written to `.env.ce` before the CE secret is touched, and the secret is created if it doesn't exist or updated if it does.
- `make_env_ce.sh` now keeps provisioned keys from the previous file instead of wiping them.
- `2_deploy.sh` refuses to deploy if they're missing.

**P2 — Teardown targets the retired app name** (`teardown.sh:15`)
- The app names are now defined in `config.sh`. Teardown deletes both split apps plus the legacy one, and says so if it found nothing to delete.

**P2 — Dev Postgres port is reachable beyond localhost** (`events_pg.sh:56`)
- It now binds to `127.0.0.1` only.

**Also fixed — the 9 other CodeRabbit threads**
- The shipped `GATEWAY_TOKEN` placeholder passed the token check. The check now rejects all known placeholders.
- `3_state_store.sh` never called the admin, login, or project-target checks. It does now.
- `ap_pieces.py` sent credentials over plain HTTP to remote hosts, and mistook an empty catalog for an unreachable Activepieces. Both fixed.
- The `test-new-pieces` Makefile target pointed at CUGA instead of the events service.
- Slack reaction events: fetching the message no longer delays the 3-second acknowledgement, and Slack retries are deduplicated on `event_ts`.
- The Box poller opened a new Postgres connection on every tick. It now reuses one per DSN.
- `mcp_servers` entries given as objects are kept instead of being turned into strings. Entries with no name, or with a URL that isn't http/https, are rejected.
- A reaction on a thread reply fetched the thread's parent message. It now fetches the right message, and Slack API errors are logged instead of silently ignored.

**Validation**
- Locally: 1690 unit tests and 539 events tests pass, lint is clean.
- A fresh image is deployed to Code Engine. Live end-to-end: 30 passed, 0 failed. Scheduled and polling flows fired 2/2. All four channels are connected.
- Anonymous requests get 401 on the admin routes, `/stream`, and `/run`.

**Known and tracked separately**
- Roles aren't enforced yet. Login is required, but every logged-in user has full access.
- After the gateway token is presented, the events service still accepts whatever identity the caller claims.
- `/openapi.json` is public.
- Your earlier asks, #5 (flatten the `events/` package) and #3 (move the docs tree), are follow-ups.
```

hopefully, after this, the fixes / changes around Eventing + Channels work will be manageable and PRs will be smaller. 

Next Steps is to rebase 605 (Whatsapp channel integration) in + docs PR.

src/cuga/backend/events/app.py
```diff
@@ -2910,6 +3127,9 @@ async def admin_list_users(request: Request):
     async def admin_add_user(request: Request):
         """Register an identity in this tenant so channel messages can resolve to a real user rather
         than the fallback. Admin-only; 501 when no user store is configured."""
+        _denied = _admin_denied(request)
+        if _denied is not None:
```
sami-marreed, line 3131: **[P1] Derive the acting admin identity from trusted authentication**

The proxy now correctly reuses `require_manage_access`, but `_principal_from` prioritizes the caller-supplied body/query `scope` over the verified `X-User-Id`. An authenticated ordinary user can supply `scope=default/default/admin` and pass the events admin check when Manage role enforcement is disabled. I reproduced the admin operation returning 403 without that scope and 200 with it. Please derive the acting identity from trusted authentication and authorize any requested scope separately.

sami-marreed (commented):
Two P1 issues remain: the events admin API trusts a caller-supplied identity through `scope`, and arming the same watcher twice can delete the first watcher’s Activepieces flow. Local PostgreSQL startup also remains broken. Please address these before merging. The cleanup comments below cover duplicate configuration, build logic, and ineffective regression tests.

src/cuga/backend/events/concierge.py
```diff
@@ -838,7 +868,9 @@ async def find_or_create_flow(
                 # the dedup-reuse check above BEFORE flow creation, so this hash never blocks a reuse.
                 import hashlib
 
-                _disc = hashlib.sha1(dedup_key.encode()).hexdigest()[:8]
+                # Hash the IDENTITY, never dedup_key: the latter is deliberately empty whenever
+                # EVENTS_FLOW_REUSE is off, which is the default.
+                _disc = hashlib.sha1(flow_identity.encode()).hexdigest()[:8]
```
sami-marreed, line 873: **[P1] Keep newly created flows distinct when reuse is disabled**

Capturing `flow_identity` fixes collisions between different repositories, but identical watches still receive the same flow name when reuse is disabled. Arming the same watch twice creates two subscriptions, while `APEngine._new_flow()` deletes the first same-named flow. The first subscription then references a deleted flow. Please give each newly created flow a unique name when reuse is disabled, or explicitly reuse the existing subscription and flow.

events/scripts/events_pg.sh
```diff
@@ -0,0 +1,80 @@
...
+    echo "✓ events postgres already running (${CONTAINER}, :${PORT})"
+  else
+    if _exists; then
+      echo "→ starting existing container ${CONTAINER} ..."
+      "$RT" start "$CONTAINER" >/dev/null
+    else
+      echo "→ creating ${CONTAINER} (postgres 16, :${PORT}) ..."
+      "$RT" run -d --name "$CONTAINER" \
+        -e POSTGRES_USER="$PGUSER" -e POSTGRES_PASSWORD="$PGPASS" -e POSTGRES_DB="$PGDB" \
+        # Loopback ONLY. A bare -p publishes on 0.0.0.0, and this container runs the
```
sami-marreed: **[P2] Move comments outside the continued container command**

The comments inside the backslash-continued container command terminate the command before its port and image arguments. The shell then executes `-p` as a command, producing `-p: command not found`; PostgreSQL does not start. Please move the comments outside the continued command and verify the startup path.

scripts/frontend_build.sh
```diff
@@ -0,0 +1,47 @@
...
+( cd "$WS" && pnpm install )
+
+echo "== build frontend (webpack, PRODUCTION) =="
+# NODE_ENV=production is REQUIRED, not a nicety. webpack.config.js keys `mode`, minification,
+# console-stripping and source-maps off `process.env.NODE_ENV === "production"`, and `pnpm run
+# build` sets nothing — so a bare build is a DEVELOPMENT build: unminified, with source maps, ~3.5x
+# larger. Committing that balloons the served bundle (main.js 1.3MB→4.6MB, vendors 7MB→24MB) and
+# turns a routine rebuild into a ~450k-line git diff, because the dev bundle has real newlines
+# where the production one is a single minified line.
+( cd "$FE" && NODE_ENV=production pnpm run build )
```
sami-marreed: Can we reuse `src/frontend_workspaces/frontend/build.sh`? It already builds production assets and copies them into the Python package. The additional setup and checks could wrap that script, keeping the build logic in one place.

events/examples/rosters/default.yaml
```diff
@@ -0,0 +1,93 @@
...
+#
+# Eight agents, one per capability area, covering every MCP server the stack ships
+# (finance · web · geo · knowledge · code · text). NOTHING here needs Activepieces:
+# AP is a TRIGGER backend (Gmail/GitHub/Box push), not an agent capability, so no agent
+# in the full roster depended on it either — this is a size trim, not an AP trim.
+#
+# The six the live harnesses exercise are load-bearing: pricebot, weatherbot, geobot,
+# wiki_dive, incident_triage, pr_reviewer. The full 27-agent example lives at
+# events/examples/rosters/supervisor_agents_full.yaml.
+supervisor:
```
sami-marreed, line 11: This configuration duplicates `docs/examples/events/supervisor_agents.yaml`. Can we keep one canonical roster and update all references? Deployment uses the new location, while `test_split_service.py` still loads the old file.

events/deploy/.env.ce.example
```diff
@@ -0,0 +1,25 @@
+# events/deploy/.env.ce  — the CE secret source for the CUGA-events app.
+# PLACEHOLDERS ONLY in this .example. Generate the real (gitignored) .env.ce with:
+#     ./make_env_ce.sh            # copies the allow-listed keys out of ../../.env
+# or copy this file to .env.ce and fill it in by hand. NEVER commit .env.ce.
```
sami-marreed: The documented manual setup is incomplete: this template omits required `AGENT_SETTING_CONFIG`, and deployment also requires `.env.ce.core`. Can we direct users to `make_env_ce.sh` and remove this template, or update the template and instructions to cover both files?

tests/events/test_flow_name_uniqueness.py
```diff
@@ -0,0 +1,84 @@
...
+
+from __future__ import annotations
+
+import hashlib
+
+EMPTY_SHA1_8 = hashlib.sha1(b"").hexdigest()[:8]
+
+
+def _identity(agent, source, cadence, cfg_tag, sink, task, owner):
+    """The identity string concierge builds — mirrored so the test pins the SHAPE, not the code."""
```
sami-marreed: These tests copy the naming algorithm and inspect source strings instead of exercising production behavior. They pass despite the repeated-watch failure described in this review. Can we replace them with behavioral cases in `test_flow_reuse_disabled.py` that verify every created subscription retains a valid flow?

events/examples/rosters/README.md
```diff
@@ -0,0 +1,87 @@
...
+and deployed, is the small no-AP core: 8 sub-agents under the `cuga` supervisor.)
+These roster files reorganize the **same schema** into **coherent
+domain-intelligence families**: one supervisor per domain, each owning a *small set of focused
+sub-agents that belong together*, so an agent means something ("Box Intelligence") instead of
+being a random pile.
+
+Grouped along the same lines as the Activepieces app catalogue — Enterprise Productivity,
+Market & Research Intelligence, and the document-/recording-/code-centric flows.
+
+> **Nothing here is wired in.** These are drop-in *alternatives* to the root roster, authored for
```
sami-marreed, line 14: **Optional scope cleanup:** Do we need all 13 alternative rosters in this PR? Could we keep the default plus one representative example and move the remaining optional examples to a separate PR?

anupamamurthi:
@sami-marreed here is a summary

1 · app.py:3131 — [P1] admin identity from scope

Fixed. Admin routes now authorize against a new _admin_actor() that reads the identity from trusted transport (X-User-Id) only, never the caller's scope. Verified live: scope=default/default/admin with a non-admin X-User-Id returns 403. Added a test with a real non-admin user that reproduces the escalation and confirms it's closed.

2 · concierge.py:873 — [P1] repeated-watch flow collision

Fixed. When reuse is off, each created flow now gets a unique suffix, so arming the same watch twice no longer lets _new_flow() delete the first flow. Replaced with a behavioral test that drives the real arming path against an engine modeling _new_flow's delete-same-named semantics — fails without the fix, passes with it.

3 · events_pg.sh:56 — [P2] comments break the container command

Fixed. Moved the comment above the command; the -p/image args are no longer dropped and Postgres starts. Verified the run line executes as a single intact command.

4 · test_flow_name_uniqueness.py:22 — ineffective tests

Agreed — removed it. The naming check is now a behavioral case in test_flow_reuse_disabled.py that verifies every created subscription keeps a live flow (see comment 2).

5 · frontend_build.sh:34 — reuse build.sh

Done. The script now delegates the build + package copy to frontend/build.sh and only adds the toolchain setup and a sanity check, so the build logic lives in one place.

6 · default.yaml:11 — roster duplication

Fixed. Removed docs/examples/events/supervisor_agents.yaml; events/examples/rosters/default.yaml is now the single canonical roster, and test_split_service.py points at it. No references to the old path remain.

7 · .env.ce.example:4 — incomplete template

Removed it. The docs now direct users to make_env_ce.sh, which generates both .env.ce and .env.ce.core; the annotated key list lives in the repo-root .env.events.example.

8 · rosters/README.md:14 — trim the 13 rosters

Sami, I would prefer to keep the full set as showcase examples in this PR rather than split them out — but happy to move them to a follow-up if you feel strongly. I am hoping this will help experiment quickly.

sami-marreed (approved):
Re-reviewed the fixes at 3d01657a. The admin identity spoofing, repeated-watch flow collision, and PostgreSQL startup issues are addressed. The concrete cleanup requests are also addressed; keeping the optional roster examples is non-blocking.

Validation: 50 focused tests passed (1 skipped), all four independent regression reproductions passed, and the PostgreSQL startup command was verified with a fake container runtime. A live Code Engine deployment was not run.

Approved.

Issue #599, opened 2026-08-06
[Bug]: CugaLite re-issues an identical rejected API call across dozens of turns; a 100%-failed execution produces no course correction


Sergey-Zeltyn:
## What happened

When every tool call in an executed script is rejected by the API, nothing in the graph registers that the turn achieved nothing. The next `call_model` turn re-issues the **same call with the same arguments**, and the cycle repeats until the task ends or `cuga_lite_max_steps` (70) is reached.

This is **turn-level** repetition, distinct from #385 (which is about how many calls one script may make). Verified by grouping error responses by their enclosing observation:

| task | rejected calls | distinct failing URLs | **distinct turns** | error message |
|---|---|---|---|---|
| `7574325_1` (med-run3) | 218 | **1** | **40** | `Invalid password reset code.` |
| `b0934aa_1` (med-run3) | 2,136 | **1** | **20** | `Your payment card doesn't have enough balance to place the order.` |
| `29a7b7e_1` (easy-run4) | 567 | **1** | **18** | `The path source_file_path (…) is invalid. It contains double slashes.` |

One endpoint, one argument shape, re-attempted across 18–40 independent model turns. None of these tasks hit the step ceiling — the loop simply ran until the task ended.

Across 5 AppWorld bundles there are **20,469** rejected API responses. **8,722** fall in this issue's scope (the remaining 11,747 are the "already exists" class, filed separately as #596).

### The strongest case: the agent states the correct diagnosis and does not apply it

In `29a7b7e_1`, the model writes out its own bug — in prose, on four separate turns (generations 4, 6, 10 and 15):

> "`source_file_path` and `destination_file_path` contained duplicated path segments (e.g. `~/documents/work/meetings//home/catherine/documents/work/meetings/2023-may-10__video_recording.mp4`). This happened because …"

and then emits the same malformed argument again. Final tally for that task:

| file/move calls | succeeded | rejected |
|---|---|---|
| 567 | **0** | **567** (all 422) |

The information needed to recover was not merely available — the agent had already derived it and written it down. The gap is between stating the correction and applying it on the next turn.

`b0934aa_1` is the same shape without the self-diagnosis: the API replies *"Your payment card doesn't have enough balance"*, the agent had already listed the available payment cards earlier in the task, and it re-sends the identical `{payment_card_id: 247, address_id: 112}` 2,136 times.

### Why this is not visible in pass rates

[...]

PR #628, opened 2026-08-11
fix(registry): treat idempotent-conflict responses as satisfied, not errors (#596)

### LINKED ISSUE #596: [Bug]: "already exists" API responses are surfaced as exceptions, so the agent retries a goal it has already achieved
## What happened

When an AppWorld API reports that the requested state **already holds** — `409 Conflict` "already exists", or a `422` whose body says "already in the playlist" / "already marked as archived" — the registry returns it to the agent as `{"status": "exception", ...}`. The agent treats it as a failed call and re-issues it, often thousands of times.

These are not failures. The post-condition the agent is trying to establish is already true; the correct interpretation is "done, move on".

**Across 5 AppWorld evaluation bundles, 11,747 of 20,469 error responses (57%) are this class.**

| count | status | message |
|---|---|---|
| 9,120 | 422 | `The song is already in the playlist.` |
| 2,371 | **409** | `Note with this user_id and title already exists.` |
| 148 | 422 | `This email thread is already marked as archived.` |
| 66 | **409** | `Playlist with this title and user_id already exists.` |
| 30 | 422 | `This email thread is already marked as read.` |
| 12 | 422 | `This email thread is already marked as unread.` |

Worst affected tasks:

| task | such errors | tool calls | match_rate |
|---|---|---|---|
| `d194965_1` | 9,114 | 985 | 0.833 |
| `0d01c76_1` | 2,295 | 463 | **1.0** |

### The clearest evidence that these are not failures

`0d01c76_1` (*"Import markdown notes … Each markdown file should become a separate note"*) **passed every check — 6/6, `match_rate = 1.0`** — while emitting 2,295 `409 "Note with this user_id and title already exists"` responses. The notes were created correctly. Every one of those 2,295 "errors" was noise the agent then reacted to.

### Why it is expensive

The responses are spread across ~17–20 separate model turns per task, not one loop: the agent re-executes essentially the same script after each "failure". So the cost is multiplied by both the in-script loop size and the number of turns. `d194965_1` spent 985 tool calls and 9,114 rejected responses on a task whose goal state was already largely established.

This is also invisible to pass-rate monitoring — `0d01c76_1` looks perfectly healthy in every report.

## How to reproduce

Any task that re-adds an existing entity. Minimal shape:

```python
# song already present in playlist 654
await spotify_add_song_to_playlist(playlist_id=654, song_id=1)
# -> {"status": "exception", "status_code": 422,
#     "message": "... {'message': 'The song is already in the playlist.'}"}
```

The agent sees `status: exception` and retries with identical arguments.

Observed in bundles (`cuga-eval`, AppWorld easy + med, 2026-08-02 / 08-03):
`20260803_082641_compare_gpt-oss-120b-aws-med-run3` — tasks `d194965_1`, `0d01c76_1`.

## Environment

- CUGA: `origin/main` (runs dated 2026-08-02 / 08-03)
- Model: `aws/gpt-oss-120b` (Bedrock via litellm), temperature 0.1, `max_completion_tokens` 16000
- Benchmark: AppWorld easy (43 tasks) + med (40 tasks), CugaLite
- Component: tools_env registry / MCP adapter

The choke point where a 4xx becomes an agent-visible exception:

`src/cuga/backend/tools_env/registry/registry/api_registry.py` (~line 436):

```python
return {
    "status": "exception",
    "status_code": error_detail['status_code'],
    "message": final_message,
    "error_type": type(e).__name__,
    "function_name": function_name,
    "error_detail": error_detail,
}
```

Every status ≥ 400 takes this path regardless of whether it denotes a real failure or an already-satisfied post-condition.

### Suggested direction

Classify idempotent-conflict responses as a **non-error** result at this choke point, e.g.:

```python
{"status": "ok", "already_satisfied": True, "status_code": 409, "message": ...}
```

so the agent reads it as "the goal is met" rather than "the call failed".

Scoping matters — a naive substring match on `"already"` would mask genuine errors. Suggested order:

1. **`409 Conflict`** first: semantically an idempotency/conflict signal, and it covers 2,437 of the observations with no message parsing at all.
2. For `422`, an explicit allow-list of known post-condition-already-true messages (the four `422` shapes above) rather than a broad regex.
3. Whatever the classification, keep the original status and message in the payload so genuine conflicts remain debuggable.

### Relationship to existing issues

- **#385** (in-script runaway) — bounds *how many* calls one script may make. A cap would limit the blast radius but would not stop the agent from re-attempting an already-satisfied goal. Complementary.
- **#584** (unknown args dropped, schema errors unrecoverable) — different layer: there the error never reaches the agent. Here it reaches the agent intact and is simply the wrong classification.

There is a separate, broader problem — the agent re-executing a script after *all* its calls were rejected, including genuinely actionable errors — which I'll file separately; this issue is deliberately scoped to the "already satisfied" case because it is cheap to fix and independently testable.

Sergey-Zeltyn:
## Bug fix

Fixes #596

### Summary

Every response with status ≥ 400 took the exception path in `ApiRegistry.call_function`, including those reporting that the requested post-condition **already holds** — the note exists, the song is in the playlist, the thread is already archived. The agent read a satisfied goal as a failed call and re-issued it.

Across five AppWorld bundles these accounted for **11,747 of 20,469 error responses (57%)**. `0d01c76_1` scored `match_rate = 1.0` while emitting 2,295 of them, so the cost is wasted turns and tokens rather than wrong answers.

**Scoping** (as set out in the issue):

- **409 Conflict** is treated as idempotent on its own — that is what the status means, and it needs no message parsing.
- **422** is reused by these APIs for genuine validation failures, so only an explicit allow-list of known "already true" messages qualifies. A broad match on `"already"` would mask real errors such as *"Your payment card doesn't have enough balance"*.
- The original `status_code` and `message` are preserved on the returned dict, so genuine conflicts remain debuggable.

Before / after, driving the real `call_function`:

| case | before | after |
|---|---|---|
| `409 "Note … already exists."` | `status=exception` | `status=success`, `already_satisfied=True` |
| `422 "The song is already in the playlist."` | `status=exception` | `status=success`, `already_satisfied=True` |
| `422 "…card doesn't have enough balance…"` | `status=exception` | **`status=exception`** (unchanged) |

Success paths in this function return the raw response body and carry no `status` key; consumers only special-case `"exception"`. So the new shape cannot be mistaken for a failure, and no existing consumer needed changing.

### Testing

`uv run pytest src/cuga/backend/tools_env/registry/tests/test_idempotent_conflict_not_error.py` → **18 passed**

- **Classifier:** 7 already-satisfied shapes accepted; 8 genuine failures rejected (insufficient balance, expired card, invalid reset code, malformed path, missing files, 400, 500, null message).
- **End-to-end through `call_function`:** 409 already-exists, 422 already-in-playlist, and a genuine 422 that must keep failing.

**Behavioural verification against the unfixed code** — driving `call_function` on `main` and on this branch produced the table above: the two idempotent cases flip, the genuine failure does not. (The test file itself cannot run against `main`, since it imports the new `is_already_satisfied`.)

No regressions: `uv run pytest src/cuga/backend/tools_env/registry/tests/` → **106 passed**. `ruff format` + `ruff check` clean.

CI already runs `./src/cuga/backend/tools_env/registry/tests/` wholesale (`tests.yml:70`), so the new file is picked up without an allowlist change.

### Not verified

No live eval run yet. The expected effect is a reduction in wasted turns and tokens on tasks that re-add existing entities (`d194965_1`, `0d01c76_1` are the clearest candidates); it is **not** expected to move pass rates, since the affected tasks were already reaching the correct end state.

## Summary by CodeRabbit

- **Bug Fixes**
  - Operations that are already complete now return successful results instead of errors for recognized conflict responses.
  - Specific “already satisfied” validation responses are handled consistently, while genuine failures continue to raise errors.
  - Original status codes and messages remain available in operation results.

- **Tests**
  - Added coverage for recognized conflicts, already-satisfied responses, and unrelated error conditions.


src/cuga/backend/tools_env/registry/registry/api_registry.py
```diff
@@ -23,6 +23,43 @@
...
+# "already" would mask real errors.
+IDEMPOTENT_CONFLICT_STATUS = 409
+
+ALREADY_SATISFIED_MESSAGES = (
+    "already in the playlist",
+    "already exists",
+    "already marked as archived",
+    "already marked as read",
+    "already marked as unread",
+)
```
sami-marreed: seems like solution is too specific to appworld

sami-marreed: maybe we shouldn't raise exception in case of 422 in general and print it that we got that response and stop the whole process...

Sergey-Zeltyn: @sami-marreed don't review meanwhile, next commit addresses problem that arose in testing, not your last comments

Sergey-Zeltyn: Thanks — both fair. Taking them in turn.

## "Too specific to appworld"

Agreed, and worth splitting the two halves because only one is coupled:

- **409 is generic.** RFC 9110 defines it as a conflict with the target resource's current state, so keying on it is HTTP semantics rather than AppWorld's. This half carries no coupling.
- **The 422 allow-list is the AppWorld-specific part** — `already in the playlist`, `already marked as read`, etc. are literally that API's wording. Another service would say "duplicate entry" or "resource already present" and get nothing from this.

The awkward bit is that the 422 half is where the volume is: of ~300 reclassifiable responses in the two most recent bundles, **270** were `422 "This email thread is already marked as read."` on `277d81d_1`. Dropping it keeps the code clean and forfeits most of the benefit.

Three options:

1. **409 only** — fully principled, no coupling, but loses ~90% of the observed cases.
2. **Configurable patterns** — e.g. `settings.advanced_features.already_satisfied_patterns`, defaulting to the current five, with an empty list disabling 422 handling entirely. Keeps the benefit and moves the API-specific knowledge into config, where a different deployment can supply its own wording.
3. Leave as-is.

I'd suggest **(2)** — happy to push that if you agree.

## "Shouldn't raise exception in case of 422 … stop the whole process"

I checked this and I don't think 422 currently stops anything: `call_function` **returns** a dict rather than raising, and the only consumer of `status == "exception"` is `api_registry_server.py:382`, which logs the message and enriches it from `error_detail`. The agent receives it as ordinary tool output and keeps going.

So if the concern is that a 422 aborts the run — it doesn't today.

If the concern is that the agent doesn't *act* on it, that's real and measured, but it's a different fix. #599 covers it: one task re-issued an identical rejected call across **40 separate turns**, and another made **567** file-move calls with **zero** successes while writing out the correct diagnosis of its own bug four times. The missing piece there is that nothing carries forward which call signatures have already been refused — a guard in the graph, not a change to how the registry reports status. Would value your view on that issue; it's the larger of the two.

## One correction to this PR since you reviewed

The first version only patched `registry.api_registry` (httpx), but AppWorld traffic is served by `mcp_manager.adapter` (requests), which builds its own error dict. So the classification **never ran in production** — bundle `20260811_165801` has 270 reclassifiable responses and zero reclassifications.

Fixed in `b82625b8`: the rule now lives in `utils/conflict_utils.py` and is applied at both error-construction sites, with tests that drive the adapter path using the exact response shape from those traces. Those tests fail against the unpatched adapter; the previous ones did not, since they constructed httpx errors and only exercised the other path.

sami-marreed:
1. Validate that agent seens error content both status and error message so it can get feedback to continue
2. Validate that when such exception does it see previous progress in the code block like variables generated?
3. Validate the flow, and define behaviour

Sergey-Zeltyn:
Moved to draft while I validate these. Answers to **1** and **2** below; they describe **`main` only** — no PR code involved — measured on bundles `20260811_115347` (easy, 43 tasks) and `20260811_100641` (med, 40 tasks). Point 3 follows in a separate comment.

## 1. Does the agent see the error content — status and message — so it can act on it?

**Yes, in full, as a Python value.**

On `main` an API 4xx never raises. `mcp_manager/adapter.py`'s handler catches the `requests` error and *returns* an error dict; FastMCP serialises it as an ordinary tool result, so `/functions/call` answers **200** and `cuga_lite/providers/registry.py::call_api` hands the dict to the generated code. Verbatim from the `20260811_100641` trace, task `ac62177_1`:

```json
{"status": "exception", "error_type": "HTTPError",
 "message": "409 Client Error: Conflict for url: .../email_threads/47823/unread {\"message\": \"The email thread with id 47823 does not exist.\"}",
 "status_code": 409, "url": "...", "method": "POST"}
```

Status, status code and the original API message are all there. It reaches the model twice — as the variable value in the next turn's variables summary, and through stdout if the code prints it.

The outer error layer (`ApiRegistry.call_function`'s `except` blocks, whose dict would become the HTTP status of `/functions/call` at `api_registry_server.py:415` and make the provider raise) **never fired** in either bundle: 0 occurrences of `HTTP Error:`, 0 `ToolError`, 0 `function_name` in any agent-visible payload.

One caveat that matters for how much the status field is worth: **the agent never branches on it in its own code.** Across the two bundles — `except Exception` 2,795, `isinstance(x, dict)` 1,896, `status == "exception"` **0**. The feedback is complete, but it is consumed only by the model reading the variables summary on the next turn, never by the code it writes.

## 2. When such an exception happens, does it still see the progress made in the block — the variables generated?

Two cases, and they differ:

**API errors (the #596 class) — the question doesn't arise.** Nothing is raised, the block runs to the end, every variable is kept. In the same `ac62177_1` trace the two 409 dicts (`not_spam_res`, `unread_res`) appear in the variables summary alongside `unspam_coworker_success`, `unread_coworker_success` and `final_check` produced by the same block.

**A genuine Python exception — no variables survive at all.** `CodeWrapper.build_async_main` ends with `return locals()`; if the body raises, `_async_main` never returns, `LocalExecutor.execute` never reaches `context_locals.update(result_locals)`, and `filter_new_variables` yields `{}`. Only stdout printed before the raise is preserved, as `Output before error: …`.

Verified against the real `LocalExecutor` with a 3-item loop failing on item 2:

| tool behaviour | agent sees | variables into next turn |
|---|---|---|
| returns the error dict (what `main` does for API errors) | stdout, block completes | `threads`, `marked`, `res`, `summary` |
| raises | `Output before error: …` + traceback | **none** |

On `main` that loss is caused by the agent's own code, not by APIs. Block aborts, easy/med:

| cause | easy | med |
|---|---|---|
| `NameError` | 452 | 451 |
| `AttributeError` | 24 | 435 |
| timeout | 57 | 150 |
| `KeyError` | 105 | 65 |
| **total aborts** | **638** | **1,136** |

The `NameError`s are almost all **fabricated tool names** — a function that was never put in the namespace, repeated dozens of times unchanged: 149 × `list_files`, 141 × `phone_get_contact_by_relationship_get`, 51 × `phone_get_contact_by_name` (easy); 125 × `spotify_search_tracks_search_get`, 110 × `read_file`, 45 × `spotify_get_user_playlists_playlists_get` (med). Two flavours: plausible-but-wrong OpenAPI-style names, and short convenience names the agent assumes exist.

So on `main`, **zero** block aborts are caused by API errors, and 1,774 are caused by ordinary failures in generated code — each taking the block's entire working set with it. Only the timeout branch tells the agent what it lost; the rest give a bare traceback.

## What this implies

For API failures the registry already gives the agent complete, non-destructive feedback on `main`. What is missing is not the content of the error but anything that makes the agent *act* on it (#599), plus the fact that a block dying from an ordinary Python error silently loses everything it computed — which is the larger effect by count and is independent of #596.

<sub>Counts are occurrences in the trace text; conversation history repeats the same error string across turns, so treat them as relative magnitudes rather than distinct events. Queries and the `LocalExecutor` probe available on request.</sub>

Sergey-Zeltyn:
Closing it currently, I did not find a good Appworld-independent solution. I"ll evaluate the issue again at the next stable version

PR #630, opened 2026-08-11
fix(cuga-lite): end the block on exit()/SystemExit instead of the process

### LINKED ISSUE #629: [Bug]: exit()/SystemExit in generated code kills the evaluation process instead of ending the block
### Bug Description

Generated code that reaches `exit()`, `quit()` or `raise SystemExit` terminates the **host evaluation process**, not just the code block.

Relaxed execution (benchmark / skills mode) hands generated code the full builtins dict via `RestrictedEnvironment.create_safe_builtins`, so `exit`, `quit` and `SystemExit` are all in scope. `SystemExit` derives from `BaseException`, so it walks past `except Exception` in `LocalExecutor.execute` and every handler above it, up through the eval loop.

An AppWorld `test_med` sweep died this way at task 23 of 40 — the remaining 18 tasks never ran, and the run reported itself "EVALUATION COMPLETE" with 22 tasks because the shutdown path summarised what had already been scored.

Worth noting the agent's *intent* is reasonable here — "stop this block, there is nothing more to do". The block is wrapped in `async def _async_main()`, so even a bare `return` would skip `return locals()` and silently discard everything the block computed. There is currently no way for generated code to end a block early without losing state.

### Steps to Reproduce

1. Run any evaluation in benchmark mode (e.g. AppWorld `test_med`).
2. Have the agent emit a block containing an early-exit guard — this shape is common when a lookup returns nothing:

```python
partner_email = None
# ...
if not partner_email:
    print("Partner email not found; cannot proceed.")
    exit()
```

3. The moment `exit()` is reached, the evaluation process terminates.

Minimal reproduction without the agent:

```python
from cuga.backend.cuga_graph.nodes.cuga_lite.executors import CodeExecutor
# state with relaxed execution enabled
await CodeExecutor.eval_with_tools_async(
    code="print('before'); exit()", _locals={}, state=state, mode='local'
)
```

### Expected Behavior

`exit()` ends the **block** early. Stdout and the variables computed before the exit are preserved and available to the next block, exactly as for a block that ended normally. The agent is told the block ended early and that its state survived.

### Actual Behavior

The Python process exits. All remaining tasks in the sweep are skipped. Nothing is written for the in-flight task.

### Environment

- macOS 15.3 (Darwin 25.3.0), Python 3.13 / 3.12
- cuga-agent on `fix/546-validate-shortlister-tool-names` (merge of `main`)
- cuga-eval AppWorld benchmark, SDK evaluator (`eval_appworld_sdk.py`), `test_med` group
- Model: `openai/gpt-oss-120b` via litellm proxy

### Error Logs / Screenshots

Run timeline — the process ends 94 ms after the code block is logged, with no execution output for the task:

```
11:16:25 | evaluate_all:616 - Tasks from eval_config.toml (group 'test_med'): 40 tasks
11:16:25 | [1/40] Task d9987f6_1
...
11:42:00 | [23/40] Task 8a13317_1
11:43:47.143 | sandbox_node:sandbox:131 - Generated code:   <block containing exit()>
11:43:47.237 | save_evaluation_results - Results saved to: appworld_sdk_20260811_111616_23405.json
```

Relevant code:

- `executors/common/restricted_environment.py:53` — relaxed execution returns the full builtins dict
- `executors/local/local_executor.py` — `except Exception` cannot catch `SystemExit`

Six further blocks in the same run contained `exit()` or `raise SystemExit` behind guards that were not taken, so this is latent on essentially every run rather than a one-off.

### Configuration

`benchmarks/appworld/config/appworld.env` (benchmark mode enabled), default `settings.toml`. No secrets involved.

### Additional Context

Stripping `exit`/`quit` from the benchmark builtins alone would not be sufficient — it does not cover `raise SystemExit`, and it would leave generated code with no way to stop a block early at all. The fix should honour the intent: catch `SystemExit` in the executor, recover the block's locals from the still-live frame on the traceback, and return them alongside stdout so nothing computed is lost.

sami-marreed:
## Bug fix

Fixes #629

### Summary

Generated code that reaches `exit()`, `quit()` or `raise SystemExit` terminated the **host evaluation process** instead of just its own code block.

Relaxed execution hands generated code the full builtins dict (`RestrictedEnvironment.create_safe_builtins`), so `exit`, `quit` and `SystemExit` are all in scope. `SystemExit` derives from `BaseException`, so it walked past `except Exception` in `LocalExecutor.execute` and every handler above it, up through the eval loop. An AppWorld `test_med` sweep died this way at task 23 of 40 — the remaining 18 tasks never ran, and the run still printed "EVALUATION COMPLETE" because the shutdown path summarised what had already scored.

**Root cause:** the intent is fine, the blast radius was not. "Stop this block, there is nothing more to do" is a reasonable thing for generated code to want, and the agent has no other way to say it — the block is wrapped in `async def _async_main()`, so even a bare `return` would skip `return locals()` and silently discard everything the block computed.

**Solution:** honour the intent rather than forbid it.

- Catch `SystemExit` in `LocalExecutor.execute`, before `except Exception`.
- `return locals()` never runs, but the frame is still alive on the traceback, so the block's variables are recovered from it (`_locals_from_frame`) and merged into `context_locals` — they persist to the next block exactly as for a block that ended normally.
- Return stdout plus a short factual note: the block ended early, the reason if one was given, and that its variables survived.

What the agent now sees:

```
Block ended early: the code called exit()/quit() or raised SystemExit (no candidate found).
This ended the block only — variables defined before it were kept and are available
in the next block. Nothing after the exit ran.
computed the total
```

#### Why not just remove `exit` from the benchmark builtins

It would not cover `raise SystemExit`, which appeared in the same run, and it would leave generated code with no way to end a block early at all — turning a recoverable stop into a `NameError` and a wasted retry.

### Testing
- [x] Verified fix locally; tests pass — `test_code_executor.py` 29 passed
- [x] Added regression test `test_generated_exit_ends_the_block_not_the_runtime`, covering: `exit()` reached (block ends, `cart_total` survives, post-exit variable does not), an explicit `raise SystemExit("reason")` surfacing its reason, and genuine exceptions still propagating
- [x] Verified locals recovery holds across an `await` boundary, not just straight-line code
- [x] `ruff check` / `ruff format --check` clean on the changed package

Note: 3 pre-existing failures in `test_e2b_lite.py` are unrelated — `RuntimeError: e2b-code-interpreter package not installed` (missing optional dependency).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of generated early-exit commands so execution stops safely without terminating the host application.
  * Preserved output and variables created before an early exit.
  * Displays the explicit reason for termination and excludes code that was not reached.

* **Tests**
  * Added coverage for early-exit behavior, output preservation, and variable handling.

### DIFF
--- a/src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/local/local_executor.py
+++ b/src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/local/local_executor.py
@@ -109,6 +109,29 @@ class LocalExecutor(BaseExecutor):
                         guidance += "(no stdout was printed before the timeout)"
                     return guidance
                 context_locals.update(result_locals)
+        except SystemExit as e:
+            # `exit()` is a reasonable thing for generated code to reach — "stop
+            # this block, there is nothing more to do". The intent is fine; the
+            # blast radius was not. Relaxed execution hands the block the full
+            # builtins, and SystemExit is a BaseException, so it walked past
+            # every `except Exception` above this and killed the host process
+            # mid-run (an evaluation sweep died at task 23 of 40 this way).
+            #
+            # So honour the intent instead of forbidding it: end the block early
+            # and keep everything a normal block would have kept. `return
+            # locals()` never ran, but the frame is still alive on the
+            # traceback, so the variables computed before the exit are readable
+            # from it and persist across blocks as usual.
+            context_locals.update(self._locals_from_frame(e, "_async_main"))
+            reason = str(e) if str(e) and not str(e).isdigit() else ""
+            note = "Block ended early: the code called exit()/quit() or raised SystemExit"
+            note += f" ({reason}).\n" if reason else ".\n"
+            note += (
+                "This ended the block only — variables defined before it were kept and "
+                "are available in the next block. Nothing after the exit ran.\n"
+            )
+            captured = stdout_buf.getvalue()
+            return f"{note}{captured}" if captured.strip() else f"{note}(no output printed)"
         except Exception as e:
             # Preserve prints from earlier successful lines so the agent still
             # sees discovery output (e.g. find_tools) when a later line raises.
@@ -122,6 +145,24 @@ class LocalExecutor(BaseExecutor):
             result = "<code ran, no output printed to stdout>"
 
         return result
+
+    @staticmethod
+    def _locals_from_frame(error: BaseException, func_name: str) -> dict[str, Any]:
+        """Read a still-live frame's locals off an exception's traceback.
+
+        Used when the block stopped somewhere other than its own ``return
+        locals()`` — the frame is kept alive by the traceback, so the variables
+        computed up to that point are still readable and need not be discarded.
+        """
+        frame = None
+        tb = error.__traceback__
+        while tb is not None:
+            if tb.tb_frame.f_code.co_name == func_name:
+                frame = tb.tb_frame
+            tb = tb.tb_next
+        if frame is None:
+            return {}
+        return {k: v for k, v in frame.f_locals.items() if not k.startswith("__")}
 
     def format_error(
         self,
--- a/src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/tests/test_code_executor.py
+++ b/src/cuga/backend/cuga_graph/nodes/cuga_lite/executors/tests/test_code_executor.py
@@ -561,3 +561,49 @@ async def test_mode_auto_detection(mock_state):
 
     assert 'x' in new_vars
     assert new_vars['x'] == 42
+
+
+@pytest.mark.unit
+@pytest.mark.asyncio
+async def test_generated_exit_ends_the_block_not_the_runtime(mock_state):
+    """exit() ends the block early and keeps its variables — it must not escape.
+
+    Relaxed execution hands generated code the full builtins dict, so `exit`,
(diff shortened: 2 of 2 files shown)


offerakrabi (approved):
## PR Review: #630 — fix(cuga-lite): end the block on exit()/SystemExit instead of the process

**Merge.** The fix is correct and the approach is sound — catching `SystemExit` at the right layer, recovering locals from the still-live frame, and returning stdout plus a clear agent-facing note. No blocking items.

sami-marreed:
Follow-up filed as #642 — the same `_locals_from_frame` recovery applies to the `asyncio.TimeoutError` path, where a block currently discards every variable it computed before stalling (`2e9b91e_1` lost three completed tool-call results this way and scored 0.200).

One constraint that surfaced while testing it, worth recording next to this PR: the timeout case needs more than reusing the helper. `LocalExecutor.execute` calls `asyncio.wait_for(async_main(), timeout=timeout)` on a bare coroutine, which cancels it and clears the frame before it can be read:

```
  wait_for(bare coro)        recovered=[]
  wait_for(shield(task))     recovered=['cart_total','denise_email']
  asyncio.wait + cancel      recovered=['cart_total','denise_email']
```

So the coroutine has to be wrapped in a Task whose cancellation is deferred, with the frame read before cancelling. Kept out of this PR to keep the `SystemExit` fix reviewable on its own.

PR #640, opened 2026-08-12
fix(agent): fall back to json_mode whenever an endpoint ignores json_schema

### LINKED ISSUE #639: [Bug]: shortlister structured output fails on gpt-oss-120b — json_schema fallback never fires
### Bug Description

Structured output fails outright for the shortlister on `gpt-oss-120b` behind a LiteLLM proxy, because the `json_schema` → `json_mode` fallback in `BaseAgent.create_validated_structured_output_chain` never fires.

`_structured_output_missing_parsed_field` recognises exactly one failure shape — a `ValueError` about a missing `'parsed'` / `refusal` field — and explicitly returns `False` for `ValidationError`:

```python
def _structured_output_missing_parsed_field(exc: BaseException) -> bool:
    if isinstance(exc, ValidationError):
        return False
    if not isinstance(exc, ValueError):
        return False
    ...
```

But that is not how these endpoints fail. The proxy accepts `response_format: json_schema`, silently ignores it, and returns prose or JSON with renamed keys. LangChain then raises `ValidationError` (or `OutputParserException`), the predicate says "not my case", the exception propagates, and `.with_retry(stop_after_attempt=3)` re-runs the **same failing** `json_schema` path three times before giving up — even though `json_mode` on the very same model succeeds in under 2 s.

Every `find_tools` call goes through this chain, so with a `gpt-oss-120b` deployment that ignores `json_schema`, tool discovery cannot return at all.

### Steps to Reproduce

```python
from langchain_openai import ChatOpenAI
from cuga.backend.cuga_graph.nodes.shared.base_agent import BaseAgent
from cuga.backend.cuga_graph.nodes.api.shortlister_agent.prompts.load_prompt import ShortListerOutputLite

llm = ChatOpenAI(model="aws/gpt-oss-120b", base_url=<litellm proxy>, api_key=..., temperature=0.1)
chain = BaseAgent.create_validated_structured_output_chain(llm, ShortListerOutputLite, prompt)
await chain.ainvoke({"input": "pick the 2 most relevant tools for sending mail"})
```

Fails 3/3. The same call with `method="json_mode"` succeeds 5/5.

### Expected Behavior

Any structured-output parse failure falls back to the `json_mode` + `PydanticOutputParser` path, which is what the fallback exists for. One wasted attempt, not three, and a working result on models that ignore `json_schema`.

### Actual Behavior

`ValidationError` propagates, `json_schema` is retried three times, the call fails. The fallback is never reached.

### Environment

- macOS 15.3 (Darwin 25.3.0), Python 3.13
- cuga-agent `main`
- LiteLLM proxy (IBM internal), `MODEL_NAME=aws/gpt-oss-120b`, `AGENT_SETTING_CONFIG=settings.openai.toml`
- Also reproduced against `azure/gpt-oss-120b` and watsonx `openai/gpt-oss-120b` for comparison

### Error Logs / Screenshots

Predicate check against the exception actually raised — the fallback would not fire:

```
model: aws/gpt-oss-120b
  json_schema run 1: ValidationError   fallback_would_fire=False   'Invalid JSON: expected value at l…'
  json_schema run 2: ValidationError   fallback_would_fire=False
  json_schema run 3: ValidationError   fallback_would_fire=False
  cuga chain  run 1: FAIL  ValidationError
  cuga chain  run 2: FAIL
  cuga chain  run 3: FAIL

model: azure/gpt-oss-120b
  cuga chain  run 1: PASS
  cuga chain  run 2: PASS
  cuga chain  run 3: PASS
```

Method matrix, same `ShortListerOutputLite` schema, 5 distinct prompts per cell (an earlier round using one repeated prompt was contaminated by response caching and is excluded):

| method | watsonx `openai/gpt-oss-120b` | litellm `aws/gpt-oss-120b` | litellm `azure/gpt-oss-120b` |
|---|---|---|---|
| `json_schema` | 5/5 · 7.3 s | **0/5 ValidationError** | 4/5 · 9.3 s |
| `json_mode` | 5/5 · 6.1 s | 5/5 · 1.9 s | 0/5 OutputParserException |
| parser only (no `response_format`) | 5/5 · 7.0 s | 5/5 · 1.2 s | 5/5 · 7.8 s |
| `function_calling` | 0/5 | 0/5 (returns `None`) | 0/5 IndexError |

Probing the proxy directly on `aws/gpt-oss-120b`, every server-side constraint is accepted and discarded:

```
guided_json  (vLLM constrained decoding)  200  -> prose
extra_body.guided_json                    200  -> prose
response_format: json_object              200  -> prose
response_format: json_schema (strict)     200  -> prose
```

### Configuration

`.env`: `MODEL_NAME="aws/gpt-oss-120b"`, `AGENT_SETTING_CONFIG=settings.openai.toml`, `OPENAI_BASE_URL` pointing at the LiteLLM proxy. Secrets redacted.

### Additional Context

Two further observations from the same benchmark, relevant to how the fallback should be ordered:

1. `function_calling` is 0/5 on **all three** backends for this model family — on `aws` it does not even raise, it returns `None` (`tool_calls: []`, prose content). It should never be selected for `gpt-oss-*`.
2. Constrained decoding is *slower* here, not faster: 6–9 s for enforced paths versus 1.2 s for the unconstrained parser path. Since `find_tools` blocks have separately been observed exceeding the 60 s execution budget, method choice has a real latency cost per shortlister call.

sami-marreed:
## Bug fix

Fixes #639

### Summary

Structured output failed outright for the shortlister on `gpt-oss-120b` behind a LiteLLM proxy. Two separate defects, both on the fallback path.

**1 — the fallback never fired.** `_structured_output_missing_parsed_field` recognised exactly one failure shape (a `ValueError` about a missing `'parsed'`/`refusal` field) and explicitly returned `False` for `ValidationError`. But endpoints that ignore `response_format` answer with prose or renamed keys, which surfaces as `ValidationError` / `OutputParserException`. The predicate said "not my case", the exception propagated, and `.with_retry(stop_after_attempt=3)` re-ran the **same failing** `json_schema` request three times.

**2 — the fallback could not have worked anyway.** `json_mode_chain` was `prompt_template | llm | parser`. That runs the model unconstrained, and a bare `PydanticOutputParser` only *validates* — it never asks for JSON. Nothing in the prompt told the model what shape to produce. Fixing (1) alone just moved the failure one step later, which is how this surfaced.

### Root Cause

The fallback encoded one endpoint's failure mode as if it were the only one, and never carried the schema into the request it fell back to.

### Solution

- `_json_schema_unusable()` treats any structured-output parse failure — `ValidationError`, `OutputParserException`, `JSONDecodeError`, plus the original missing-`parsed` shape — as "this endpoint will not honour the schema, retrying is pointless". Transport errors (auth, rate limit, connection, timeout) keep propagating so the caller's retry still does its job.
- The fallback prompt now appends `BaseAgent.get_format_instructions(parser)`, passed as a **partial** so the schema's own braces aren't parsed as template variables.
- The warning now logs the exception type and message instead of asserting `parsed=None, refusal=None`, which was only ever true for one of the shapes.

### Verification against the live endpoint

Same `ShortListerOutputLite` schema, real `create_validated_structured_output_chain`:

```
aws/gpt-oss-120b     before  0/3 FAIL (ValidationError x3)   after  3/3 PASS
azure/gpt-oss-120b   before  3/3 PASS                        after  3/3 PASS
```

The `azure` run is the no-regression check: where `json_schema` already worked, it still takes that path and the fallback is untouched.

Supporting matrix, 5 distinct prompts per cell (an earlier round with one repeated prompt was contaminated by response caching and is excluded):

| method | watsonx `openai/gpt-oss-120b` | litellm `aws/gpt-oss-120b` | litellm `azure/gpt-oss-120b` |
|---|---|---|---|
| `json_schema` | 5/5 · 7.3 s | **0/5** | 4/5 · 9.3 s |
| `json_mode` | 5/5 · 6.1 s | 5/5 · 1.9 s | 0/5 |
| parser + format instructions | 5/5 · 7.0 s | 5/5 · 1.2 s | 5/5 · 7.8 s |
| `function_calling` | 0/5 | 0/5 (returns `None`) | 0/5 |

### Testing
- [x] Verified fix locally; tests pass — 9 passed in `tests/unit/test_structured_output_fallback.py`
- [x] New tests cover both directions: four schema-ignored shapes trigger the fallback, four transport errors still propagate, and one end-to-end pass through the `except` branch asserting json_mode is reached exactly once
- [x] Verified live against two deployments (above), including a no-regression check
- [x] `ruff check` / `ruff format` clean

### Follow-ups (not in this PR)

- `function_calling` is 0/5 on **all three** backends for `gpt-oss-*`; on `aws` it returns `None` rather than raising. It should probably never be selected for that family.
- Constrained decoding is *slower* here than the parser path (6–9 s vs 1.2 s). Given `find_tools` blocks have been observed exceeding the 60 s execution budget, method ordering per model family is worth revisiting.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved structured-output handling when schemas are unsupported or responses are invalid, incomplete, or malformed.
  * Added a fallback to JSON mode with clearer formatting guidance when structured responses cannot be used.
  * Transport and operational errors continue to be reported rather than silently retried.

* **Tests**
  * Added coverage for schema incompatibility, parsing and validation failures, missing fields, malformed JSON, and transport errors.


sami-marreed:
## Extended validation — 320 live calls, before/after

Baseline is a pristine `origin/main` worktree, so both sides run the real `BaseAgent.get_chain` against the same endpoints. **5 models × 4 schema shapes × 5 prompts per run**, three independent prompt sets.

Schema shapes: `flat` (scalars), `nested` (list of objects — the shortlister's shape), `optional` (`Optional` + defaulted fields), `deep` (two levels of nesting, produces `$defs`).

### Per-schema, `aws/gpt-oss-120b`

| run | prompts | flat | nested | optional | deep | total |
|---|---|---|---|---|---|---|
| before (main) | set A | 2/5 | 3/5 | **0/5** | 5/5 | **10/20** |
| before (main) | set B | 3/5 | 1/5 | **0/5** | 5/5 | **9/20** |
| after (this PR) | set A | 5/5 | 5/5 | 5/5 | 5/5 | **20/20** |
| after (this PR) | set C | 5/5 | 5/5 | 5/5 | 5/5 | **20/20** |

Two things this shows that the original report did not:

- The failure was **partial and shape-dependent**, not total — `deep` passed 5/5 even on main, while `optional` (`Optional[str] = None`, `int = 0`) was **0/5 on both baseline runs**. Schemas with optional or defaulted fields never worked.
- It reproduces across two independent prompt sets, so it isn't prompt luck.

### No regression on other models

Every model that reaches `create_validated_structured_output_chain` — `ChatOpenAI` and `ChatWatsonx` alike:

| model | before | after |
|---|---|---|
| `aws/gpt-oss-120b` | 10/20, 9/20 | **20/20, 20/20** |
| `azure/gpt-oss-120b` | 20/20 | 20/20 |
| `azure/gpt-5.3-chat` | 20/20 | 20/20 |
| `Azure/gpt-5.2-chat-2025-12-11` | 20/20 | 20/20 |
| watsonx `openai/gpt-oss-120b` | 20/20 | 20/20 |

Where `json_schema` already worked, the request is unchanged and the fallback is never entered — the four unaffected models stay at 20/20 across every schema shape.

### Note on latency

An intermediate run showed 0.3 s medians on the Azure models; that was proxy response caching from the preceding run with identical prompts. The final `after` run used prompt set C, never sent to either side, so its medians are real: `aws` 6.5 s, `azure/gpt-oss` 3.0 s, `gpt-5.3` 4.4 s, `gpt-5.2` 7.3 s, watsonx 12.7 s.

offerakrabi (commented):
## PR Review: #640 — fix(agent): fall back to json_mode whenever an endpoint ignores json_schema

**Merge with fixes.** The fix is correct and the predicate tests are solid; one minor gap in the end-to-end test coverage.
Approve once #1 is addressed or consciously deferred.

|---|---|---|---|---|
| 1 | Minor | End-to-end test bypasses the real fallback wiring | `tests/unit/test_structured_output_fallback.py:80-92` | `test_chain_falls_back_on_validation_error` re-implements `_invoke_with_fallback` by hand → a change to the actual closure (e.g. wrong chain invoked, condition moved) would leave all tests green. |

## Findings

_No Critical or Major findings._

## Notes

- The shortlister prompt already bakes format instructions in at construction time (`shortlister_agent.py:128`). When the fallback fires, `json_mode_prompt` appends a second copy of the same instructions. Redundant but harmless — not a finding.

sami-marreed:
Thanks — finding #1 is right, and it was the one that mattered. The hand-rolled `invoke` duplicated `_invoke_with_fallback`, so it covered the predicate (already covered by the parametrized tests) and nothing about the wiring — including the `cuga_format_instructions` partial, which is the half of the fix that only surfaced against a live endpoint and had no unit coverage at all.

Replaced in 7fb986d with a `_FakeLLM` that composes with `|` and records call order, driving the real `create_validated_structured_output_chain`. It also asserts the fallback prompt actually renders the format instructions, and adds the negative case (transport error must propagate without reaching json_mode).

Mutation-checked so it isn't just green-by-construction — each of these now fails the suite:

| mutation | result |
|---|---|
| drop `.partial(cuga_format_instructions=...)` | 1 failed |
| fallback re-invokes `json_schema` instead of `json_mode` | 1 failed |
| revert predicate to `_structured_output_missing_parsed_field` | 1 failed |

10 passed on the branch after rebasing onto the main merge.

---

On the note about duplicate format instructions — confirmed, and I'd keep it. `shortlister_agent.py:128` does pass `format_instructions` into its own template, so the shortlister gets a second copy when the fallback fires: **646 chars, ~161 tokens**. Two reasons not to dedupe: the chain is generic and can't know whether a given caller already baked them in (most don't), and on `aws/gpt-oss-120b` the fallback fires on every shortlister call, so a conditional would need to be right rather than merely clever. Agreed it's harmless at that size — flagging the number in case it looks different to you.

sami-marreed:
1. run on test aws/gpt-oss-120b

sami-marreed:
Difficulty Breakdown
--------------------

Diff  Tasks    Pass@1      Tokens    Tok/Task  LLM Calls   LLM/Task   Duration   Dur/Task
─────────────────────────────────────────────────────────────────────────────────────────
1        43     83.7%           0         0.0          0        0.0         --         --


Test-Set Breakdown (AppWorld)
-----------------------------

Test Set   Tasks    Pass@1      Tokens    Tok/Task  LLM Calls   LLM/Task   Duration   Dur/Task
──────────────────────────────────────────────────────────────────────────────────────────────
challenge     24     83.3%           0         0.0          0        0.0         --         --
normal        19     84.2%           0         0.0          0        0.0         --         --

Sergey-Zeltyn (approved):

PR #645, opened 2026-08-12
fix(code-extraction): keep only the first fenced block from reasoning

### LINKED ISSUE #626: [Bug]: Reasoning draft code fences are combined into executable Assistant_code
Part of #616.

## What happened

When a reasoning model puts multiple ```python fences in `Assistant_reasoning` (drafts / retries), `extract_code_from_model_response` falls back to that reasoning text and `extract_and_combine_codeblocks` joins **every** fence into one string.

That combined blob is shown/executed as `Assistant_code`, so retry snippets from thinking become one redundant program (e.g. six nearly identical `find_tools(...)` calls) instead of a single final executable block.

Expected: only the intended executable code runs — typically the last fenced block from the final response, or a single chosen block from reasoning — not a concatenation of every draft fence in the reasoning stream.

## How to reproduce

1. Run Cuga Lite / code-agent with a model that emits fenced ```python inside reasoning (and little/no fenced code in the final content).
2. Give a task that triggers tool discovery (e.g. Spotify / `find_tools`) so reasoning contains several draft ```python attempts.
3. Inspect the trajectory: `Assistant_reasoning` has multiple fences; `Assistant_code` is all of them joined with blank lines.

Unit-level: pass reasoning text with N ```python blocks into `extract_code_from_model_response("", reasoning)` and observe `"\n\n".join(blocks)` behavior in `extract_and_combine_codeblocks`.

## Environment

- Observed on branch synced to `origin/main` (`209335fb`)
- Affected path: `src/cuga/backend/cuga_graph/nodes/cuga_agent_core/execution/code_extraction.py` (`extract_and_combine_codeblocks`, `extract_code_from_model_response`)
- Triggered when final `content` has no extractable code and extraction falls back to `reasoning_content`
- Relevant to AppWorld / autonomous tool-use evals where reasoning models draft multiple discovery scripts in-thinking

Trajectory shows:
1. `Raw_Assistant_Response` — final NL saying tools weren’t found
2. `Assistant_reasoning` — multiple draft ```python / `find_tools` attempts
3. `Assistant_code` — those drafts concatenated into one executable block

Root cause (current behavior is intentional by docstring/name, but wrong for reasoning drafts):

```python
# extract_and_combine_codeblocks
return "\n\n".join(blocks)

# extract_code_from_model_response
code = extract_and_combine_codeblocks(content, ...) if content else ""
if not code and reasoning_content:
    code = extract_and_combine_codeblocks(reasoning_content, ...)
```

## Possible approaches

1. **Prefer last fence** — when extracting from `reasoning_content` (or when multiple fences exist), take only the last ```python block instead of joining all drafts.
2. **Use NL auto-continue** — if the final response is NL / refusal / “still searching” with draft code only in reasoning, do **not** execute the concatenated reasoning fences; classify via `cuga_lite_nl_auto_continue` / `classify_nl_auto_continue` and loop back to `call_model` so the next turn emits a single executable block (similar spirit to probing truncation forcing a fresh model turn).
3. **First-block + continue** — execute only the first fence, then auto-continue for subsequent discovery/retries once tool output is visible (reuse `tools_needing_probing`-style isolation).
4. **Ignore fenced code in reasoning unless content is empty and a single fence is present** — treat multiple reasoning fences as non-executable drafts.

## Checklist

sami-marreed:
## Bug fix

Fixes #626

### Summary

When a reasoning model emits multiple ```python fences in `Assistant_reasoning` and
the final content has no code, extraction fell back to the reasoning text and joined
**every** fence into one string, which was then shown and executed as `Assistant_code`.

Fix: on the reasoning fallback only, keep the **first** fenced block. 0–1 fences pass
through unchanged, so single-fence reasoning models are unaffected; `content` still
combines all of its blocks into one program.

### Why the first fence (validated on cuga-eval trajectories)

Scanned `cuga-eval/benchmarks/appworld/logging/trajectory_data` — 775 trajectory files,
2401 `Assistant_code` turns across the AppWorld gpt-oss runs. 10 distinct turns hit this
bug (content fenceless, >1 reasoning fence, code == the join).

- **10/10: the first reasoning fence is the `find_tools(...)` discovery call.** Multi-fence
  reasoning is the model drafting a whole sequential plan in thinking, not retrying one call.
- **4/10: the last fence is downstream code** calling tools discovery has not bound yet
  (`venmo_accept_payment`, `gmail_list_threads_threads_get`, `spotify_create_playlist_post`).
  All 4 crashed with `Output before error: # Found N Matching Tool(s)` — find_tools ran,
  printed, then the concatenated remainder died. Keeping the *last* fence would crash the
  same way **and** throw the discovery output away.
- The other 6: the last fence is also a `find_tools` call with a near-identical query, so
  taking the first loses nothing.

Reasoning fences are written blind — with zero tool output — so only the first has its
preconditions met at execution time. Same principle as the probing isolation already in
this module: run the discovery, let a fresh turn write the real program with the result visible.

Replaying all 10 real turns through the patched extractor: 10/10 now execute exactly the
discovery call (e.g. `9ef798c_1`: 1125 chars of 12 concatenated fences → 104 chars).

### Why not NL auto-continue (approach 2 in the issue)

| Scenario | first fence | NL auto-continue |
|---|---|---|
| #626 itself: "couldn't find Spotify tools" + drafts | runs the discovery, next turn sees real tool output | classifier fast-path misses and the prompt rule *"if ANY clause says tools are not available, prefer false"* applies → turn ENDs, nothing executes |
| empty content + drafts | runs the discovery | fast-path misses → depends on an LLM call |
| planning content + drafts | runs the discovery | auto-continues, but the model gets no new information — same context, likely redrafts in reasoning until the step limit |
| single reasoning fence | runs it (today's behavior) | regression: reasoning-only models that put all code in reasoning stop executing |

Auto-continue costs an extra model turn that adds no tool output in its best case, and it
can't be scoped to "multiple fences only" without the same fence counting this fix already
does. The auto-continue path is unchanged for reasoning with no code at all.

### Testing

- `test_reasoning_drafts_do_not_concatenate` — discovery-then-dependent-calls shape from the real trajectories yields only the discovery fence
- `test_content_blocks_still_combine` — content blocks still join into one program
- Trajectory replay: 10/10 real #626 turns produce the discovery call alone
- `uv run pytest src/cuga/backend/cuga_graph/nodes/cuga_agent_core src/cuga/backend/cuga_graph/nodes/cuga_lite` → 578 passed; the 5 failures (e2b / workspace-fs) reproduce on a clean tree and are unrelated
- `uv run ruff check` / `ruff format` clean

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved code extraction when multiple fenced sections are present by consistently selecting the first Python code block.
  - Preserved the correct combination of content fences into a single executable program.

- **Tests**
  - Added regression coverage for multi-step reasoning and tool-discovery scenarios.


offerakrabi (commented):
## PR Review: #645 — fix(code-extraction): keep only the first fenced block from reasoning

**Merge with fixes.** The fix is correct, well-evidenced, and cleanly scoped; one small compliance item to land first.
Approve once #1 is addressed.

|---|---|---|---|---|
| 1 | Minor | New tests lack required type marker | `src/cuga/backend/cuga_graph/nodes/cuga_agent_core/tests/execution/test_code_extraction.py` | `AGENTS.md` and `pyproject.toml` require every new/changed test to carry a type marker; the two new tests (and the whole pre-existing file) have none → `pytest -m unit` silently skips this entire suite. |

PR #656, opened 2026-08-13
fix(supervisor): pass known variables when delegation omits them

### LINKED ISSUE #655: [Bug]: Supervisor drops structured variables on delegation so sub-agents ask the user for data it already has
Part of #258.

## What happened

`CugaSupervisor` can already hold a value such as `user_id=user_alice_99` after `user_finder` returns, but the next `delegate_to_account_manager(...)` call often omits `variables=["user_id"]`. Delegation then invokes the sub-agent with `variables=None`. The sub-agent asks for the user ID; the supervisor echoes that question back instead of retrying with the value it already has. The bonus step never runs.

Seen in CI: `test_supervisor_coordination_with_variable_passing` — 100 SDK tests passed, this live Watsonx run failed with:

> Could you please provide the account value for user id **user_alice_99**?

Expected: pass `user_alice_99` into `account_manager`, then `bonus_processor`, and mention a bonus in the answer.

This is a small, self-contained supervisor-delegation bug (no new Feature). Parent is Epic #258 because it is CugaSupervisor orchestration.

Related but different: #445 (CugaLite finalizes on a question in autonomous mode).

## How to reproduce

1. Run the SDK test (live LLM):

```bash
uv run pytest src/cuga/sdk_core/tests/test_supervisor_advanced.py::TestSupervisorAdvanced::test_supervisor_coordination_with_variable_passing -v
```

2. Or invoke a supervisor with three agents (`get_user_id` → `get_user_account_value` → `process_special_bonus`) and task: “Find the user ID for Alice, then get her account value, and finally process a special bonus for her.”
3. When the model generates `delegate_to_account_manager("Get account value...")` without `variables=`, the sub-agent does not receive `user_alice_99`.

## Environment

- CUGA `main` / PR CI (Python 3.12, Watsonx)
- `src/cuga/sdk_core/tests/test_supervisor_advanced.py`

Delegation logs showed `Delegating to account_manager: Get account value for user ID...` then the sub-agent asked for a user ID even though `user_alice_99` was already known. `delegation.py` only resolves structured variables when the generated code passes `variables=[...]`.

## Checklist

sami-marreed:
## Bug fix

Fixes #655

### Summary

When the supervisor already has values in its variable manager (e.g. `user_id=user_alice_99`) but generated code calls `delegate_to_account_manager(...)` without `variables=[...]`, the sub-agent was invoked with `variables=None` and asked the user for data the supervisor already held.

Delegation now forwards the supervisor VM when `variables` is omitted. `variables=[]` still opts out. The supervisor prompt also tells the model not to echo a sub-agent's clarifying question for data already in Variables.

### Testing
- `uv run pytest src/cuga/backend/cuga_graph/nodes/cuga_supervisor/tests/test_delegation_recording.py` — 9 passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved conversational agent delegation by automatically passing available context variables when no specific variables are provided.
  * Preserved explicit empty-variable requests to prevent unintended context sharing.
  * Added retry behavior when delegated agents request data already available in context.
  * Improved support for forwarding variables created during an ongoing workflow.
  * Ensured external agent delegation only shares variables when variable passing is enabled.

* **Documentation**
  * Updated delegation guidance to recommend explicit variable selection for structured parameters.


haroldship:
@coderabbitai when the script calls delegate_to_account_manager(..., variables=['user_id'])
and user_id is in the supervisor VM, what exactly does the sub-agent receive?
Please trace which frame resolve_names_from_caller_frame's f_back lands on.

haroldship:
@coderabbitai Same question for the two A2A branches, since they only have that path.

src/cuga/backend/cuga_graph/nodes/cuga_supervisor/prompts/supervisor_lite_prompt.jinja2
```diff
@@ -79,7 +79,8 @@ Your output MUST be one of these two types. **Do not mix them.**
...
 4.  **CHECK VARIABLES FIRST:** Before delegating to an agent, check if variables from a previous code execution already contain the data you need.
 5.  **COORDINATE MULTIPLE AGENTS:** You can call multiple agents in sequence across separate execution cycles, using results from one agent to inform the next agent's task.
-6.  **DYNAMIC PARAMETER PASSING:** After each agent execution, you will see the results in the execution output and available variables. You can analyze these results and decide what parameters to pass to the next agent. This allows for dynamic, adaptive workflows where Agent B's task depends on Agent A's results.
+6.  **DYNAMIC PARAMETER PASSING:** After each agent execution, you will see the results in the execution output and available variables. You can analyze these results and decide what parameters to pass to the next agent. This allows for dynamic, adaptive workflows where Agent B's task depends on Agent A's results. Prefer `variables=["name"]` so the next agent receives structured values. If you omit `variables`, known Variables are still forwarded.
```
haroldship: @coderabbitai given what you answered in my other questions, does the new instruction "Prefer variables=["name"] so the next agent receives structured values." direct the supervisor to the correct code path?

src/cuga/backend/cuga_graph/nodes/cuga_supervisor/delegation.py
```diff
@@ -78,9 +87,10 @@ async def delegate_to_agent(task: str, variables: Optional[List[str]] = None) ->
         logger.info(f"Delegating to {agent_name}: {task[:100]}...")
 
         if isinstance(agent_or_config, CugaAgent):
-            vars_to_pass = {}
             if variables is not None:
                 vars_to_pass = resolve_names_from_caller_frame(variables)
+            else:
+                vars_to_pass = _variables_from_supervisor_vm()
```
haroldship, line 111: @coderabbitai if the supervisor's variable context is very very large, how will the downstream agents behave?

haroldship (changes requested):
The omitted-`variables` path works and does fix #655. Blocking on the explicit path.

**`variables=[...]` forwards nothing.** `resolve_names_from_caller_frame` does a
single `frame.f_back`, and it's called from inside `delegate_to_agent` — so it lands
on that frame, whose globals are `delegation.py`'s module dict. The script's variables
are one frame further up, in `_async_main.__globals__`. It returns `{}` silently, and
the sub-agent is invoked with `variables=None`.

This drops into your own test file, uses only your helpers, and fails on `525b9b3`:

<details>
<summary>Failing test</summary>

```python
@pytest.mark.unit
@pytest.mark.asyncio
async def test_delegation_forwards_explicit_variables():
    mock_agent = _fake_agent()
    with patch("cuga.sdk.CugaAgent", _FakeCugaAgent):
        await _run_delegate(
            _empty_delegation_state(),
            mock_agent,
            "async def _run():\n    return await delegate('get account value', variables=['user_id'])\n",
            variable_manager=_FakeVM({"user_id": "user_alice_99"}),
        )

    assert mock_agent.invoke.await_args.kwargs["variables"] == {"user_id": "user_alice_99"}
```

```
E       AssertionError: assert None == {'user_id': 'user_alice_99'}
1 failed, 9 passed
```

</details>

**Why this blocks rather than being pre-existing.** The defect predates the PR, but
rule 6 is new: "Prefer `variables=["name"]`". A model that follows it always takes the
explicit branch and never reaches the fallback you added — so it gets the #655
behaviour back. Rule 7's retry advice lands on the same path.

**Ask:** fix the explicit path so rules 6 and 7 hold as written. Resolving against the
VM alone won't do it — new variables only enter the VM after execution completes
(`execute_agent_tool.py:92-95`), so mid-script values aren't there yet, which is the
case caller-frame resolution exists for. Bounding the frame walk to the frame holding
`SUPERVISOR_EXEC_KEY` (that's `_async_main`), then filling gaps from the VM, covers
both. The test above is the regression test.

If you'd rather keep this PR narrow: drop "Prefer `variables=[...]`" from rule 6,
soften rule 7, and handle the resolver separately.

Both A2A branches use the same helper, so external agents receive nothing either and
`pass_variables_a2a` has no observable effect today. The same fix closes that.

Non-blocking: `_variables_from_supervisor_vm()` copies the entire VM, and sub-agent
results bridge back into it (`delegation.py:100-117`), so every omitted-`variables`
delegation forwards everything accumulated in prior turns. Reasonable default for #655
— was the full copy deliberate over a most-recent-N limit?

One more thing: `test_supervisor_graph_adapter.py:172` calls the helper directly from the test body, so f_back lands on the test function. It can't fail at the depth production uses.

src/cuga/backend/cuga_graph/nodes/cuga_supervisor/prompts/supervisor_lite_prompt.jinja2
haroldship: earlier you wrote:

resolve_names_from_caller_frame() uses only one frame hop:

caller = frame.f_back

doesn't this contradict what you answered here?

src/cuga/backend/cuga_graph/nodes/cuga_supervisor/delegation.py
haroldship: Accepted, no change needed. Cutting to the most recent N values could drop `user_id`, which is the value this PR exists to keep. Copying everything on the omitted path is the right default.

src/cuga/backend/cuga_graph/nodes/cuga_supervisor/prompts/supervisor_lite_prompt.jinja2
haroldship: No longer relevant. When I asked this, `variables=["name"]` resolved to nothing, so the instruction did not match the code. After `edc25c15` the resolver reaches the script frame and falls back to the variable manager, so the instruction and the code now agree.

haroldship:
Fixed:

resolve_names_from_caller_frame now walks up to the frame that holds SUPERVISOR_EXEC_KEY instead of stopping at delegate_to_agent, and fills anything still missing from the variable manager. The three new tests cover the explicit list, the mid-script assignment winning over the stored value, and variables=[] still opting out. CI is green on this head.


Closing/agree:

the size cap on the omitted path. I raised this and I accept the answer. Trimming to the most recent N values could drop the exact value this PR is meant to keep, which is user_id. Copying everything is the right default for now. No change needed.
my question about whether the prompt points at the right code path. When I asked, variables=["name"] resolved to nothing, so the instruction and the code disagreed. After edc25c15 they agree. Nothing to do here.


Still open:

test coverage for the two external agent paths. CodeRabbit asked for this on 15 August and I want it done rather than waved through. All the tests here use the internal agent. The two external paths call the same resolver, and this PR changed what that resolver returns, so their behaviour changed as well and nothing checks it. Please add tests for both external paths covering three cases each: variables left out, an explicit list of names, and variables=[]. The two paths do different things when there is nothing to send, so assert that difference: the A2A SDK path sends None (line 162), the older path sends an empty dictionary (line 178).
please check the setting that controls sending values to outside agents. There is a setting, pass_variables_a2a, read at line 102. Only one of the two external paths checks it. Line 155 checks it before resolving. Line 172 does not — that path resolves whenever variables is not None, with no reference to the setting. This did not matter before, because the resolver always came back empty. It matters now. After this change that path will send real supervisor values to an outside endpoint even when the setting is off. There is a second part to it. The new fallback fills any missing name from the variable manager. So on the external paths a name the generated script never assigned can still be filled in from stored values and sent out. That may be what you want, but it should be a decision rather than a side effect. Either add the same setting check at line 172, or limit the variable-manager fallback to the internal path — whichever matches the intent. Please say which one you picked and why.

src/cuga/backend/cuga_graph/nodes/cuga_supervisor/tests/test_delegation_recording.py
```diff
@@ -111,37 +111,128 @@ def test_record_delegation_updates_state_fields():
...
+        )
+
+    assert mock_agent.invoke.await_args.kwargs["variables"] == {"user_id": "user_alice_99"}
+
+
+@pytest.mark.unit
+@pytest.mark.asyncio
+async def test_delegation_forwards_explicit_variables():
+    mock_agent = _fake_agent()
+    with patch("cuga.sdk.CugaAgent", _FakeCugaAgent):
```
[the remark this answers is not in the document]
sami-marreed: Added. Both external paths now cover omitted, explicit names, and `variables=[]`. SDK sends `None` when empty; legacy sends `{}`. `pass_variables_a2a` now gates the legacy path as well.

sami-marreed:
@haroldship Both open items are on this head.

Tests: both external paths now cover omitted, an explicit list, and `variables=[]`. Empty payload is `None` on the SDK path and `{}` on the legacy path.

Setting: I added the same `pass_variables_a2a` check on the legacy path. I did not limit the variable-manager fallback to the internal path.

`pass_variables_a2a` is the documented gate for sending values to an outside agent, default off. The SDK path already used it. The legacy path skipping it was only safe while the resolver returned empty. `variables=["name"]` is supposed to resolve prior-turn values from the VM; that is the same helper on every path once sending is allowed. With the setting off, neither external path sends anything.

https://github.com/cuga-project/cuga-agent/commit/8c60cdaf — setting gate
https://github.com/cuga-project/cuga-agent/commit/d445771b — tests

haroldship (approved):
All open items are addressed on c0739963.

Frame resolution (edc25c15): resolve_names_from_caller_frame now walks to the frame holding SUPERVISOR_EXEC_KEY and falls back to the variable manager, so variables=["name"] actually forwards values. Three tests cover it, including the mid-script assignment winning over the stored value.

Prompt wording (edc25c15): rule 6 now says known Variables are forwarded to internal agents, which is what the code does.

Setting gate (8c60cdaf): line 172 now matches line 155, so both external paths send only when pass_variables_a2a is on. That fits the rest of the codebase, where the setting defaults to false and controls whether variables even appears in the A2A tool signature. You kept the variable-manager fallback on the external paths and gave the reason, which is what I asked for.

External path tests (d445771b): both paths cover variables omitted, an explicit list, and variables=[], and assert the different empty payload, None on the SDK path and an empty dictionary on the legacy path. The setting-off case is covered too.

Two I closed with no change needed. The size cap on the omitted path, since trimming to the most recent N values could drop the exact value this PR exists to keep. And my question about whether the prompt points at the right code path, which edc25c15 resolved.

Approving.

Issue #624, opened 2026-08-11
[Feature]: Pluggable shortlister for CUGA Lite — embedding (cosine) and hybrid strategies


segevshlomovIBM:
Part of #612.

### What you want and why

CUGA Lite reduces a large tool catalogue before the model sees it. That reduction is **always LLM-based**, at two hardcoded call sites, with no way to swap the ranker.

| Seam | Function | Fires when | Returns |
|---|---|---|---|
| **A** — runtime discovery | `PromptUtils.find_tools` | `total_tools > advanced_features.shortlisting_tool_threshold` (**35**) | markdown + reasoning |
| **B** — bind-time cap | `PromptUtils.shortlist_tool_names` | `candidates > cuga_lite_bind_tools_max_count` (**128**) | `List[str]` top-K |

Four reasons to make this pluggable:

**1. Cost & latency.** Every `find_tools` call is a full LLM round-trip whose prompt embeds the serialized schema of *every* tool in the app (`_build_shortlister_payload` includes `args_schema`, `_response_schemas`, `_param_constraints`). On a 300-tool catalogue that is a very large prompt, paid on every discovery call. Seam B adds a second round-trip **per `call_model`** — `helpers/bind_tools.py` documents this cost explicitly.

**2. Accuracy — we have evidence this is a top failure mode.** Closed issue #150 (AppWorld trace analysis) rated shortlister mis-selection **high severity** and the *first divergence* in many failed tasks:

> - `Place an order for all wrench sets in my amazon wish list` → Step 3 shortlister picked **product search** instead of wishlist tools
> - `By when should everything from my yesterday's amazon order arrive` → shortlister emphasized **order-creation** instead of order-lookup
> - `How many people follow the artist of the currently playing song` → shortlister **omitted the artist-followers endpoint**

#150 was closed with "review the prompt" — the accuracy problem has only ever been attacked prompt-side. A hybrid (cosine prefilter → LLM rerank) attacks it from the retrieval side, and is independently measurable.

**3. Model coupling.** Seam B's shortlister runs on *the same model being bound*. A model without native structured output cannot shortlist at all — that is precisely why `BindToolsUnsupportedError` exists (`bind_tools/cap.py`). A non-LLM ranker removes the coupling: shortlisting keeps working on weak/local models.

[...]

Issue #150, opened 2026-04-16
[Enhancement]: Shortlister enhancement


Sergey-Zeltyn:
Analysis of failed Appworld logs via trace comparison tool indicated shortlister issues as a top priority. 

The tool provided the following diagnostic:

"**description":** "Shortlister frequently surfaced or prioritized tools that did not align with the task’s core intent, causing the first divergence in many tasks. 
**Typical patterns:** choosing generic product search over wishlist retrieval; surfacing cart-manipulation tools when the intent required checkout; omitting critical endpoints (e.g., place order or add-to-cart) needed to complete workflows."

**"representative_examples":** 
- "Task: Place an order for all wrench sets in my amazon wish list — Step 3 Shortlister picked product search instead of wishlist tools",
 - "Task: Everything in my amazon cart is for my friend, Denise... — Step 4 Shortlister listed many cart CRUD tools instead of just showing cart",
 - "Task: How many people follow the artist of the currently playing song — Step 3 Shortlister omitted the artist-followers endpoint in the successful sequence",
 - "Task: Update all my Spotify playlist titles... — Step 3 Shortlister listed playlist listing tools instead of track and title update mix",
 - "Task: By when should everything from my yesterday's amazon order arrive — Step 3 Shortlister emphasized order-creation tools instead of order-lookup",
 - "Task: Add all the songs from Lily Moon... — Step 3 Shortlister missed the add-to-queue endpoint initially",
 - "Task: Songs of which genre have I liked the most — Step 3 Shortlister selected liked_songs without the detail endpoint that contains genre"
 
**"severity":** "high"

**Plan**
- Review shortlister prompt and check if some of these issues can be solved
- Perform rigorous evaluation of improvements

PR #661, opened 2026-08-14
refactor(cuga-lite): extract a pluggable shortlister strategy seam (#624)

segevshlomovIBM:
> **Reviewing this stack?** Start with this PR only — it is a pure refactor with a byte-identical-output proof in the comments below. #662 and #663 are drafts until this merges (CI and CodeRabbit do not run against a feature-branch base).

## Feature

Refs #624 — part 1 of 3. **Merge order: this → #B → #C.**

### Summary

Tool shortlisting is hardcoded to the LLM at two call sites — `find_tools` (runtime discovery) and the `bind_tools` provider cap — with no way to swap the ranker. This routes both through a `ShortlisterStrategy` protocol resolved from `[shortlister]` settings, a `configurable` key, or a dotted class path.

**Pure refactor: no behavior change and no new ranker.** The only built-in strategy is `llm`, a faithful port of the existing chain — same prompt template, same schema, same model, same composed query string, unknown names still silently dropped by the caller rather than retried. Cosine ranking (#B) and the SDK surface (#C) land on top of this.

`PromptUtils.find_tools` and `shortlist_tool_names` keep their signatures, so `cap.py`, `bind_tools.py`, `graph_adapter.py` and `prepare_node.py` need no changes. `find_tools` gains an optional `task_context` so the step query and initial user message can travel separately; the LLM strategy re-joins them into the identical string.

`prompt_utils.py` shrinks ~100 lines net — the markdown half of `find_tools` moves to `render.py` so ranking and presentation can vary independently.

Also fixes a latent bug: `advanced_features.shortlisting_tool_threshold` had no dynaconf validator, and `prepare_node` reads it without a `getattr` guard, so a settings file lacking the key raised `AttributeError` mid-run.

### Testing
- [x] Tested locally; tests pass — 223 passed (includes all pre-existing shortlister/bind-tools suites unchanged)

One existing test changed: `test_find_tools_composes_query_with_initial_user_message` asserted `find_tools` receives a *pre-joined* query string. Replaced with two tests — one pinning the new plumbing, one asserting `compose_query` is byte-identical to the old `_compose_find_tools_shortlister_query`, so the LLM prompt provably did not change.

## Summary by CodeRabbit

- **New Features**
  - Added configurable tool discovery and ranking strategies.
  - Tool selection can now consider both the current query and conversation context.
  - Added support for fallback strategies when the preferred ranking method is unavailable.
  - Improved tool result formatting with descriptions, parameters, response details, and schemas.

- **Bug Fixes**
[...]

PR #659, opened 2026-08-14
feat(cuga-lite): pluggable shortlister with cosine and hybrid strategies (#624)

segevshlomovIBM:
## Feature

Closes #624

> **Reviewing this?** Jump to the **Reviewer guide** below — 28 files, but only 3 need real scrutiny, and there is a mechanical proof that the refactor half is a no-op.

### Summary

Tool shortlisting is hardcoded to the LLM at two call sites — `find_tools` (runtime discovery) and the `bind_tools` provider cap — with no way to swap the ranker. This routes both through a `ShortlisterStrategy` protocol selectable by config name, dotted class path, or the SDK.

**Default behavior is unchanged.** `strategy = "llm"` is a faithful port of the existing chain, and the cosine stage is additionally gated behind `threshold = 128`, so catalogues at or below that size are untouched whatever the strategy.

| Strategy | Ranking | Cost |
|---|---|---|
| `llm` (default) | asks the model | one LLM call per shortlist |
| `embedding` | local cosine similarity | no LLM call, ~67ms warm |
| `hybrid` | cosine cuts to `top_k`, LLM picks | one LLM call, much smaller prompt |

**Why:** #150 (AppWorld trace analysis) rated shortlister mis-selection a **high-severity top cause of first divergence** — generic product search over wishlist retrieval, order-creation over order-lookup — and was closed with "review the prompt", so accuracy has only ever been attacked prompt-side. `hybrid` attacks it from the retrieval side. Separately, the bind-cap shortlister runs on *the same model being bound*, which is why `BindToolsUnsupportedError` exists; a non-LLM ranker removes that coupling.

No new dependencies (`fastembed` is already core). Embeddings default to local `all-MiniLM-L6-v2` — no network call, nothing billed.

### Reviewer guide

**Only 3 files need real scrutiny.**

| Priority | File | What to check | Lines |
|---|---|---|---|
| 🔴 1 | `cuga_lite/prompt_utils.py` | The only edit to existing logic. Both public functions keep their signatures and delegate; **net −100 lines** because rendering moved out | +84 / −182 |
| 🔴 2 | `shortlister/llm.py` | Faithful port of the chain that lived in `prompt_utils`? (proof below) | new, 100 |
| 🟠 3 | `shortlister/render.py` | Moved verbatim from `find_tools`; skim for drift — this output is what the agent reads from stdout | new, 128 |
[...]

PR #662, opened 2026-08-14
feat(cuga-lite): add cosine and hybrid shortlister strategies (#624)

segevshlomovIBM:
## Feature

Refs #624 — part 2 of 3. **Stacked on the seam PR; review/merge that first.**

### Summary

Adds the two non-LLM rankers on top of the strategy seam. `embedding` ranks by cosine similarity with no LLM call; `hybrid` uses it to prefilter to `top_k` and lets the LLM make the final pick. Both opt-in — `strategy` stays `"llm"` by default.

Motivation is #150: AppWorld trace analysis rated shortlister mis-selection a **high-severity top cause of first divergence**, and it was closed with "review the prompt" — so accuracy has only ever been attacked prompt-side. `hybrid` attacks it from the retrieval side. Separately, the bind-cap shortlister runs on *the same model being bound*, which is why `BindToolsUnsupportedError` exists; a non-LLM ranker removes that coupling.

**Gated behind `threshold = 128`** — the existing `cuga_lite_bind_tools_max_count`, so one number means one thing. At or below that many candidates the cosine stage does not run and both seams behave exactly as before, whatever the strategy is set to.

No new dependencies: `fastembed` is already core. Embeddings default to local `all-MiniLM-L6-v2` — no network call, nothing billed.

<details>
<summary>What cosine compares, and its known limitation</summary>

One vector from the question against one per tool, built from **name + description + parameter names + return field names**. Tool names are machine-generated (`crm_get_contacts_contacts_get`) and embed poorly until split back into words; CRM-style endpoints often carry no description at all, so parameter text is frequently the only natural language available. `args_schema` JSON is excluded — it makes the LLM prompt expensive and dilutes a fixed-size vector.

Query and task context are blended as **weighted unit vectors**, not concatenated: a long first message would otherwise swamp a short per-step query and make every step of a task retrieve the same tools.

Cosine is a **recall** device, not a precision one — `get_contacts` (list) and `get_contact` (by id) embed almost identically, exactly the #150 failure class. A stability test measures recall and asserts the CRUD-sibling margin stays narrow, so promoting `embedding` to final selection requires re-measuring rather than assuming.
</details>

Two safety properties worth review: `top_k` is a ceiling the caller sets and config can lower but never raise (so the bind cap can't be pushed past what the provider accepts), and `find_tools` caps rendered results at `max_results` (10) because 128 fully-rendered tools would exceed `execution_output_max_length` and be silently truncated mid-render.

A query never waits on a model download: if the backend is not resident the call raises `ShortlisterUnavailableError`, the caller degrades to the LLM, and a background thread loads it — mirroring `knowledge/reranker.py`.

### Testing
- [x] Tested locally; tests pass — 286 passed, plus 8 stability tests against the real model

Real-model accuracy on a 212-tool catalogue using real machine-generated name shapes: precision@1 4/4, recall@5 4/4, ~67ms warm, 1.9s to embed 212 tools.

### DIFF
--- a/src/cuga/backend/cuga_graph/nodes/cuga_lite/prompt_utils.py
+++ b/src/cuga/backend/cuga_graph/nodes/cuga_lite/prompt_utils.py
@@ -354,6 +354,15 @@ class PromptUtils:
         plan = ShortlisterRouter.resolve(
             settings, seam="discovery", configurable=_configurable_of(run_config)
         )
+        # Below the threshold the cosine stage would add cost without cutting
+        # anything, so the LLM ranks the full set exactly as it does today.
+        # Uniform on purpose: "at or below threshold, shortlisting behaves as it
+        # always did" is easier to reason about than per-strategy exceptions.
+        # Set ``threshold = 0`` to always engage the configured strategy.
+        engage_cosine = len(all_tools) > plan.threshold
+        if not engage_cosine:
+            plan = plan.model_copy(update={"strategy": "llm", "instance": None})
+
         candidates = await run_shortlister(
             plan,
             ShortlistRequest(
@@ -361,6 +370,8 @@ class PromptUtils:
                 tools=all_tools,
                 apps=all_apps,
                 task_context=task_context,
+                top_k=plan.top_k if engage_cosine else None,
+                max_results=plan.max_results if engage_cosine else None,
                 llm=llm,
                 run_config=run_config,
             ),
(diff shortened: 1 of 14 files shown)


segevshlomovIBM:
**Stack (merge in order):**
1. #661 — strategy seam, no behavior change
2. #662 — cosine + hybrid strategies
3. #663 — SDK config + docs

Supersedes #659, which contained all three as one ~4k-line PR.

segevshlomovIBM:
**Draft until #661 merges — please review #661 first.**

Marked draft deliberately: while this PR targets a feature branch, CI reports `skipping` and CodeRabbit reports *"reviews are disabled for this base branch"*. Reviewing it now would mean reviewing something no check has validated.

Once its base merges, GitHub retargets this to `main` automatically, CI and CodeRabbit run for real, and I'll mark it ready. Locally it is green: `286 passed + 8 stability`.

Stack: #661 → #662 → #663.

segevshlomovIBM:
Consolidated into #664 — a single PR against `main`, to get this reviewed and merged in one cycle rather than three.

Same content and same commits (this branch's three commits are unchanged and can still be reviewed one at a time). #664 carries over the reviewer guide and the byte-identical-output proof from this PR.

Closing to keep one review thread.

Issue #660, opened 2026-08-14
[Feature]: Expose shortlister strategy through the server and demo UI


segevshlomovIBM:
Part of #612.

### What you want and why

#659 made tool shortlisting pluggable (`[shortlister] strategy = llm | embedding | hybrid`), reachable from **settings.toml, env vars, `configurable`, and the SDK** — but **not from the server / demo UI**. Anyone driving CUGA through the demo or the manage-routes API cannot switch strategy, so the feature is invisible to exactly the audience most likely to try it casually.

This also blocks measurement. Comparing `llm` vs `hybrid` on Vakra/M3 is far easier if the strategy is a UI toggle than if it needs a settings edit and a restart per run.

The neighbouring setting `advanced_features.shortlisting_tool_threshold` is already threaded through this whole chain, so this is a fill-in-the-pattern change rather than new design.

### How it could work

Follow `shortlisting_tool_threshold` exactly, end to end:

1. **`server/manage_routes/helpers.py`** — `extract_agent_feature_overrides` currently returns a fixed dict of five keys (`enable_todos`, `reflection_enabled`, `shortlisting_tool_threshold`, `cuga_lite_max_steps`, `enable_filesystem_tools`), reading each from `feature_flags` then falling back to `advanced_features`. Add the shortlister keys the same way — at minimum `shortlister_strategy`, `shortlister_top_k`, `shortlister_max_results`.
2. **`helpers.py`** feature-flag defaults block — the sibling of the `if "shortlisting_tool_threshold" not in ff:` branch, so the UI receives current values rather than blanks.
3. **`server/manage_routes/draft_ops.py`** — the `if overrides[...] is not None: agent.<attr> = ...` sequence; assign the shortlister overrides onto the agent before `await agent.build_graph()`.
4. **`utils/agent_loop.py`** — constructor param + `self.<attr>` + the block that writes into `config["configurable"]` (where `shortlisting_tool_threshold` is set today).
5. **`cuga_graph/graph.py`** — the matching constructor param and assignment.
6. **Frontend** — a strategy selector alongside the existing feature-flag controls.

The plumbing target is the same `shortlister_*` `configurable` keys the SDK already emits, so no new resolution logic is needed: `ShortlisterRouter.resolve` already reads them. `shortlister_to_configurable()` in `cuga_lite/shortlister/config.py` produces exactly that key set and can be reused rather than hand-rolling names.

[...]

Issue #641, opened 2026-08-12
[Bug]: Trim the duplicated schemas in find_tools (free, no new machinery)


sami-marreed:
Part of #616.

## What happened

`find_tools` discovery output duplicates the same schema information twice: **Input Schema** / **Output Schema** (JSON) and **Parameters** / **Response Schema**. The JSON forms alone are ~55–65% of discovery output and carry no additional information.

Measured shares of discovery output:

| Task | Input Schema | Output Schema | Combined |
|---|---|---|---|
| `80acbaf_1` | 32.6% | 32.0% | **64.6%** |
| `6d59d90_1` | 22.3% | 36.4% | **58.7%** |

Expected: emit each schema once (keep Parameters + Response Schema, drop the duplicate Input/Output Schema JSON — or the reverse), with no new machinery.

## How to reproduce

1. Run AppWorld / CugaLite tasks that call `find_tools` (e.g. `80acbaf_1`, `6d59d90_1`).
2. Inspect discovery / shortlist prompt or tool-result text for each matched tool.
3. Observe both Input/Output Schema JSON blocks and Parameters/Response Schema sections describing the same fields.

## Environment

AppWorld evals / CugaLite tool discovery (`find_tools`).

## Logs, screenshots, or config

Token-share breakdown from discovery output on `80acbaf_1` and `6d59d90_1` (above). Fix should be a free win: strip the duplicate rendering path only.

## Checklist

- [x] I searched existing issues and this doesn’t look like a duplicate

PR #664, opened 2026-08-14
feat(cuga-lite): pluggable shortlister with cosine and hybrid strategies (#624)

### LINKED ISSUE #624: [Feature]: Pluggable shortlister for CUGA Lite — embedding (cosine) and hybrid strategies

segevshlomovIBM:
## Feature

Closes #624

> **Reviewing this?** Jump to the **Reviewer guide** below — 28 files, but only 3 need real scrutiny, and there is a mechanical proof that the refactor half is a no-op.

### Summary

Tool shortlisting is hardcoded to the LLM at two call sites — `find_tools` (runtime discovery) and the `bind_tools` provider cap — with no way to swap the ranker. This routes both through a `ShortlisterStrategy` protocol selectable by config name, dotted class path, or the SDK.

**Default behavior is unchanged.** `strategy = "llm"` is a faithful port of the existing chain, and the cosine stage is additionally gated behind `threshold = 128`, so catalogues at or below that size are untouched whatever the strategy.

| Strategy | Ranking | Cost |
|---|---|---|
| `llm` (default) | asks the model | one LLM call per shortlist |
| `embedding` | local cosine similarity | no LLM call, ~67ms warm |
| `hybrid` | cosine cuts to `top_k`, LLM picks | one LLM call, much smaller prompt |

**Why:** #150 (AppWorld trace analysis) rated shortlister mis-selection a **high-severity top cause of first divergence** — generic product search over wishlist retrieval, order-creation over order-lookup — and was closed with "review the prompt", so accuracy has only ever been attacked prompt-side. `hybrid` attacks it from the retrieval side. Separately, the bind-cap shortlister runs on *the same model being bound*, which is why `BindToolsUnsupportedError` exists; a non-LLM ranker removes that coupling.

No new dependencies (`fastembed` is already core). Embeddings default to local `all-MiniLM-L6-v2` — no network call, nothing billed.

### Reviewer guide

**Only 3 files need real scrutiny.**

| Priority | File | What to check | Lines |
|---|---|---|---|
| 🔴 1 | `cuga_lite/prompt_utils.py` | The only edit to existing logic. Both public functions keep their signatures and delegate; **net −100 lines** because rendering moved out | +84 / −182 |
| 🔴 2 | `shortlister/llm.py` | Faithful port of the chain that lived in `prompt_utils`? (proof below) | new, 100 |
| 🟠 3 | `shortlister/render.py` | Moved verbatim from `find_tools`; skim for drift — this output is what the agent reads from stdout | new, 128 |
| 🟡 4 | `shortlister/{plan,factory,base,doc,embedding,hybrid,config}.py` | New, self-contained. No existing behavior touched | new |
| ⚪ 5 | `config.py`, `settings.toml`, `sdk.py`, `__init__.py` | Config surface + a **bug fix**: `shortlisting_tool_threshold` had no validator and `prepare_node` reads it without a `getattr` guard | small |
| ⚪ 6 | tests, README, `docs/design/` | ~2.6k of the 4k lines | — |

**Deliberately untouched** — worth confirming from the file list: `bind_tools/cap.py`, `helpers/bind_tools.py`, `adapter/graph_adapter.py`, `adapter/prepare_node.py`, `prompts/shortlister/system.jinja2`.

### Proof the refactor is a no-op

Stubbing `BaseAgent.get_chain` (no credentials needed), I captured on **`main`** and on **this branch**, at both call sites: rendered system prompt, structured schema, prompt message structure, exact `input`/`instructions` strings, full serialized tool payload, final rendered markdown, and returned names.

**Byte-identical** — same sha256 after normalizing Python object addresses in a `repr` (per-process noise, not prompt content).

```
system prompt sha   : ba2b45656d43346e6d235236…   (unchanged)
structured schema   : ShortListerOutputLite       (unchanged)
prompt messages     : System, Human, AI, Human    (unchanged)
find_tools instrs   : ''                          (unchanged)
bind-cap instrs     : 'Return the 2 most relevant tools (or fewer if not enough are relevant), …'
markdown bytes      : 2117                        (identical output)
hallucinated name dropped : True                  (same silent-drop behavior as main)
```

That last line matters: at the time this was captured, `main` silently dropped names matching no tool, and so did this — no retry logic was duplicated from what was then #549's unreleased work.

**#549 has since merged.** The branch now calls its `_ainvoke_shortlister_with_name_validation` from `LLMShortlister`, so its bounded retries are preserved exactly and its own test suite passes unmodified. The equivalence capture above therefore describes the pre-#549 baseline; the seam itself is unchanged by that merge, and cosine deliberately does not inherit the retry — an embedding ranker draws names from the candidate list and cannot invent one.

### Testing

```
303 passed   # shortlister + all pre-existing shortlister/bind-tools suites
  8 passed   # stability, against the real embedding model
```

**Full suite, `main` vs this branch, back to back:** `main` **118 failed / 2256 passed**; this branch **97 failed / 2374 passed**. Set difference (branch − main) is **empty** — no new failures. Pre-existing ones are cross-test DB pollution in `sdk_core` policies / `test_output_schema` / `test_filesystem_sync`, which pass in isolation on both.

Verified end to end against a real LLM and real embeddings, no mocks: cosine ranking through both seams; the threshold gate leaving sub-threshold catalogues on the LLM path with zero vectors embedded; fallback to the LLM when a backend cannot load; per-call SDK overrides.

CI passed all 13 checks (including CodeRabbit) on the seam commit as PR #661 before consolidation.

### Configuration

Documented in the README with every option, four ways to set it (settings.toml incl. per-seam sections, env, SDK per-agent and per-call, raw `configurable`), a which-strategy-when table, and gotchas.

```toml
[shortlister]
strategy = "hybrid"   # llm | embedding | hybrid | "my.pkg.MyShortlister"
threshold = 128       # engage the cosine stage only above this many candidates
top_k = 128
max_results = 10
```

### Defaults, and why each one

Shipped values, all in `[shortlister]`. The first row is the one that matters: **nothing about the existing behaviour changes until someone opts in.**

| Key | Default | Why this value |
|---|---|---|
| `strategy` | `"llm"` | The feature ships **inert**. This is a faithful port of the existing chain — same prompt, schema, model and composed query — so a deployment that does nothing sees no change at all. Cosine is opt-in, not opt-out. |
| `threshold` | `128` | Not a new number: it is the existing `cuga_lite_bind_tools_max_count`, so one value means one thing across both seams — *"more than 128 candidates is too many"*. **At or below 128, the cosine stage never runs**, which confines the blast radius to large catalogues, exactly where the LLM ranker is slowest and least accurate. `0` disables the gate. |
| `top_k` | `128` | What the cosine stage keeps, and the `hybrid` prefilter width. Matches the provider cap so seam B needs no second number. At the bind cap this may only ever **lower** the caller's cap, never raise it past what the provider accepts. |
| `max_results` | `10` | `find_tools` only, and it exists for a concrete reason: each rendered tool carries description, reasoning, parameter docs and both JSON schemas — roughly 400–1500 chars. 128 of them exceeds `advanced_features.execution_output_max_length` and gets **silently truncated mid-render**, so the rendered count is capped separately from `top_k`. |
| `min_score` | `0.15` | Deliberately low. Cosine here is a **recall** filter, not a precision knob — `get_contacts` and `get_contact` are separated by ~0.03 on real data, so a high floor would drop needed tools without meaningfully improving the pick. Precision is the LLM's job in `hybrid`. |
| `query_weight` | `0.7` | Step query vs. initial user message when embedding the query. The step query dominates while the task context still disambiguates. Explicit, rather than letting the weighting fall out of how long each string happens to be — which is what string concatenation would have done. |
| `embedding_provider` | `"local"` | Ranking stays offline and unbilled by default. This section deliberately does **not** inherit `[storage.embedding]`, which a deployment may already have pointed at OpenAI. |
| `embedding_model` | `BAAI/bge-small-en-v1.5` | The same weights knowledge and policy already load, so there is **one** ONNX session per process rather than two, and `preload_models.py` already fetches it — airgapped servers need no new download. |
| `fallback_strategy` | `"llm"` | What runs when the chosen strategy cannot (model still downloading, missing dependency, unknown provider). Degrading to today's behaviour is always safe. |

Two guardrails worth noting, because they are what make the first row true:

- The cosine gate requires **both** a non-default strategy **and** a catalogue above `threshold`. Keying on size alone would have capped the default LLM path above 128 — pinned now at N = 10 / 129 / 400.
- `warm_tool_vectors()` and the server startup/reload hooks are **no-ops** on the default strategy, so a deployment that has not enabled cosine loads no model and embeds nothing.

### Notes for review

- **One existing test changed.** `test_find_tools_composes_query_with_initial_user_message` asserted `find_tools` receives a *pre-joined* query string; query and task context now travel separately so a non-LLM ranker can weight them. Replaced with two tests — one pinning the new plumbing, one asserting `compose_query()` is byte-identical to the old `_compose_find_tools_shortlister_query()`. Behavior unchanged, test contract changed — flag it if you'd rather keep the old shape.
- **Not included, deliberately:** `prompt_preselect` and a per-seam LLM `model` key. `docs/design/pluggable-shortlister.md` records each omission and why. (#549 has since merged and its name-validation retry is now integrated into the LLM strategy; catalogue warm-up shipped in response to review, so the embed/sync hook no longer waits on #660 — only the UI strategy toggle does.)
- **Commits are split by concern** (seam / strategies / SDK+docs, then the two review-response commits) if you would rather review them one at a time — the seam commit is the pure refactor and stands alone.

## Summary by CodeRabbit

* **New Features**
  * Added configurable tool shortlisting with LLM, embedding, hybrid, and custom strategies.
  * Added thresholding, ranking, caching, fallbacks, result limits, and task-context-aware selection.
  * Added TOML, environment, SDK, and per-call configuration options.
  * Added ranked Markdown tool rendering with descriptions, reasoning, parameters, and schemas.
* **Documentation**
  * Added configuration, strategy, rollout, and customization guidance.
* **Tests**
  * Added coverage for ranking, configuration, caching, fallbacks, rendering, embedding behavior, and exception handling.


segevshlomovIBM:
Addressed the CodeRabbit pass — pushed as `90308392`.

**Two were real defects in claims this PR makes**, both now fixed with regression tests:

1. **`max_results` was enforced only inside `EmbeddingShortlister`.** So with `strategy = "hybrid"` — the setting I recommend for discovery, which ends on an LLM pick — the render cap did nothing. That cap is what stops `find_tools` output exceeding `execution_output_max_length` and being silently truncated mid-render, so the protection I documented was not actually in effect for the recommended configuration. Now enforced at the render boundary so it covers every strategy including custom ones, but only when the cosine stage is engaged, so the default LLM path keeps its historical "no fixed result count" behavior. Both halves are pinned by tests.

2. **`tests/unit/test_find_tools_exception.py` was not in any CI invocation** — the one existing test this PR modifies would not have run. Added to the workflow.

Also fixed: empty query no longer reported as backend unavailability (it wrongly triggered the fallback, which would get the same empty query); `query_weight`/`min_score` clamped to their documented ranges; unused `_NOISE_SUFFIXES` removed; the CRUD-sibling stability test no longer `KeyError`s if a sibling drops out of the top 2; embedding tests assert on query/passage channels rather than call order; README documents the real precedence order and qualifies "no LLM call" as "no LLM call *for ranking*"; and the design doc no longer presents `ShortlistResult`, `prewarm`, `model` or `prompt_preselect` as shipped.

Re-verified: **305 unit + 8 stability passing**, plus the 17-check end-to-end run against a real LLM and real embeddings. Lint clean.

The remaining CodeRabbit items are nitpicks I left alone deliberately — happy to take them if you'd prefer.

segevshlomovIBM:
All 13 CodeRabbit findings are now addressed — `0c581b10`. Every review thread is resolved.

The last pass covered the three that survived the first fix:

- **README** — the offline/free claim now says *ranking* makes no network call and is not billed, notes that `embedding_provider = "openai"` trades that away, and clarifies that the agent still calls an LLM for its own reasoning regardless.
- **Design doc** — the second `ShortlisterPlan` block still listed `model` and `prewarm` as if they shipped (my earlier fix patched the other block); a stale "cuts the prompt to 50" now reads `top_k`; and the `ShortlisterUnavailableError` description is scoped to genuine backend failure, since an empty query returns no candidates rather than claiming unavailability.
- **SDK** — `shortlister` documented in the `invoke()` and `stream()` docstrings, and `_apply_shortlister` typed `Optional["Shortlister"]` rather than `Optional[Any]`.

Green after the fixes: **305 unit + 8 stability**, lint clean, and the end-to-end run against a real LLM and real embeddings still passes all 17 checks.

Ready for a maintainer pass whenever you have a moment.

segevshlomovIBM:
@coderabbitai review

segevshlomovIBM:
@coderabbitai full review

segevshlomovIBM:
Full re-review addressed — `d208d103`.

The `full review` pass was worth running: it found **two real defects the incremental pass missed**, both now fixed with regression tests.

**1. `ShortlisterPlan.cache_key` omitted `min_score`.** `EmbeddingShortlister` takes it as a constructor argument and instances are cached by `cache_key()`, so two plans differing only in `min_score` shared one instance — a per-invoke `shortlister_min_score` override silently did nothing after the first resolution. Every constructor field is now part of the key; `top_k` and `max_results` stay out because they travel per request, not per instance.

**2. `_build_from_dotted_path` probed the constructor with `try: cls(plan=plan) except TypeError: cls()`.** A `TypeError` raised *inside* a constructor that does accept `plan` was indistinguishable from one that does not, so a broken custom strategy was silently rebuilt unconfigured — or its real error was masked. The signature is inspected instead.

Docs, all three valid:
- The design doc claimed `top_k <= 0` *"disables the cosine stage rather than returning nothing"*. It returns no candidates. Corrected, so custom-strategy authors do not implement the opposite.
- The README precedence list had raw `configurable` keys **below** the SDK objects. Since `_apply_shortlister` merges with `setdefault`, raw keys actually win — I had it backwards. Rewritten as an ordered list with the reason stated.
- Added the required `unit` marker to `test_compose_query_matches_legacy_find_tools_format`.

Green: **308 unit + 8 stability**, lint and format clean.

src/cuga/backend/cuga_graph/nodes/cuga_lite/prompt_utils.py
```diff
@@ -432,151 +454,49 @@ async def find_tools(
...
-            run_config=run_config,
+        plan = ShortlisterRouter.resolve(
+            settings, seam="discovery", configurable=_configurable_of(run_config)
         )
+        # Below the threshold the cosine stage would add cost without cutting
+        # anything, so the LLM ranks the full set exactly as it does today.
+        # Uniform on purpose: "at or below threshold, shortlisting behaves as it
+        # always did" is easier to reason about than per-strategy exceptions.
+        # Set ``threshold = 0`` to always engage the configured strategy.
+        engage_cosine = len(all_tools) > plan.threshold
```
sami-marreed: `engage_cosine` is keyed on `len(all_tools) > threshold`, not on whether cosine is actually configured.

With the shipped defaults (`strategy = "llm"`, `threshold = 128`, `top_k = 128`, `max_results = 10`) and a 200-tool catalogue, we still call the LLM — but we now pass `top_k=128` (historical prompt had empty instructions) and then slice the render to 10. Reproduced: N=128 stays uncapped; N=129 renders 10. `#624` required the LLM path byte-for-byte unchanged.

`test_default_llm_path_keeps_no_fixed_result_count` uses 20 tools, so it never hits this.

Please gate `top_k` / `max_results` / the strategy rewrite on the resolved strategy being `embedding` or `hybrid`, not on catalogue size. Leave default `llm` uncapped at every N.

src/cuga/backend/cuga_graph/nodes/cuga_lite/prompt_utils.py
```diff
@@ -432,151 +454,49 @@ async def find_tools(
...
+                tools=all_tools,
+                apps=all_apps,
+                task_context=task_context,
+                top_k=plan.top_k if engage_cosine else None,
+                max_results=plan.max_results if engage_cosine else None,
+                llm=llm,
+                run_config=run_config,
+            ),
+        )
+        candidates = result.candidates
```
sami-marreed, line 644: Both seams assume `ShortlistResult` (`result.candidates`, `result.notes`). `EmbeddingShortlister.shortlist` returns a list. After the model is warm, `strategy = "embedding"` AttributeErrors here.

Seam A swallows it into an error string for the agent. Seam B (`shortlist_tool_names`) raises. Reproduced through `PromptUtils` with a fake embedder.

Tests mock `shortlist` to return `ShortlistResult()`, so CI does not catch this. Please return `ShortlistResult` from every strategy and add an unmocked seam test through `PromptUtils.find_tools` / `shortlist_tool_names`.

src/cuga/backend/cuga_graph/nodes/cuga_lite/shortlister/embedding.py
```diff
@@ -0,0 +1,353 @@
...
+
+        if not kept:
+            # Nothing cleared the floor. Returning nothing is worse than
+            # returning the best guesses: the agent has no next move, and the
+            # bind-time cap raises on an empty ranking.
+            fallback = min(_MIN_FALLBACK_RESULTS, limit, len(tools))
+            kept = [(int(i), float(scores[i])) for i in order[:fallback]]
+        return kept
+
+    async def shortlist(self, request: ShortlistRequest) -> List[ShortlistCandidate]:
```
sami-marreed: Must return `ShortlistResult`, not `List[ShortlistCandidate]`. Empty cases `return []` will also fail `result.candidates` in `prompt_utils`.

This is the contract that `#549` introduced (`notes` for the filtered-names footer). LLM was updated; embedding was not.

src/cuga/backend/cuga_graph/nodes/cuga_lite/shortlister/hybrid.py
```diff
@@ -0,0 +1,68 @@
...
+class HybridShortlister:
+    """Embedding prefilter to ``top_k``, then :class:`LLMShortlister` ranks."""
+
+    name: ClassVar[str] = "hybrid"
+
+    def __init__(self, embedding: EmbeddingShortlister, llm: LLMShortlister) -> None:
+        self._embedding = embedding
+        self._llm = llm
+
+    async def shortlist(self, request: ShortlistRequest) -> List[ShortlistCandidate]:
```
sami-marreed: Empty tools `return []` has the same seam bug. The success path happens to work only because it forwards `LLMShortlister`’s `ShortlistResult`.

Line 56 iterates the embedding result as a list (`for c in prefiltered`). Reproduced: if embedding is wrapped in `ShortlistResult` without updating this to `prefiltered.candidates`, this becomes `TypeError: 'ShortlistResult' object is not iterable`.

src/cuga/backend/cuga_graph/nodes/cuga_lite/shortlister/embedding.py
```diff
@@ -0,0 +1,353 @@
...
+    ``BAAI/bge-*`` models are trained with a query-side instruction prefix.
+    ``all-MiniLM-L6-v2`` (our default) is symmetric — applying bge's prefix to
+    it *degrades* results. Default to symmetric so an unknown model behaves
+    sanely rather than oddly.
+    """
+    lowered = (model_name or "").lower()
+    return "bge" in lowered and "reranker" not in lowered
+
+
+class _LocalBackend:
```
sami-marreed, line 88: This is a second fastembed loader, not the existing one.

Knowledge: `_FastEmbedEmbeddings` (`BAAI/bge-small-en-v1.5`).
Policy: `embedding_service._create_local` (`_embedding_model_cache`, same bge-small).
Here: new `TextEmbedding(...)` into `_MODELS`, default MiniLM.

Same `FASTEMBED_CACHE_PATH` on disk, two ONNX sessions in process. `preload_models.py` only preloads bge-small, so airgapped server images cannot load MiniLM (`local_files_only`); cosine then always falls back to LLM.

Please reuse `create_embedding_function` / the cached local `TextEmbedding`, default to `BAAI/bge-small-en-v1.5`. Do not inherit `[storage.embedding]` when that section is `openai`. Required if cosine is a supported server strategy.

src/cuga/backend/cuga_graph/nodes/cuga_lite/shortlister/embedding.py
```diff
@@ -0,0 +1,353 @@
...
+        logger.warning("Shortlister embedding backend {!r} failed to load: {}", key, e)
+        return False
+    with _LOCK:
+        _MODELS[key] = backend
+        _RETRY_AFTER.pop(key, None)
+    logger.info("Shortlister embedding backend ready: {}", key)
+    return True
+
+
+def ensure_loading(provider: str, model_name: str) -> None:
```
sami-marreed, line 187: Query-lazy + LLM fallback is correct for the SDK. It is not enough for server mode when cosine is enabled.

Knowledge and policy already load embeddings in `lifespan`. Tool onboard (`PATCH /config/draft/tools`) already reloads the registry and calls `rebuild_agent_from_config` — it never embeds.

Please:

- Server `lifespan`: load the shared fastembed session and embed the current tool catalogue.
- Onboard / registry reload: embed new and changed tools only (content-hash). Do not reload the model.
- Query time: embed the question plus cache misses only.
- SDK: keep lazy.

Otherwise the first `find_tools` after boot or after adding tools is served by `llm`, and new tools are not in sync until a later request happens to miss the cache.

sami-marreed (changes requested):
## Request changes

The seam is well done: default `llm` below 128, `#549` retries stay on the LLM strategy, rank vs render is a clean split, and CI is green. Two bugs still break the claims this PR makes. Two more items are in scope if cosine is actually used on the server.

Reproduced against `015905e22` with a fake embedder through `PromptUtils` (PR unit tests stay green — they never put a real `EmbeddingShortlister` through the seams).

### Must fix (correctness)

**1. `embedding` cannot be consumed by either seam**

`find_tools` and `shortlist_tool_names` use `result.candidates` (`ShortlistResult`). `EmbeddingShortlister` returns a list. After the model is warm, `strategy = "embedding"` AttributeErrors. Seam A swallows it; seam B raises. Hybrid’s success path only works because it forwards the LLM result; empty hybrid still `return []`. If embedding is wrapped in `ShortlistResult` without updating hybrid, `for c in prefiltered` becomes `TypeError: 'ShortlistResult' object is not iterable`.

Tests mock `shortlist` to return `ShortlistResult()`, so CI does not catch this. Return `ShortlistResult` from every strategy, have hybrid read `prefiltered.candidates`, and add one unmocked seam test through `PromptUtils`.

**2. Default `llm` is not unchanged above 128**

`engage_cosine = len(all_tools) > threshold` is keyed on catalogue size, not on whether cosine is configured. With shipped defaults, 128 tools stay uncapped; 129 tools still use the LLM but inject `top_k=128` and truncate the render to 10. `#624` required the default LLM path byte-for-byte unchanged. `test_default_llm_path_keeps_no_fixed_result_count` uses 20 tools, so it never hits this.

Gate `top_k` / `max_results` / the strategy rewrite on `embedding` or `hybrid` only. Leave default `llm` uncapped at every N.

AppWorld `test-easy` stays a non-blocking smoke — after (2), not instead of it.

### Required if cosine runs on the server (not default-`llm` bugs)

These were deferred (`prewarm`, `#660`). They are required if `embedding` / `hybrid` is a supported server strategy.

**3. Reuse the existing fastembed stack**

This PR builds a second `TextEmbedding` (default MiniLM). Knowledge and policy already load `BAAI/bge-small-en-v1.5` via `embedding_service` / `_FastEmbedEmbeddings`. Same disk cache, two ONNX sessions. `preload_models.py` only preloads bge-small, so airgapped servers cannot load MiniLM.

Route local shortlisting through `create_embedding_function` (batch + query/passage), default to `BAAI/bge-small-en-v1.5`. Do not inherit `[storage.embedding]` when that section is `openai`. SDK/query fallback if the model is not ready can stay.

**4. Server: embed at startup and on onboard, not on the first query**

Query-lazy is correct for the SDK. It is not enough for server mode when cosine is enabled.

Knowledge and policy already boot embeddings in `lifespan`. `PATCH /config/draft/tools` already reloads the registry and rebuilds the graph. This PR does neither for tool vectors, so the first `find_tools` after boot or after adding tools falls back to `llm`.

- Image preload: weights only (same model as knowledge).
- Server `lifespan`: load the shared fastembed session, embed the current catalogue.
- Onboard / registry reload: embed new and changed tools (content-hash). Do not reload the model.
- Query time: embed the question plus cache misses only.
- SDK: keep lazy.

Restart re-embeds tools (RAM cache dies) and must not re-download a warm model.

### Not blocking

Tool documents are the right text (split name + description + params + return field names; no `args_schema` JSON). Ranking is cosine. Query vs task context is a per-call vector blend, not a saved task index. `#660` can still own the UI strategy toggle; the embed/sync hook should not wait on that if cosine is on in server mode.

segevshlomovIBM:
Thanks @sami-marreed — all four addressed. Both correctness items reproduced first; you were right on both, and right about why CI missed them.

**1 · `embedding` unusable through either seam** — confirmed. `EmbeddingShortlister` returned a bare list while both seams read `result.candidates`. Every strategy now returns `ShortlistResult`, and hybrid reads `prefiltered.candidates` (the `TypeError` you predicted from a naive wrap). Added `test_shortlister_seam_unmocked.py`: it drives a **real** `EmbeddingShortlister` through `PromptUtils` and mocks only the embedding backend, so this class of bug can't hide behind a mocked `shortlist` again.

**2 · default `llm` capped above 128** — confirmed. `engage_cosine` keyed on catalogue size alone, so 129 tools on the shipped default still used the LLM but got `top_k=128` injected and the render truncated to 10. The gate now requires *a non-default ranker* **and** a catalogue worth cutting, at both seams; the bind cap keeps the caller's `top_k` untouched on the default path. Pinned at N = 10 / 129 / 400 — my old test used 20 tools, which is exactly why it passed.

**3 · reuse the existing fastembed stack** — done. Added `get_shared_text_embedding()` to `embedding_service`; `_LocalBackend` now uses it, so there is one ONNX session shared with knowledge and policy (verified at runtime: `sessions cached: 1`, `same object: True`). Default model moved to `BAAI/bge-small-en-v1.5`, which `preload_models.py` already fetches — so no preload change and airgapped servers work. bge is asymmetric and the existing query/passage path covers it. `[shortlister]` still does not inherit `[storage.embedding]`.

**4 · embed at startup and on onboard** — done. `warm_tool_vectors()` is the entry point; `warm()` on the embedding and hybrid strategies loads the shared session and embeds cache misses only. Wired into the server `lifespan` and into `PATCH /config/draft/tools` after the registry reload. Content-hashed, so adding two tools embeds two documents and does not reload the model. No-op on default `llm`, and neither hook can fail startup or a tools update. SDK stays lazy.

Green: **329 unit + 8 stability**, lint clean. AppWorld `test-easy` smoke after (2), as you asked — flag if you want that run before you re-review.

segevshlomovIBM:
## E2E across the config matrix

@sami-marreed — since both bugs you found were invisible to mocked tests, here is an unmocked run over the whole config matrix. Real `BAAI/bge-small-en-v1.5`, real LLM (watsonx `gpt-oss-120b`), driven through `PromptUtils.find_tools` / `shortlist_tool_names`. Nothing about the shortlister is stubbed. **24/24 pass** on `26ee8a7e`.

```
configured default model: BAAI/bge-small-en-v1.5

[A] DEFAULT (strategy=llm) — must be unchanged at every N
  PASS  3 tools: LLM path, no cosine
  PASS  3 tools: not truncated to max_results          — # Found 1 Matching Tool(s)
  PASS  203 tools (>threshold): LLM path, no cosine
  PASS  203 tools (>threshold): not truncated          — # Found 1 Matching Tool(s)
  PASS  bind cap: caller top_k honoured
[B] strategy=embedding — both seams, real cosine
  PASS  find_tools uses cosine
  PASS  top hit correct
  PASS  respects max_results=10                        — 10 rendered
  PASS  bind cap returns real names                    — ['crm_delete_lead_leads_lead_id_delete', …]
  PASS  needed tool present
[C] strategy=hybrid — cosine prefilters, LLM decides
  PASS  returns markdown
  PASS  LLM produced the ranking (not cosine text)
[D] threshold gate
  PASS  below threshold -> LLM
  PASS  below threshold -> zero vectors embedded
[E] per-seam config
  PASS  discovery seam resolves                        — hybrid
  PASS  bind_cap seam resolves                         — embedding
[F] fallback when backend cannot load
  PASS  degrades to LLM
  PASS  still usable
[G] server warm path
  PASS  no-op on default llm                           — embedded 0
  PASS  warms catalogue for cosine                     — embedded 203/203
  PASS  re-warm embeds only the delta                  — embedded 1
[H] shared ONNX session
  PASS  one session shared with knowledge/policy       — 1 cached

ALL E2E MATRIX CHECKS PASSED
```

The lines that map directly to your review:

- **[A] at 203 tools** is the case that was broken — default `llm` above the threshold was being handed `top_k=128` and truncated to 10. Now uncapped at every N.
- **[B]** is your bug 1: a real `EmbeddingShortlister` consumed by both seams. Before the fix this raised `AttributeError` once the model was warm.
- **[G] `re-warm embeds only the delta — embedded 1`** is the content-hash property: adding one tool embeds one document and does not reload the model.
- **[H] `1 cached`** confirms the shortlister shares the fastembed session with knowledge and policy rather than opening a second ONNX session.

Two honest notes on reading it: the LLM legs report `# Found 1 Matching Tool(s)` because the model judged one tool relevant for that query — that is its call, not a cap; the `max_results` behaviour is asserted in [B] where cosine drives the count. And [D] asserts *zero vectors embedded* rather than just "no cosine text", so it cannot pass by accident.

Still outstanding on my side: the AppWorld `test-easy` smoke you asked for after fix 2. It needs the benchmark environment, so say the word if you want it before re-reviewing.

sami-marreed (approved):
## Approve

All four request-changes items are addressed on `26ee8a7e`.

1. Strategies return `ShortlistResult`; hybrid reads `prefiltered.candidates`; unmocked seam tests go through `PromptUtils`.
2. Default `llm` is uncapped at every N (`not plan.is_llm_only`).
3. Local shortlisting reuses `get_shared_text_embedding()`, defaults to `BAAI/bge-small-en-v1.5`, and does not inherit `[storage.embedding]`.
4. Server `lifespan` and `PATCH /config/draft/tools` warm the catalogue; the SDK stays lazy.

AppWorld default path is unchanged (per-app ≤92, cosine never engages). Not blocking: knowledge still constructs its own `TextEmbedding`; `_VECTORS` is not pruned on tool removal; AppWorld `test-easy` smoke is still outstanding.

sami-marreed:
Approved.

Hold merge until the earlier approved PRs land first:

- https://github.com/cuga-project/cuga-agent/pull/644 — `fix(cuga-lite): trim duplicate find_tools schema blocks (#641)`
- https://github.com/cuga-project/cuga-agent/pull/654 — `ci: parallel PR jobs, directory test discovery, cheaper stability`

This one can follow them.

sami-marreed:
@segevshlomovIBM synced this with `main` (`#644` + `#654`). Conflicts resolved:

- `.github/workflows/tests.yml` — kept `#654` directory discovery; dropped the file allowlist
- `prompt_utils.py` / `render.py` — kept the pluggable shortlister and routed markdown through `#644`'s `_render_find_tools_markdown` so schema trim still applies

Where the new tests run now (no YAML edit needed):

| Tests | CI job |
|---|---|
| `src/cuga/backend/cuga_graph/nodes/cuga_lite/tests/test_shortlister_*.py` (`unit`) | **Unit (graph)** / `unit-a` |
| `tests/unit/test_shortlister_config_surfaces.py`, `tests/unit/test_find_tools_exception.py` | **Unit (tests/)** / `unit-b` |
| `test_shortlister_real_model.py` (`stability` + `slow`) | **not** in the stability workflow (that only collects `src/system_tests/e2e`). `unit-a` collects the whole `cuga_lite/tests/` dir and does not pass `-m "not stability"`, so this module will now run in Unit (graph). Flag if you want that opted out of the 8-minute unit job.

Local check after the merge: 156 shortlister + schema-trim unit tests passed.

PR #675, opened 2026-08-17
fix(deps): skip memray on Windows in the profiling extra

sami-marreed:
## Bug fix

Fixes none

### Summary
Windows CI (`uv sync --all-extras --dev`) started failing after the `profiling` extra landed on main. That extra pulled in `memray`, which has no Windows wheels and cannot be built from source without pkg-config, so the Windows job died compiling it.

Skip `memray` on Windows with a PEP 508 marker (`sys_platform != 'win32'`). Linux/macOS still install it via `cuga[profiling]`.

### Testing
- [x] `uv run pytest tests/unit/test_profiling_extra_windows.py` passes
- [x] `uv lock` records `sys_platform != 'win32'` on memray
- [ ] Windows CI `uv sync --all-extras --dev` succeeds on this PR

## Summary by CodeRabbit

* **Bug Fixes**
  * The optional profiling feature is no longer installed on Windows, while remaining available on Linux and macOS.

* **Tests**
  * Added coverage to verify the platform-specific profiling dependency configuration.


offerakrabi (approved):
one thing, do you need to add the test to `.github/workflows/tests.yml` ?
do we want to?

PR #679, opened 2026-08-18
fix(knowledge): only load profiles with a known name

### LINKED ISSUE #677: [Chore] Close all open CodeQL alerts
Part of #262.

## Issue type

[Chore]

## What you want and why

Code scanning has **14 open CodeQL alerts** on `main`, the oldest from 2025-09-12. Five are high severity.

Three are real leaks. The rest are either the same pattern repeated elsewhere, or cases where the tool objects to a construct that is not actually a risk here.

Closing them all gets the count to zero, so that a new alert stands out instead of being lost among old ones.

| # | Rule | Severity | Where |
|---|---|---|---|
| 2 | `py/stack-trace-exposure` | medium | `server/main.py:2416` |
| 53 | `py/stack-trace-exposure` | medium | `server/main.py:3110-3117` |
| 146 | `py/stack-trace-exposure` | medium | `server/main.py:3241` |
| 166 | `py/stack-trace-exposure` | medium | `manage_routes/config_routes.py:442` |
| 167, 168, 169 | `py/stack-trace-exposure` | medium | `server/a2a/router.py:228, 244, 267` |
| 29 | `py/clear-text-logging-sensitive-data` | high | `registry/tests/test_auth/auth_server.py:138` |
| 54, 55 | `py/clear-text-logging-sensitive-data` | high | `registry/authentication/appworld_auth_manager.py:292, 296` |
| 175 | `py/bind-socket-all-network-interfaces` | medium | `system_tests/profiling/startup/measure_server_startup.py:45` |
| 68, 69 | `py/path-injection` | high | `knowledge/config.py:419-422` |
| 170 | `py/weak-sensitive-data-hashing` | high | `llm/models.py:580` |

The three that matter most:

- **`main.py:3116`** puts a full stack trace into the body of an HTTP response, where the caller can read it.
- **`knowledge/config.py:419`** builds a file path out of a profile name supplied in published configuration, without checking it first. Two of the four places that call it do check the name; two do not.
- **`appworld_auth_manager.py`** prints the body and headers of a failed request to standard output. The request in question is a sign-in form, so the body contains a password. Line 242 also writes the password to the log directly, which CodeQL did not report.

## How it could work

Three pull requests, grouped by who owns the files rather than by which rule was reported.

The rules on `main` require a review from an owner of every area a change touches, dismiss all approvals whenever the branch is pushed to, and squash each pull request into a single commit. One large pull request would therefore wait on every owner at once, lose all its approvals on each update, and land as one commit spanning unrelated areas, which cannot be reverted piece by piece.

**PR A — stop internal details being sent out** (alerts 2, 29, 53, 54, 55, 146, 166, 167, 168, 169, 175)

A shared helper writes the error to the server log and returns a fixed message plus a short reference code, never anything taken from the exception. Applied to the seven reported handlers. Also: hide credentials in the failure reports from AppWorld sign-in, stop the auth test server printing its credentials at startup, and have the startup timing script listen only on this machine.

**PR B — only accept known profile names** (alerts 68, 69)

Check the name against the list of known profiles inside `load_profile`, and build the path with the existing helper that refuses to leave the profiles directory.

**PR C — alert #170** (alert 170)

Dismissed as a false positive rather than fixed. Details below.

### Why alert #170 cannot be fixed in code

The rule wants a password-hashing function that is deliberately slow, such as bcrypt or PBKDF2. Every ordinary hash fails it, so changing MD5 to SHA-256 does not help. A deliberately slow function is the wrong choice here: this code runs every time a model is requested, and those functions take upwards of 100 milliseconds by design. Passing `usedforsecurity=False` has no effect, because CodeQL does not read that argument in any language ([github/codeql#13637](https://github.com/github/codeql/issues/13637), open since 2023). Removing the credential from the data before hashing does not work either, because of how CodeQL follows values stored in a dictionary.

The value in question is only used as a key in an in-memory dictionary. It is never saved, never sent anywhere, and is held in the same memory as the credential it was made from, so anyone who can read it can already read the credential.

### A note about the sanitizer settings

The model pack in `.github/codeql/extensions/` can tell CodeQL that a function makes a value safe. This only works for rules written to read those entries. `py/path-injection` is one of them, which is why the existing pack works. `py/stack-trace-exposure`, `py/clear-text-logging-sensitive-data` and `py/weak-sensitive-data-hashing` are not, so an entry naming one of them would be loaded and then ignored, with no warning. That is the same kind of mistake recorded in #633, so PR A writes the explanation into the configuration file.

## Links or extra context

- Alerts: https://github.com/cuga-project/cuga-agent/security/code-scanning
- Earlier related work: #634 (making the sanitizer settings take effect) and #633 (settings that looked active but were not)
- Each branch is checked by building a CodeQL database locally and comparing it against `main` before the pull request is opened

## Checklist

haroldship:
## Bug fix

Part of #677. Closes CodeQL alerts **68 and 69** (`py/path-injection`, high).

This is not written as `Fixes #677`, because that issue covers all three pull requests.

### Summary

`load_profile` joined the profile name straight into a file path:

```python
path = _PROFILES_DIR / f"{profile_name}.toml"
```

Two of the places that call it pass a name that a user can set. `KnowledgeConfig.from_settings` reads `search.rag_profile` directly from published configuration, and `awareness` passes on the same value. A name chosen carefully could therefore be used to read a file outside the profiles directory.

### Root Cause / Solution

The restriction already existed as a rule; it simply was not applied where the path was built. `KnowledgeConfig.__post_init__` rejects a profile name that is not in the known list, and two of the four callers check the name before calling. The other two do not.

So the check moves into `load_profile` itself:

- Reject any name that is not in the known list, as the first thing the function does.
- Build the file name with `child_path_under`, the helper this repository already uses for this purpose. It refuses any name containing a directory separator or `..`, and it is one of the functions the CodeQL settings already recognise as making a path safe.
- Import that helper inside the function rather than at the top of the file. The module it lives in brings in about 2800 other modules and 1.6 seconds of loading time, which loading a profile has no other reason to pull in. This matches the existing deferred import in the same function.

**Nothing changes for a valid profile name.** An invalid one already failed, at `__post_init__`; it now fails earlier and says what was wrong. `list_profiles()` still returns all four profiles.

### Testing

New `tests/unit/test_knowledge_profile_paths.py`, 14 cases:

- Every known profile still loads and still has its `search` section.
- Names that try to escape the directory (`../../../etc/passwd`, `..`, `nested/standard`, a backslash version, an absolute path, and an empty name) are rejected, with a stand-in for `open` proving **no file is opened at all**, rather than merely that reading it failed.
- An ordinary unknown name raises `ValueError` rather than `FileNotFoundError`, which records the distinction this change introduces.
- A readable file outside the profiles directory still cannot be loaded.

Also run: 86 related unit tests pass, and `ruff check` and `ruff format --check` are clean.

Checked locally by building a CodeQL database from this branch and comparing it with `main`: both alerts are gone and nothing new appeared.

A local review raised one point, that the name was being checked after the deferred imports rather than before, fixed in 69facfe0. The review after that was clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved profile loading validation to reject unknown, absolute, and path-traversal profile names.
  * Ensured approved profiles remain restricted to the designated profile directory.
  * Preserved clear errors when requested profile files are missing.

* **Tests**
  * Added coverage for valid profiles and invalid path or profile-name inputs.


offerakrabi (approved):
## PR Review: #679 — fix(knowledge): only load profiles with a known name

One minor gap in the test suite; the security fix itself is correct and complete.
Decide whether to fix or deffer.


|---|---|---|---|---|
| 1 | Minor | `child_path_under` containment layer has no test coverage | `tests/unit/test_knowledge_profile_paths.py` | All rejection tests hit the allowlist check first; the containment helper is never exercised → removing it would not break any test. |

Issue #685, opened 2026-08-20
[Bug]: cancel upload leaves ingest running so re-upload 409s as already indexed


sami-marreed:
Filed without a parent: small self-contained bug (cancel + duplicate-ingest guard). Related to #683 (stale `running` tasks after restart).

## What happened

Cancelling a knowledge upload does not actually stop the ingest. The UI marks the row cancelled immediately. Re-uploading the same file then fails with 409:

```
file already indexed: medical-coverage.md (an ingest for this file is already running; task_id=task_f0f171881f6a)
```

Expected: Cancel ends the ingest (task `cancelled`), and a new upload of the same filename is allowed. The file should not be treated as indexed while a cancelled/stuck ingest is still `running`.

## How to reproduce

1. Open agent knowledge and start uploading a file (e.g. `medical-coverage.md`).
2. Cancel the upload while it is still in progress.
3. UI shows cancelled.
4. Upload the same file again.

Observed: 409 `file already indexed: … (an ingest for this file is already running; task_id=…)`.

The same lock remains across pod restart: `cancel_task` for a `running` task only sets an in-memory `cancel_event`. That is lost on restart, metadata stays `running`, and `_sanitize_and_validate` still rejects the filename.

## Environment

- Hosted CUGA / Sovereign Core, `cuga:0.3.1`
- `storage.mode=prod` (Postgres knowledge metadata)
- Agent knowledge upload UI
- Observed 2026-08-20 on tenant `36f6decd-3ef4-48cb-bbe7-6c283b4ac1c4`

## Logs, screenshots, or config

Client/server message:

```
file already indexed: medical-coverage.md (an ingest for this file is already running; task_id=task_f0f171881f6a)
```

Code:

- UI cancels locally first and treats server cancel as best-effort (`KnowledgeConfig.tsx` `handleCancelUpload`). Comment already notes `cancel_event` is only checked between stages and that a later re-upload will 409.
- `KnowledgeEngine.cancel_task` only writes `status=cancelled` when the parent task is still `pending`. For `running` it sets `_active_tasks[task_id]` and returns without updating metadata.
- Duplicate guard in `_sanitize_and_validate` / `_create_task_entry_internal` 409s any `pending`/`running` task that still lists that filename.

Workaround on a live deploy: mark stuck tasks failed in `cuga_knowledge_meta_tasks` (same SQL as #683), then retry the upload.

## Checklist

- [x] I searched existing issues and this doesn’t look like a duplicate

[...]

PR #687, opened 2026-08-20
fix(knowledge): tolerate status-less file-tasks in restart recovery

### LINKED ISSUE #683: [Bug]: knowledge warmup KeyError on Postgres stale tasks missing file-task status
Filed without a parent: small self-contained bug (SQLite recovery already handles this; Postgres was missed).

## What happened

On hosted CUGA (`storage.mode=prod`), knowledge warmup crashes with `KeyError: 'status'` while recovering interrupted ingest tasks.

`PostgresKnowledgeMetadata.recover_stale_tasks()` assumes every nested `file_tasks_json` entry has a `status` key. Progress/partial ingest writes often omit it (filename + stage + progress only). SQLite already defaults missing `status` to `pending` and continues; Postgres does `ft["status"]` and raises.

Warmup is caught, so the demo process still starts, but the knowledge subsystem is marked failed and OOBE PDF seed is skipped.

Expected: stale running/pending tasks are marked failed on startup, including file-task entries with no `status`. Knowledge warmup succeeds.

## How to reproduce

1. Run CUGA in prod/Postgres knowledge mode.
2. Leave (or insert) a row in `cuga_knowledge_meta_tasks` with `status` in (`running`, `pending`) and `file_tasks_json` like:

```json
{"summary-plan-description.md": {"filename": "summary-plan-description.md", "stage": "parsed", "progress": {"done": 182, "total": 182}}}
```

3. Restart the demo/server so `KnowledgeEngine.warmup()` → `_ensure_metadata_ready()` → `recover_stale_tasks()` runs.

Without a data workaround, warmup fails every restart until those parent rows are closed.

## Environment

- Hosted CUGA, Linux container (`sandbox_mode` fell back from native to local)
- `storage.mode=prod` / Postgres knowledge metadata
- Observed 2026-08-20 on a running demo (`cuga start` knowledge/manage mode)

```
ERROR | cuga.backend.server.main:_warm:734 - Knowledge subsystem warmup failed: %s
  File ".../knowledge/metadata/postgres_store.py", line 207, in recover_stale_tasks
    if ft["status"] in ("pending", "processing"):
       └ {'filename': 'summary-plan-description.md', 'stage': 'parsed', 'progress': {'done': 182, 'total': 182}}
KeyError: 'status'

WARNING | ...demo_manage_setup:seed_demo_knowledge_oobe_pdf_via_engine_if_needed:800 - Knowledge subsystem failed; skip OOBE PDF seed
```

Workaround without a new image: mark the parent tasks failed (or delete running/pending rows) in `cuga_knowledge_meta_tasks`, then restart the same pod.

## Checklist

dolev31:
## Bug fix

Fixes #683

### Summary

Knowledge warmup on `storage.mode=prod` crashed with `KeyError: 'status'` while recovering interrupted ingest tasks, taking the whole knowledge subsystem down.

### Root cause

`PostgresKnowledgeMetadata.recover_stale_tasks()` did `ft["status"]` on every nested file-task entry. Those entries legitimately have no `status`:

- the ingest worker's progress emits write `{filename, stage, progress}` and **deliberately** omit `status`, so a late emit cannot un-flip `status="completed"` back to `"processing"` (`_emit_progress`, engine.py);
- `update_task` **replaces** the whole `file_tasks_json` blob rather than merging it.

So any task killed mid-ingest leaves a status-less entry behind. This is normal steady state, not corrupt or legacy data.

SQLite has defaulted the missing key since `92618226`; that guard was never applied to Postgres, so the two backends diverged silently for ~7 weeks.

**Blast radius is wider than the report suggests.** `_ensure_metadata_ready` has no `try/except` and never latches `_metadata_ready`, so the `KeyError` re-raises on *every* subsequent call — not just the skipped OOBE PDF seed, but 500s from `list_documents`, `get_tasks`, `ingest`, `delete_document` and `reindex` until the DB is hand-edited.

### Solution

Because the status-less write is intentional and race-preventing, the correct fix is a tolerant reader, not a changed writer. To stop the two backends drifting again, the rules now live in one place:

- `normalize_file_tasks()` and `mark_file_tasks_interrupted()` added to `metadata/base.py` — the only domain module both stores already import (there is deliberately no shared base class: `SqliteKnowledgeMetadata(LocalRelationalStore)` vs `PostgresKnowledgeMetadata(ProdRelationalStore)`).
- Both `recover_stale_tasks` implementations reduce to the same single line and route through it.
- Postgres now also preserves an existing `ft["error"]` instead of overwriting it with the generic restart message.
- A comment at `_emit_progress` points at the helper, so the next reader of a status-less entry knows it is by design.

**A second defect, present in both stores, is fixed by the same change:** a non-object payload (corrupt JSON, `null`, a list) skipped the marking loop via an `isinstance` guard but was still re-serialized back into the column — leaving a value that every other reader treats as a mapping. Normalizing on read means the `UPDATE` can only ever persist an object.

### Type of changes

### Testing

New `tests/unit/test_knowledge_recover_stale_tasks.py` (9 tests, `@pytest.mark.unit`, lands in the `unit-b` shard). Postgres is exercised through a fake that replaces only the asyncpg boundary, so the store's real recovery logic runs with no live Postgres; SQLite is tested end-to-end against a real temp DB.

Verified to be genuine regression tests, not just green: against unfixed `main` three of them fail, including the exact `KeyError: 'status'` at `postgres_store.py:207` from the issue.

- `uv run pytest tests/unit/test_knowledge_recover_stale_tasks.py` — 9 passed
- `uv run pytest tests/unit/ -k knowledge` — 576 passed, no regressions

Notable coverage: the `#683` payload verbatim; corrupt/`null`/list/scalar payloads; a pre-existing specific error surviving recovery; and a **parity test** asserting both backends produce identical output for the same input, which fails if either stops routing through the shared helper.

Also validated end to end against a real local Postgres deployment (`deployment/deploy-local-postgres.sh`, `storage.mode=prod`): planting the exact `#683` payload in `cuga_knowledge_meta_tasks` and restarting the server now recovers cleanly instead of failing warmup.

### Notes for the reviewer

- Deliberately **not** changing the progress emits or making `update_task` merge. The omission is load-bearing and the existing drain before the completion write already handles the race it guards against. Changing it is a bigger blast radius than an urgent prod fix should carry.
- `cancel_task` has the same unguarded `ft["status"]` subscript. Left alone here on purpose: it is unreachable today (a `pending` task always has `status` from `_create_task_entry_internal`), and the #685 fix rewrites that whole function. Flagging so it does not read as an oversight.

### DIFF
--- a/src/cuga/backend/knowledge/engine.py
+++ b/src/cuga/backend/knowledge/engine.py
@@ -2501,6 +2501,11 @@ class KnowledgeEngine:
             progress_futures: list[_cf.Future] = []
 
             async def _emit_progress(stage: str, done: int, total: int) -> None:
+                # Like the "parsed" emit above, this omits ``status`` on
+                # purpose, and ``update_task`` replaces the whole entry — so a
+                # task killed here leaves a status-less file-task behind.
+                # Readers must tolerate that; see ``normalize_file_tasks`` in
+                # knowledge/metadata/base.py (#683).
                 await self._metadata.update_task(
                     task_id,
                     file_tasks={
--- a/src/cuga/backend/knowledge/metadata/base.py
+++ b/src/cuga/backend/knowledge/metadata/base.py
@@ -2,8 +2,17 @@
 
 from __future__ import annotations
 
+import json
 from datetime import datetime, timedelta, timezone
 from typing import Any, Protocol
+
+#: Error recorded on file-task entries that a restart interrupted mid-ingest.
+INTERRUPTED_ERROR = "interrupted by server restart"
+
+#: File-task statuses that a restart should re-mark as ``failed``. A missing
+#: ``status`` is treated as ``pending`` and therefore recovered too — see
+#: ``normalize_file_tasks`` for why the key can be absent.
+_RECOVERABLE_FILE_STATUSES = ("pending", "processing")
 
 
 def utc_now_iso() -> str:
@@ -12,6 +21,55 @@ def utc_now_iso() -> str:
 
 def iso_cutoff_days_ago(days: int) -> str:
     return (datetime.now(timezone.utc) - timedelta(days=days)).isoformat()
+
+
+def normalize_file_tasks(raw: Any) -> dict[str, Any]:
+    """Coerce a stored ``file_tasks_json`` value into a mapping.
+
+    Accepts the raw JSON text from the DB or an already-deserialized value,
+    and always returns a dict so callers can iterate and re-serialize without
+    guarding every access.
+
+    Two shapes on disk make this necessary:
+
+    1. **A file-task entry may legitimately have no ``status`` key.** The
+       ingest worker's progress emits replace the whole entry with
+       ``{filename, stage, progress}`` on purpose, so a late emit cannot
+       un-flip ``status="completed"`` back to ``"processing"``
+       (see ``_emit_progress`` in ``knowledge/engine.py``). Any task killed
+       mid-ingest therefore leaves a status-less entry behind. Readers must
+       treat a missing status as *non-terminal*, never assume the key exists
+       — assuming it crashed Postgres warmup with ``KeyError: 'status'``
+       (#683) while SQLite tolerated it, and the two backends silently
+       diverged for weeks.
+    2. **The payload may not be an object at all** — corrupt or truncated
+       JSON, ``null``, or a list written by an older build. Returning ``{}``
+       keeps the caller's re-serialization from persisting a non-object back
+       into a column the rest of the code reads as a mapping.
+    """
+    if isinstance(raw, (str, bytes, bytearray)):
+        try:
+            raw = json.loads(raw)
+        except (TypeError, ValueError):
+            return {}
+    return raw if isinstance(raw, dict) else {}
+
+
+def mark_file_tasks_interrupted(file_tasks: dict[str, Any]) -> dict[str, Any]:
+    """Re-mark non-terminal file-task entries as failed, in place.
+
+    Shared by every backend's ``recover_stale_tasks`` so the tolerance rules
+    live in one place. Entries that are not dicts are skipped rather than
+    dropped, and an existing ``error`` is preserved because it is more
+    specific than the generic restart message.
+    """
+    for ft in file_tasks.values():
+        if not isinstance(ft, dict):
+            continue
+        if ft.get("status", "pending") in _RECOVERABLE_FILE_STATUSES:
+            ft["status"] = "failed"
+            ft["error"] = ft.get("error") or INTERRUPTED_ERROR
+    return file_tasks
 
 
 class KnowledgeMetadataStore(Protocol):
--- a/src/cuga/backend/knowledge/metadata/postgres_store.py
+++ b/src/cuga/backend/knowledge/metadata/postgres_store.py
@@ -8,7 +8,12 @@ from typing import Any
 
 import psycopg
 
-from cuga.backend.knowledge.metadata.base import iso_cutoff_days_ago, utc_now_iso
+from cuga.backend.knowledge.metadata.base import (
+    iso_cutoff_days_ago,
+    mark_file_tasks_interrupted,
+    normalize_file_tasks,
+    utc_now_iso,
+)
 from cuga.backend.storage.relational.prod import ProdRelationalStore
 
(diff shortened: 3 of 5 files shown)


dolev31:
@sami-marreed this supersedes #684 — your diagnosis was right and the approach is kept, so please review the delta rather than the whole thing.

Two additions on top of your fix:

1. **The CodeRabbit finding on #684 was real, and it applied to SQLite too.** `isinstance(file_tasks, dict)` skipped the marking loop for a non-object payload, but the `UPDATE` still re-serialized the original value — so a stale row could persist `[1,2]` into a column every other reader treats as a mapping. Normalizing on read closes it in both stores.
2. **Moved the tolerance rules into `metadata/base.py`** so this class of bug cannot recur. #683 happened because the guard landed in SQLite (`92618226`) and was never back-ported; two hand-synced copies had already drifted once. Both `recover_stale_tasks` bodies are now one shared line, and there is a parity test that fails if either backend stops routing through the helper.

Worth a look first: the `normalize_file_tasks` docstring — it is where I wrote down *why* a file-task can have no `status`, which is the part that was never documented anywhere.

I also verified the tests actually fail without the fix rather than just passing with it: on `main` they reproduce the exact `KeyError: 'status'` at `postgres_store.py:207` from your report, and against your `b7eccb4e` the non-object test fails with `assert [1, 2] == {}`.

One thing I left alone deliberately: `cancel_task` has the same unguarded `ft["status"]` subscript, but it is unreachable today and #685's fix rewrites that function — so it belongs there, not here.

dolev31:
@sami-marreed could you take a look? Small, self-contained, and prod is degraded until it lands.

The part worth reviewing first is the `normalize_file_tasks` docstring — it is where I wrote down *why* a file-task entry can legitimately have no `status`, which was never documented anywhere and is the reason this bug was possible at all. Progress emits omit the key on purpose; SQLite tolerated that, Postgres did not.

Two things beyond the obvious KeyError fix:

1. **A second defect in both stores.** A non-object `file_tasks_json` payload skipped the marking loop but was still re-serialized back into the column, leaving a value every other reader treats as a mapping. Normalizing on read closes it.
2. **The rules now live in one place.** This bug existed because the guard landed in SQLite and never reached Postgres — two hand-synced copies that had already drifted once. Both `recover_stale_tasks` bodies are now a single shared line, plus a parity test that fails if either backend stops routing through the helper.

I checked the tests actually fail without the fix rather than just passing with it: on `main` they reproduce the exact `KeyError: 'status'` at `postgres_store.py:207` from your report. Also validated end to end against a real local Postgres deployment — planting your payload and restarting now recovers instead of failing warmup.

One deliberate omission: `cancel_task` has the same unguarded `ft["status"]`, but it is unreachable today and #685's fix rewrites that function, so it belongs there.

dolev31:
Validated end-to-end against a real local Postgres 17 + pgvector deployment (`storage.mode=prod`, `cuga start demo_knowledge`).

**How the poisoned row actually appears.** Started an ingest of a 156 KB .md, then `SIGKILL`ed the server mid-flight. The stranded row was:

```
task_1a54fd30d8a9 | running | {"medical-coverage.md": {"filename": "medical-coverage.md", "stage": "parsed", "progress": {"done": 96, "total": 96}}}
```

No `status` key — the exact shape from the report, produced by an ordinary progress emit. This is normal steady state after any mid-ingest kill, not legacy data.

**A/B on the identical database row:**

| | `main` (8b45234a) | this branch |
|---|---|---|
| `/api/knowledge/health` | `status=failed`, `error='status'` | `status=ready` |
| warmup log | `KeyError: 'status'` | `Recovered 1 stale task(s) from previous crash` → `warmup complete` |
| `GET /documents` | **500** | 200 |
| `GET /tasks` | **500** | 200 |
| the row afterwards | stays `running` forever | → `failed` |
| same filename re-uploadable | blocked | yes, completes |

That confirms the blast-radius point in the description: it is not only the skipped OOBE seed — the read surface 500s until the row is hand-edited, because `_ensure_metadata_ready` never latches and re-raises on every call.

Full unit suite on the merged result: 1410 passed. (Three `test_sdk_citations` failures show up in a full-suite run but reproduce identically on clean `main` and pass in isolation — pre-existing ordering issue, unrelated.)

sami-marreed (approved):

PR #688, opened 2026-08-20
fix(knowledge): make cancel end the ingest instead of wedging the filename

### LINKED ISSUE #685: [Bug]: cancel upload leaves ingest running so re-upload 409s as already indexed

dolev31:
## Bug fix

Fixes #685

### Summary

Cancelling a knowledge upload did not stop the ingest, and left the filename permanently wedged: re-uploading the same file failed with `file already indexed: medical-coverage.md (an ingest for this file is already running; task_id=…)`.

### Root cause — two defects, not one

1. **`cancel_task` recorded intent, never outcome.** It persisted a terminal status only when the parent task was still `pending`. For a `running` task it set an in-memory `asyncio.Event` and returned, so the row stayed `running` forever.
2. **The worker checked that event exactly once**, immediately before parse. Nothing after it checked. By contrast `_check_supersede`, sitting in the same function, is checked at four points including inside the collection lock — the cancel path was simply never given the same treatment.

So intent and guard never met: `cancel_task` wrote to memory, the dedup guard reads the persisted parent status.

Blast radius is wider than the report: the wedged `running` row also consumed a `max_pending_tasks` slot (default 10 — ten of them 429 the whole collection) and blocked `reindex` via `ReindexBusyError`. All three read the same `("pending","running")` filter.

**One correction to the issue text.** It says the lock survives pod restart. Restart recovery exists and *would* clear it — `recover_stale_tasks` flips `running`→`failed` at startup. It only survived restart because #683 crashed that recovery on Postgres. That half is fixed by #687, not here.

### Solution

Mirrors the supersede mechanism this file already uses, so there is one pattern rather than two.

- **`IngestCancelledError` + `_check_cancelled()`**, called at five points: before the `status="running"` write, after parse, after normalize, and twice inside the collection lock. A terminal handler writes parent `status="cancelled"` with the nuance in `file_tasks[filename]`, exactly like `ReindexSupersededError`. **No schema migration** — the parent status stays inside the existing `CHECK` constraint.
- The first checkpoint is deliberately **before** the `status="running"` write. A cancel landing while the worker is queued on `_ingest_sem` would otherwise be clobbered by that write and re-block the guard.
- **`cancel_task` persists a terminal status for running tasks too**, read-modify-writing `file_tasks` because `update_task` replaces the whole blob. A missing file-task status defaults **non-terminal** (the #683 lesson).
- **`_blocks_reupload()`** replaces the guard that was copy-pasted across `_sanitize_and_validate` and `_create_task_entry_internal`, and additionally skips file-tasks whose own status is already terminal.
- **`routes.py`**: a `DocumentExistsError` from the *atomic* layer escaped as a **500 with a leaked temp file**. Mapped to the same 409 with an `unlink`, plus a catch-all `unlink` so no future exception leaks it. This PR makes that path more reachable, since a cancel can now land between the advisory and atomic checks.

### The trap this deliberately avoids

Simply persisting `cancelled` releases the dedup guard immediately. If the worker is mid-embed, a re-upload is then admitted — and `_insert_documents_async` does `delete_by_source(source_id)` → `add_documents(docs)` under `replace_duplicates`. The two ingests serialize on the collection lock but in **unconstrained order**, so the cancelled worker can land *after* the replacement and overwrite it. "Cancel, then re-upload a corrected file with the same name" is the user's actual flow, so this is not exotic.

`_uncancellable_tasks` prevents it: once the worker clears its last checkpoint, cancel is **refused** rather than writing a terminal status. The check and the flag-set are adjacent and synchronous on purpose — with no `await` between them the event loop cannot interleave `cancel_task`, which makes the handoff atomic without a lock (the flock guarantees a single process). There is a comment saying so, because inserting an `await` there would silently reopen the race.

The invariant: **a task row is `cancelled` iff the worker is guaranteed never to touch the vector store, the documents table, or the files dir.**

### Frontend

`handleCancelUpload` awaited nothing: it aborted the poll, dropped the localStorage resume entry, painted "cancelled" locally, then fire-and-forgot the server call — and its comment explicitly rationalised the 409 as acceptable. It now awaits the server and only tears down the poll + resume entry once a terminal status is confirmed. A refused cancel keeps polling and renders "Finishing…" rather than claiming a cancel that did not happen.

`src/cuga/frontend/dist/**` is **generated** (`frontend/build.sh`) and committed in its own commit — no hand edits there.

### Type of changes

### Testing

New `tests/unit/test_knowledge_cancel_task.py` (10 tests) and 3 additions to `test_knowledge_routes.py`. Deterministic rendezvous, no sleeps: paired `threading.Event`s around a fake `_load_document` for the parse window, a stubbed `_insert_documents_async` + `asyncio.Event` for the insert window.

Verified as genuine regression tests, not just green:

| Code under test | Result |
|---|---|
| `main` (unfixed) | 8/10 cancel tests fail — incl. the exact `DocumentExistsError: File already indexed: f.txt (an ingest for this file is already running; task_id=t1)` from the issue, and `'running' == 'cancelled'` |
| `main` routes.py | both new route tests fail with the uncaught `DocumentExistsError` (the 500 path) |
| this branch | all pass |

- `uv run pytest tests/unit/ -k knowledge` — 580 passed
- `uv run pytest tests/unit/` — 1389 passed. 3 pre-existing failures in `test_sdk_citations.py` (live IBM watsonx `InvalidCredentialsError`); confirmed identical on clean `main` with this branch stashed.
- `tsc --noEmit` in `agentic_chat` — 0 errors in `KnowledgeConfig.tsx`
- `ruff check` / `ruff format --check` — clean

Coverage worth noting: the re-upload is asserted **while the doomed worker is still parsing** (the user should not have to wait for a cancelled ingest); one test stubs the insert to `raise AssertionError` to prove a cancelled worker never reaches the vector store; and two tests pin that cancel is *refused* mid-insert and that the duplicate window stays shut.

### Known limits, stated rather than hidden

- Cancel during a Docling parse does not *stop* the parse — `asyncio.to_thread` is not interruptible, so it burns CPU until it returns, then aborts. Harmless: the doomed worker writes nothing. Racing it with `asyncio.wait(FIRST_COMPLETED)` was considered and rejected — it would release `_ingest_sem` while the thread still runs, so `max_ingest_workers` would stop bounding real CPU/GPU work.
- One residual interleaving: if the worker's `completed` write lands between `cancel_task`'s read and its `UPDATE`, the row says `cancelled` while the document is indexed. Stale *label* only — the vector store and `documents` agree, and the next `loadDocuments()` shows the truth. Closing it needs a task-write lock; deferred deliberately.
- Cancel cannot reach a worker in another process. The flock makes that impossible today; there is a comment so the coupling is not rediscovered the hard way if it is ever relaxed.
- No frontend unit test: `handleCancelUpload` is a closure inside a ~2000-line component with no seam, and extracting one would be a larger, riskier diff than the fix itself.

### Follow-ups (not in this PR)

- `_active_tasks` is registered in `_run_ingest` but popped in `_ingest_inner`, so it leaks if `_run_ingest` raises in between — which wedges `wait=true` uploads permanently, with no cancel involved.
- No wall-clock cap on the upload path (the reindex path has `_REINDEX_WORKER_TIMEOUT_S`); a hung embed still leaves a `running` row. The new `cancel_task` at least makes it user-fixable.

## Summary by CodeRabbit

* **New Features**
  * Improved document-upload cancellation with durable status updates.
  * Uploads now show a “Finishing…” state during final processing.
  * Duplicate document uploads return a clear conflict response.

* **Bug Fixes**
  * Preserved backend error messages for failed uploads.
  * Prevented cancelled or in-progress uploads from corrupting duplicate detection.
  * Ensured cancellation does not interrupt uploads after data insertion begins.
  * Cleaned up temporary files when task creation fails.

* **Tests**
  * Added coverage for cancellation, duplicate uploads, status transitions, and in-progress processing safeguards.


dolev31:
Validated against a real local Postgres 17 + pgvector deployment (`storage.mode=prod`, `cuga start demo_knowledge`), driving the same HTTP surface the UI uses. 29/29 checks pass; here is what was actually exercised, using a 156 KB .md so parse and embed take real time.

| scenario | result |
|---|---|
| cancel mid-ingest, re-upload same filename **immediately** | cancel persists `cancelled`; re-upload **200, not 409** — the reported bug |
| batch upload 5 files | all 5 complete, all 5 listed, no duplicates |
| batch of 5 then cancel **all** mid-flight | 4 `cancelled`, 1 `completed` (that one was past the point of no return — by design); all 5 filenames immediately re-uploadable |
| 4 concurrent uploads of the same filename | exactly one 200, three 409, **no 500s**, exactly one copy indexed |
| delete during an in-flight ingest | 404 (not yet indexed), task still terminalizes, re-upload works |
| publish (`POST /api/manage/config`) mid-ingest | 200, version bumped, ingest still completes |
| cancel unknown id / double cancel / cancel after terminal | 404 / 200,200 / 200 — no 500s |
| final invariant | zero rows left `pending`/`running` |

The batch-cancel row is the interesting one: one task legitimately refused cancellation because it had already crossed into the vector insert, and it still completed while the other four cancelled — and crucially the filename was *still* re-uploadable straight away. That is the point-of-no-return design doing exactly what it is meant to, rather than the naive "just mark it cancelled" which would have let the re-upload race that live insert.

Restart behaviour, same deployment: `SIGKILL` mid-ingest → restart → `Recovered 1 stale task(s)` → warmup complete → zero stranded rows → same filename re-uploads and completes. (The recovery half of that depends on #687; the two together are what make the reported "survives pod restart" symptom go away.)

Full unit suite on the merged result: 1410 passed.

dolev31:
Both review findings were valid and are fixed in `ddd47e68`.

**The Major one was a real hole in my fix, and my own tests missed it.** The cancel event is in-memory, and `cancel_task` can run *before* `_run_ingest` ever registers it — routes.py creates the task row and only then schedules the worker coroutine. In that window `_active_tasks` is empty, so cancel persists `cancelled` with no event to set, and the worker then builds its own fresh unset event. The in-memory-only checkpoint sailed straight past it, wrote `running`, and went on to parse, embed and insert — resurrecting a row the dedup guard had already released, which is exactly the duplicate-content race the point of no return exists to prevent.

My existing test set the event by hand, so it never exercised that window. I wrote the regression first and confirmed it fails on the previous commit (the worker reached the vector store), then fixed it: re-read the persisted status before the first write and return if it is already terminal. Return rather than raise, so a row that is genuinely `completed`/`failed` keeps its own outcome instead of being overwritten by the cancel handler.

The Minor one was also right — a task that was **already** `failed` when cancel was requested got painted "Cancelled" with its error cleared, hiding a real ingest failure behind a user action that did not cause it. Now `failed` renders as an error and keeps the file-task message.

CI green. Full unit suite across all three knowledge PRs merged together: 1415 passed.

dolev31:
Closed the last known interleaving in `577f4eee`, so this no longer ships with a documented caveat.

**What was still open:** `update_task` is last-write-wins with no CAS, so the worker's `status="completed"` could land between `cancel_task`'s read (which saw `running`) and its UPDATE. The row then read `cancelled` while the document was really indexed — a stale label, but one the UI shows the user.

**Fix:** a `_task_write_lock` held across read-decide-write in `cancel_task`, and around each of the worker's four terminal writes (completed / cancelled / superseded / failed). Whichever side goes second re-reads, sees the terminal status the other just wrote, and declines to overwrite it.

One ordering detail worth checking in review: the cancel **event** is still set *before* the lock is taken. The worker's checkpoints are synchronous reads of that event, so a worker that runs while cancel is waiting on the lock still observes the cancel and aborts. Setting it after the lock would reintroduce a window.

The test asserts the *invariant* rather than a fixed winner — either side may legitimately win the race; what must never happen is the row and the document disagreeing. My first attempt at that test held the lock by hand and deadlocked against the worker's terminal write, which is itself a decent demonstration that the serialization is real.

582 knowledge tests pass.

PR #690, opened 2026-08-20
fix(knowledge): return 400 for a missing or malformed JSON body

### LINKED ISSUE #689: [Bug]: knowledge endpoints return 500 on missing or malformed JSON body
Filed without a parent: small self-contained bug (input validation, no design change).

## What happened

Four knowledge endpoints call `await request.json()` without guarding it. A request with **no body** or a **malformed body** raises `json.decoder.JSONDecodeError`, which escapes as an unhandled ASGI exception → **HTTP 500** and a reset connection, instead of a 400.

Found while abuse-testing the knowledge endpoints against a local Postgres (`storage.mode=prod`) deployment.

Measured against `main` (`8b45234a`):

| endpoint | no body | malformed body |
|---|---|---|
| `POST /api/knowledge/search` | **500** | **500** |
| `POST /api/knowledge/documents/url` | **500** | **500** |
| `DELETE /api/knowledge/documents` | **500** | **500** |
| `POST /api/knowledge/reindex` | 200 | **500** |
| `PATCH /api/knowledge/session/settings` | 400 | 400 |

`POST /reindex` already guards the *empty* case (`... if request.headers.get("content-length","0") != "0" else {}`) but still parses an existing body unguarded, so a malformed body still 500s. `PATCH /session/settings` is the one that behaves correctly today.

## Why it matters

- A 500 is the wrong contract for bad client input; clients cannot distinguish "my request was malformed" from "the server is broken".
- It logs a full traceback per occurrence, so trivially-malformed traffic pollutes error monitoring and can mask real failures.
- `DELETE /documents` is reachable from the UI, so a client that ever omits the body gets a reset connection rather than a clear error.

## How to reproduce

1. Run CUGA with knowledge enabled (any storage mode).
2. Send a request with no body:

```bash
curl -i -X DELETE -H "X-Agent-ID: cuga-default" \
  http://127.0.0.1:7860/api/knowledge/documents
```

3. Observe HTTP 500 and `json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)` in the server log, raised from `routes.py` `delete_document`.

Same for `POST /search`, `POST /documents/url`, and `POST /reindex` with `--data-binary 'not json'`.

## Expected

`400 Bad Request` with a clear detail message (e.g. `"invalid JSON body"`), no traceback.

## Environment

- `main` @ `8b45234a`
- Local Postgres (`storage.mode=prod`), knowledge enabled, `cuga start demo_knowledge`
- Observed 2026-08-20

## Checklist

dolev31:
## Bug fix

Fixes #689

### Summary

Four knowledge endpoints returned **500** on a missing or malformed JSON body instead of 400. Found while abuse-testing the knowledge API against a local Postgres (`storage.mode=prod`) deployment.

### Root cause

`await request.json()` raises `json.decoder.JSONDecodeError` when the body is absent or not JSON. Unhandled, that escapes as an ASGI 500 with a full traceback.

Measured against `main` (`8b45234a`):

| endpoint | no body | malformed body |
|---|---|---|
| `POST /api/knowledge/search` | **500** | **500** |
| `POST /api/knowledge/documents/url` | **500** | **500** |
| `DELETE /api/knowledge/documents` | **500** | **500** |
| `POST /api/knowledge/reindex` | 200 | **500** |
| `PATCH /api/knowledge/session/settings` | 400 | 400 |

`POST /reindex` guarded only the *empty* case via a `content-length` check, so a malformed body still reached the unguarded parse. `PATCH /session/settings` was the only endpoint already doing it right.

### Why it matters

- Wrong contract: callers cannot distinguish "my request was malformed" from "the server is broken", and 500s are what alerting escalates on.
- Every occurrence logs a traceback, so trivially-malformed traffic buries real failures in error monitoring.
- `DELETE /documents` is reachable from the UI, so a client that omits the body gets a reset connection rather than a usable error.

### Solution

Generalize the guard `patch_session_settings` already had into `_json_body()` and route every body-reading endpoint through it:

- absent/whitespace body → 400 `"request body must be a JSON object"` (or `{}` when the endpoint opts in via `allow_empty`)
- unparseable body → 400 `"invalid JSON body"`
- valid JSON that is not an object (a bare list or scalar) → 400, since every caller immediately does `body.get(...)`

`allow_empty=True` on `/reindex` preserves the documented no-body call **and** replaces the `content-length` special case, closing the malformed-body hole in the same change. After this, `request.json()` is called in exactly one place.

### Type of changes

### Testing

New `tests/unit/test_knowledge_json_body.py` (12 tests, `@pytest.mark.unit`), parameterized across every body-reading endpoint: missing body, malformed body, non-object JSON, whitespace-only body, plus two regression guards for `/reindex` (empty body still accepted; malformed body now rejected).

Verified to be genuine regression tests: **11 of 12 fail against `main`**, all with the 500 this fixes. The one that passes on `main` is the empty-body-reindex guard, which is exactly the behaviour that had to be preserved.

- `uv run pytest tests/unit/test_knowledge_json_body.py` — 12 passed
- `uv run pytest tests/unit/ -k knowledge` — 579 passed, no regressions
- Re-probed live against a running server on Postgres prod mode: all four endpoints now answer 400.

### Note

Independent of #687 and #688 (also knowledge, also found in the same session) — no overlapping lines, no merge-order dependency.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for JSON request bodies across knowledge-related actions.
  * Invalid, malformed, empty, or unsupported request data now returns a clear HTTP 400 error instead of an unexpected server error.
  * Reindexing continues to support requests with no body.

### DIFF
--- a/src/cuga/backend/knowledge/routes.py
+++ b/src/cuga/backend/knowledge/routes.py
@@ -59,6 +59,31 @@ def _ensure_enabled(engine: KnowledgeEngine) -> None:
         raise HTTPException(status_code=503, detail="Knowledge engine is disabled")
 
 
+async def _json_body(request: Request, *, allow_empty: bool = False) -> dict[str, Any]:
+    """Parse a JSON object body, or raise 400.
+
+    ``await request.json()`` raises ``JSONDecodeError`` on an absent or
+    malformed body, and an unhandled raise here escapes as a 500 with a full
+    traceback — the wrong contract for bad client input, and noise that buries
+    real failures in error monitoring (#689).
+
+    ``allow_empty`` accepts a request with no body at all and yields ``{}``,
+    for endpoints where every field is optional.
+    """
+    raw = await request.body()
+    if not raw.strip():
+        if allow_empty:
+            return {}
+        raise HTTPException(status_code=400, detail="request body must be a JSON object")
+    try:
+        body = await request.json()
+    except Exception:
+        raise HTTPException(status_code=400, detail="invalid JSON body")
+    if not isinstance(body, dict):
+        raise HTTPException(status_code=400, detail="request body must be a JSON object")
+    return body
+
+
 def _extract_task_error(task: dict[str, Any], fallback: str = "Ingestion failed") -> str:
     """Return the most useful per-file ingestion error from a task payload."""
     file_tasks = task.get("file_tasks")
@@ -262,7 +287,7 @@ async def get_settings(request: Request):
 @knowledge_agent_manage_router.post("/settings")
 async def update_settings(request: Request):
     engine = _get_engine(request)
-    body = await request.json()
+    body = await _json_body(request)
     knowledge_settings = body.get("knowledge", body)
     try:
         return engine.update_settings(**knowledge_settings)
@@ -310,7 +335,7 @@ async def search(
 
     engine = _get_engine(request)
     _ensure_enabled(engine)
-    body = await request.json()
+    body = await _json_body(request)
     scope = body.get("scope", "agent")
     query = body.get("query", "")
     limit = body.get("limit", engine._config.default_limit)
@@ -650,7 +675,7 @@ async def ingest_url(
 ):
     engine = _get_engine(request)
     _ensure_enabled(engine)
-    body = await request.json()
+    body = await _json_body(request)
     scope = body.get("scope", "agent")
     ensure_agent_scope_manage_if_needed(identity, scope)
     collection = resolve_collection(identity, scope, request)
@@ -748,7 +773,7 @@ async def delete_document(
 ):
     engine = _get_engine(request)
     _ensure_enabled(engine)
-    body = await request.json()
+    body = await _json_body(request)
     scope = body.get("scope", "agent")
     ensure_agent_scope_manage_if_needed(identity, scope)
     filename = body.get("filename", "")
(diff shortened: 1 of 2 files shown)


dolev31:
@sami-marreed one more from the same session — this one fell out of abuse-testing the knowledge API against a local Postgres prod-mode deployment, not from either of your reports.

Four endpoints 500'd on a missing or malformed JSON body. `POST /search`, `POST /documents/url` and `DELETE /documents` were fully unguarded; `POST /reindex` guarded only the empty case via `content-length`, so a malformed body still reached the unguarded parse. `PATCH /session/settings` was the one endpoint already doing it right, so the fix is just that guard generalized into `_json_body()` and applied everywhere — `request.json()` is now called in exactly one place.

Worth a look: the `allow_empty` flag. It preserves the documented no-body `/reindex` call while removing the `content-length` special case that was the actual hole, so both behaviours are decided in one function instead of two.

11 of the 12 new tests fail against `main`. The one that passes is the empty-body-reindex guard, which is precisely the behaviour that had to survive the change.

Independent of #687 and #688 — no shared lines, any merge order works.

dolev31:
All three review findings were valid and are fixed in `a669a3de`.

1. **`except Exception` was too broad** — it turned any failure during the read into a 400, including ones that are not the client sending bad JSON (a disconnect mid-read, for one). Narrowed to `ValueError` (which `JSONDecodeError` subclasses) and `UnicodeDecodeError`; anything else propagates.
2. **Whitespace-only bodies** — `raw.strip()` treated `b"   "` as absent, so `/reindex` with `allow_empty` accepted it. That is a malformed body, not a missing one. Now `not raw`, so whitespace falls through to parsing and 400s. The docstring says why the exception is narrow.
3. **The empty-body reindex assertion was weak** — `!= 400` would have passed on a 500. Now asserts 200.

Added coverage for whitespace-only bodies across every endpoint including `/reindex`, plus a non-UTF-8 body case for the `UnicodeDecodeError` path. 26 tests in the file, CI green.

sami-marreed (approved):

PR #692, opened 2026-08-20
fix(knowledge): make delete stop an in-flight ingest of the same file

### LINKED ISSUE #691: [Bug]: deleting a document while it is uploading does nothing (404) or is resurrected by the ingest
Filed without a parent: small self-contained bug (delete/ingest coordination).

## What happened

Clicking **Delete** on a knowledge document while an upload of that same filename is in flight does not work. Found by hand-testing the UI against a local Postgres (`storage.mode=prod`) deployment, then reproduced over the API.

There are two variants, both reachable from the UI:

**A. Delete during a first-time upload → 404, silently does nothing.**
`delete_document` calls `mark_deleting()`, which only matches a row in the documents table. That row is written by `add_document` near the *end* of ingest, so for almost the whole upload the document does not exist yet: delete returns `404 document not found`, nothing is cancelled, and the document appears anyway when the ingest finishes.

**B. Delete during a re-upload over an existing document → 200, then resurrected.**
The row exists, so the delete succeeds and the document disappears from the list. Then the in-flight ingest reaches `add_document` and writes it straight back. To the user, the document they just deleted returns on its own.

Measured (156 KB `.md`, `replace_duplicates=true`):

| variant | delete during ingest | document present after ingest finishes |
|---|---|---|
| A — new file, delete at +0.0s / +0.05s / +0.2s / +0.5s | **404** | **yes** (delete had no effect) |
| B — existing doc, delete at +0.0s / +0.1s / +0.3s | **200** | **yes** (resurrected) |

## Root cause

Delete and an in-flight ingest for the same filename are not coordinated:

- `delete_document` has no knowledge of active ingest tasks, so it neither cancels one nor waits for it.
- The ingest's `add_document` runs **outside** the per-collection lock that `delete_document` takes, so even a delete that wins the lock can be overwritten moments later.

## Expected

Deleting a document should make it not be there. Concretely: cancel any in-flight ingest that owns that filename, then remove whatever exists, and do not report 404 when an ingest was actually cancelled.

## How to reproduce

1. Run CUGA with knowledge enabled (`cuga start demo_knowledge`), any storage mode.
2. Upload a document large enough to take a second or two.
3. While it is uploading, delete that filename via the UI or:

```bash
curl -i -X DELETE -H "X-Agent-ID: cuga-default" -H "Content-Type: application/json" \
  -d '{"scope":"agent","filename":"medical-coverage.md"}' \
  http://127.0.0.1:7860/api/knowledge/documents
```

4. Wait for the upload to finish and list documents. The file is there.

## Environment

- `main` @ `8b45234a`
- Local Postgres 17 + pgvector, `storage.mode=prod`, `cuga start demo_knowledge`
- Observed 2026-08-20

## Notes

Now cheaply fixable because #685 makes `cancel_task` actually stop a running ingest and persist the outcome; before that there was nothing for delete to call.

## Checklist

dolev31:
## Bug fix

Fixes #691

> **Stacked on #688** (base branch `fix/685-knowledge-cancel-running-ingest`). It depends on that PR making `cancel_task` actually stop a running ingest — before it, there was nothing for delete to call. Review #688 first; this diff is only the two files listed below.

### Summary

Clicking **Delete** on a knowledge document while an upload of that filename is in flight did nothing. Found by hand-testing the UI against a local Postgres (`storage.mode=prod`) deployment.

### Root cause

Delete and an in-flight ingest were not coordinated at all:

- `delete_document` → `mark_deleting()` only matches a **documents row**, and that row is written by `add_document` near the *end* of ingest. So during a first upload the row does not exist, the request 404s, nothing is cancelled, and the ingest indexes the file anyway.
- During a re-upload the row does exist, so the delete succeeds and the document disappears — then the ingest reaches `add_document` and writes it straight back. `add_document` runs **outside** the per-collection lock that `delete_document` takes, so even a delete that wins the lock is overwritten moments later.

Measured on `main` (156 KB `.md`, `replace_duplicates=true`):

| variant | delete during ingest | present after ingest finishes |
|---|---|---|
| A — first upload, delete at +0.0 / +0.05 / +0.2 / +0.5s | **404** | **yes** (no-op) |
| B — existing doc + re-upload, delete at +0.0 / +0.1 / +0.3s | **200** | **yes** (resurrected) |

### Solution

`_stop_ingest_for()` cancels any active task owning that filename, then the normal delete flow runs.

Three details that matter:

- **Called before the collection lock is taken.** Waiting on a worker while holding that lock would deadlock — the worker needs it to insert.
- **If the cancel is refused** because the ingest is past its point of no return, wait for the task to terminalize. The terminal status is written *after* `add_document`, so waiting guarantees the delete runs last rather than being resurrected. Bounded at 30s so a wedged embedding call cannot hang the request; on expiry the delete proceeds anyway.
- **A missing documents row is no longer a 404 when an ingest was actually stopped.** The user's intent — "make this not be here" — is satisfied, and 404 would misreport work that really happened. A genuine miss (no document, no ingest) still 404s.

### Type of changes

### Testing

New `tests/unit/test_knowledge_delete_during_ingest.py` (4 tests, `@pytest.mark.unit`): both bug variants, plus two guards that the ordinary paths are unchanged (a genuine miss still raises `DocumentNotFoundError`; deleting an indexed document with nothing in flight still works).

Verified as real regression tests: **the two behaviour tests fail without the fix**, and the two unchanged-behaviour tests pass either way — so the fix is doing something and not breaking the normal path.

Re-verified end to end against real Postgres 17 + pgvector, driving the same HTTP surface the UI uses:

| variant | before | after |
|---|---|---|
| A, delete at +0.0 / +0.2 / +0.5s | 404, document present | **200, ingest `cancelled`, document absent** |
| B, delete at +0.0 / +0.2s | 200 then resurrected | **200, ingest `cancelled`, not resurrected** |

- `uv run pytest tests/unit/ -k knowledge` — 582 passed

### Note on the UI

No frontend change needed. `handleDelete` already checks `res.ok` and surfaces the error; the problem was entirely server-side — it was returning 404 (variant A) or succeeding and then being undone (variant B).

## Summary by CodeRabbit

* **Bug Fixes**
  * Deleting a document now stops any active upload or re-upload in progress.
  * Prevents cancelled uploads from recreating deleted documents.
  * Deletion completes safely even when an upload has not yet created a document record.
  * Preserves the expected not-found response when neither a document nor upload exists.

* **Tests**
  * Added coverage for deleting documents during initial uploads, re-uploads, and normal indexed-document deletion.

### DIFF
--- a/src/cuga/backend/knowledge/engine.py
+++ b/src/cuga/backend/knowledge/engine.py
@@ -63,6 +63,11 @@ _BACKGROUND_REINDEX_TASKS: set[Any] = set()
 # terminal: doing so would let a second ingest race a live insert.
 _TERMINAL_FILE_STATUSES = frozenset({"indexed", "failed", "skipped", "cancelled", "superseded"})
 
+# How long a delete waits for an ingest it could not cancel (already past the
+# point of no return) to finish writing. Bounded so a wedged provider call
+# cannot hang the delete request; on expiry the delete proceeds anyway.
+_DELETE_INGEST_WAIT_S = 30
+
 
 # Docling's plugin factory emits a WARNING every time it scans for plugins:
 #   "The plugin langchain_docling will not be loaded because Docling is being
@@ -3048,6 +3053,47 @@ class KnowledgeEngine:
 
     # --- Delete (5-step compensating flow per plan) ---
 
+    async def _stop_ingest_for(self, collection: str, filename: str) -> bool:
+        """Stop any in-flight ingest that owns ``filename``. Returns whether one was.
+
+        Delete used to ignore active tasks entirely (#691), which made the
+        button do nothing in two different ways:
+
+        * during a first upload the documents row does not exist yet — it is
+          written near the END of ingest — so ``mark_deleting`` found nothing
+          and the delete 404'd while the ingest carried on and indexed it;
+        * during a re-upload the delete succeeded, then the ingest reached
+          ``add_document`` and wrote the document straight back.
+
+        Must be called BEFORE taking the collection lock: waiting on a worker
+        while holding that lock would deadlock, since the worker needs it to
+        insert.
+        """
+        stopped = False
+        for task in await self._metadata.list_tasks(collection):
+            if not self._blocks_reupload(task, filename):
+                continue
+            task_id = task["task_id"]
+            logger.info(f"Delete {filename}: stopping in-flight ingest {task_id}")
+            await self.cancel_task(task_id)
+            stopped = True
+            # A cancel past the point of no return is refused, and that worker
+            # WILL write the document. Wait for it to reach a terminal status —
+            # which it only does after add_document — so our delete runs last
+            # instead of being resurrected by it.
+            deadline = time.monotonic() + _DELETE_INGEST_WAIT_S
+            while time.monotonic() < deadline:
+                current = await self._metadata.get_task(task_id)
+                if not current or current.get("status") in ("completed", "failed", "cancelled"):
+                    break
+                await asyncio.sleep(0.1)
+            else:
+                logger.warning(
+                    f"Delete {filename}: ingest {task_id} still running after "
+                    f"{_DELETE_INGEST_WAIT_S}s; deleting anyway"
+                )
+        return stopped
+
     async def delete_document(self, collection: str, filename: str) -> None:
         """Delete a document. Idempotent compensating flow across stores."""
         await self._ensure_metadata_ready()
@@ -3069,6 +3115,9 @@ class KnowledgeEngine:
                 f"Reindex in progress for {collection}; deletes are rejected until it completes."
             )
 
+        # Outside the lock on purpose — see _stop_ingest_for.
+        stopped_ingest = await self._stop_ingest_for(collection, filename)
+
         async with self._get_collection_lock(collection):
             # Re-check under the lock: reindex() flags the collection AND
             # snapshots its file_list under this same lock, so a delete that
@@ -3079,6 +3128,13 @@ class KnowledgeEngine:
                     f"Reindex in progress for {collection}; deletes are rejected until it completes."
                 )
             if not await self._metadata.mark_deleting(collection, filename):
+                if stopped_ingest:
+                    # There is no documents row because the ingest never got
+                    # far enough to write one — but we did stop it, so the
+                    # user's intent ("make this not be here") is satisfied.
+                    # 404 here would be a lie about work we actually did.
+                    logger.info(f"Delete {filename}: ingest stopped before it indexed anything")
+                    return
                 raise DocumentNotFoundError(filename)
             await self._ensure_vector_store_cached(collection)
             try:
--- a/tests/unit/test_knowledge_delete_during_ingest.py
+++ b/tests/unit/test_knowledge_delete_during_ingest.py
@@ -1,0 +1,152 @@
+"""Deleting a document must make it not be there, even mid-upload.
+
+Delete used to ignore in-flight ingests entirely (#691), so the button did
+nothing in two different ways:
+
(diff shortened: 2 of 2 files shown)


dolev31:
@sami-marreed this one came out of hand-testing the UI rather than a report — clicking Delete during an upload does nothing, and it fails in two different ways depending on whether the document already existed.

Stacked on #688 because it needs a cancel that actually stops a running ingest; the diff here is just `engine.py` + one new test file.

Two things worth your eye:

1. **`_stop_ingest_for` is called before the collection lock is taken.** Waiting on a worker while holding that lock deadlocks — the worker needs it to insert. That ordering is load-bearing and easy to "tidy" into a bug later.
2. **The wait after a refused cancel.** If the ingest is past its point of no return we cannot stop it, so we wait for it to terminalize — the terminal status is written *after* `add_document`, which is exactly why waiting stops the resurrection. Bounded at 30s so a wedged embedding call cannot hang the delete request.

Also worth flagging: my earlier stress test had a `delete during ingest` case that **passed** and shouldn't have — it asserted only "does not 500", and a 404 that silently does nothing satisfies that. Same weak-assertion trap CodeRabbit caught me on elsewhere in this stack. The new tests assert the actual end state instead.

dolev31:
@sami-marreed :rotating_light: **high priority** — rebased onto `main` now that #688 has merged, so this is no longer stacked and can go in on its own.

This one came from hand-testing the UI: **clicking Delete during an upload does nothing.** Two variants, both reproduced against real Postgres:

| | delete during upload | file present after upload finishes |
|---|---|---|
| first upload | **404** | **yes** — silent no-op |
| re-upload over an existing doc | **200** | **yes** — resurrected by the ingest |

The documents row is only written near the *end* of ingest, so for most of an upload there is nothing for `mark_deleting` to match; and `add_document` runs **outside** the collection lock delete takes, so even a delete that wins the lock gets overwritten moments later.

Two things worth reviewing:

1. **`_stop_ingest_for` is called before the collection lock is taken.** Waiting on a worker while holding that lock deadlocks — the worker needs it to insert. Load-bearing ordering.
2. **The wait after a refused cancel.** Past the point of no return we cannot stop the ingest, so we wait for it to terminalize — the terminal status is written after `add_document`, which is exactly what stops the resurrection. Bounded at 30s so a wedged embed cannot hang the delete request.

Also flagging honestly: my earlier stress suite had a "delete during ingest" case that **passed and should not have** — it asserted only "does not 500", and a 404 that silently does nothing satisfies that. The new tests assert the actual end state.

Tracked under #694 in the knowledge v2 epic (#444).

dolev31:
Good catch — that finding was valid and it was the same bug I was fixing, just relocated into the timeout window. Fixed in `624d3b91`.

On deadline expiry `_stop_ingest_for` logged a warning and **deleted anyway**. That worker is past its point of no return, so it still has an `add_document` ahead of it — the delete would have returned success and then been silently undone by that write. Reachable by any ingest whose insert phase runs longer than the deadline, which a large PDF easily does.

It now raises `IngestStillFinishingError` → **409 with a retry hint**. The delete genuinely did not happen, so saying so is the honest answer; the caller retries once the ingest terminalizes. Trading a bounded wait for a lie was the wrong trade and I should not have written it that way.

Added the coverage you asked for: a worker parked past the point of no return with the deadline collapsed via monkeypatch, asserting the delete **raises rather than half-completes**, the document survives, and a retry after the ingest finishes succeeds and sticks.

One honest note on that test: against the old `delete anyway` code it does not fail cleanly, it *hangs* — the delete blocks on the collection lock the parked worker holds. So it is a valid guard for the new contract, but the failure mode it demonstrates on the old code is a stuck request rather than a resurrection. Both are bad; the resurrection is the one that bites in production, where `add_document` runs after the lock is released.

611 knowledge tests pass, lint clean.

PR #693, opened 2026-08-20
fix(knowledge): serialize terminal task writes so a row cannot lie

dolev31:
## Bug fix

Refs #685

> **Follow-up to #688**, which was merged ~7 minutes before this commit was pushed, so this hardening did not make it into `main`. Everything else from that PR (including the pre-registration cancel guard) did land — verified against `main`. This is the one remaining piece.

### Summary

Closes the last known interleaving in the cancel path: a task row could read `cancelled` while its document was actually indexed.

### Root cause

`update_task` is last-write-wins with no CAS. So the worker's `status="completed"` could land **between** `cancel_task`'s read (which saw `running`) and its own `UPDATE`. The row then said `cancelled` over an indexed document.

It is a stale *label* rather than data loss — the vector store and the documents table still agree with each other, and the next `loadDocuments()` shows the truth — but it is the label the UI puts in front of the user, so it is worth closing.

### Solution

- `_task_write_lock` held across **read → decide → write** in `cancel_task`, and around each of the worker's four terminal writes (`completed`, `cancelled`, `superseded`, `failed`).
- Whichever side goes second re-reads inside the lock, sees the terminal status the other just wrote, and declines to overwrite it.
- **The cancel event is still set _before_ the lock is taken.** The worker's checkpoints are synchronous reads of that event, so a worker that runs while cancel is waiting on the lock still observes the cancel and aborts. Setting it after would reintroduce a window — worth a look in review, since it reads like something to "tidy up".

### Type of changes

### Testing

The test asserts the **invariant** rather than a fixed winner: either side may legitimately win the race, and what must never happen is the task row and the document disagreeing.

- `uv run pytest tests/unit/ -k knowledge` — 607 passed
- Re-validated on a real Postgres deployment (both a native PG17 and the helm `deployment/deploy-local-postgres.sh` chart on Kubernetes): 29/29 stress checks, plus kill-mid-ingest → restart → clean recovery.

An aside worth recording: my first attempt at this test held the lock by hand and deadlocked against the worker's terminal write. That is itself reasonable evidence the serialization is real.

### Remaining known limits (unchanged, neither a correctness hole)

- Cancel during a Docling parse does not *stop* the parse — `asyncio.to_thread` is not interruptible, so it burns CPU until it returns, then aborts. Harmless: that worker writes nothing.
- The point-of-no-return handoff is safe **because** there is no `await` between the check and the flag set. It is commented, but a future edit could silently break it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task cancellation and completion handling to prevent inconsistent final statuses.
  * Ensured cancelled tasks do not retain indexed documents, while completed tasks remain properly indexed.
  * Prevented cancellation from overriding successful ingestion results.

* **Tests**
  * Added coverage for race conditions between task cancellation and ingestion completion.

### DIFF
--- a/src/cuga/backend/knowledge/engine.py
+++ b/src/cuga/backend/knowledge/engine.py
@@ -1835,6 +1835,15 @@ class KnowledgeEngine:
         # head on, so a terminal write here would corrupt data.
         self._uncancellable_tasks: set[str] = set()
 
+        # Serializes "read a task's status, decide, write a terminal status".
+        # update_task is last-write-wins with no CAS, so without this a
+        # worker's status="completed" could land between cancel_task's read
+        # and its UPDATE — leaving the row "cancelled" while the document is
+        # actually indexed. Holding this across read+decide+write makes the
+        # check-then-act atomic: whichever side goes second re-reads, sees the
+        # terminal status the other just wrote, and declines to overwrite it.
+        self._task_write_lock = asyncio.Lock()
+
         # Reindex coordination flags (in-memory, single-process only — flock ensures this)
         self._reindex_in_progress: set[str] = set()
         self._reindex_deferred: set[str] = set()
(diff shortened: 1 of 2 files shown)


dolev31:
@sami-marreed :rotating_light: **high priority** — this is the one piece of #688 that missed the merge.

You merged #688 at 13:50; this commit was pushed at 13:57. Everything else from that PR did land (I verified the pre-registration cancel guard is in `main`), but the task-write lock is not — `grep _task_write_lock` on `main` returns nothing. So the cancel path on `main` today still has the interleaving where a row can read `cancelled` over an indexed document.

Small diff, one file plus a test. The thing worth your eye: **the cancel event is set _before_ the lock is taken**, deliberately. The worker's checkpoints are synchronous reads of that event, so a worker running while cancel waits for the lock still sees the cancel. Moving the event-set inside the lock would look tidier and would reintroduce a window.

Re-validated on both a native Postgres 17 and your `deployment/deploy-local-postgres.sh` helm chart on Kubernetes: 29/29 stress checks, kill-mid-ingest → restart → clean recovery, 607 knowledge tests green.

dolev31:
CI note before you look: the red `SDK integration` check here is a **pre-existing flake, not this change**.

It is `test_supervisor_advanced.py::TestSupervisorAdvanced::test_supervisor_coordination_with_variable_passing`, which asserts a literal variable name appears inside an LLM's free-text question. It fails with different phrasing every run:

- `assert 'user_alice_99' in 'What bonus amount (as an integer) would you like to process for Alice?'`
- `assert 'user_alice_99' in 'Could you let me know the bonus amount (as an integer) you'd like to process for Alice's account?'`

Evidence it is not mine:

- **It was the only failure in both of `main`'s failed Tests runs on 18 Aug** (`32167815716`, `32166690816`) — two days before this work started.
- This PR touches `knowledge/engine.py` and one knowledge test. Nothing on the supervisor/SDK path.
- `Unit (tests/)`, `Unit (graph)`, `Policy (unit)`, `Policy (e2e)` and `pgvector Integration` all pass here. `Unit Tests` is just the aggregator gate that fails when any sibling job does.
- The same job passed on #688, #690 and #692 with the same base.

I have re-run it. Worth someone filing separately — it has cost several re-runs across this batch and it gates merges.

Issue #705, opened 2026-08-23
[Bug]: variables-summary Created timestamps leak the host wall-clock into frozen-clock benchmark runs (model judges 2023-world dates against the real date)


Sergey-Zeltyn:
## What happened

In AppWorld benchmark mode the sandbox wall clock is frozen to the in-world date (#544, item 3), so agent code and tool data consistently see the world's "today" (May 2023 for the current task sets). But the **variables summary** that CugaLite injects into the prompt annotates every variable with a **host wall-clock** timestamp:

```
## qualifying_products
- Type: list
- Items: 1
- Description: Created during code execution
- Created: 2026-08-23 18:35:25        <-- real host time, not world time
- Value Preview: [{'produc…
```

The model now sees two conflicting "today"s in one context — the frozen world clock (2023) in tool data and the real date (2026) in prompt metadata — and sometimes trusts the metadata. When it does, every in-world-valid dated resource looks expired.

Observed concretely on AppWorld `test_med` task `b0934aa_1` (buy a board game): the world seeds five payment cards, expiries 2023–2024, exactly one funded and valid in-world. The agent's final answer:

> "**Payment method** – All payment cards returned by the Amazon API are expired **relative to the current date (2026-08-23)**. An exp…"

— and the purchase is abandoned, failing the task. The same code on the same commit passed this task in the immediately preceding run, where the model happened not to consult the timestamps: **the leak is intermittent**, which makes it a silent reliability tax rather than a deterministic failure.

Impact beyond this one task: any date-sensitive AppWorld task (card expiry, "this month", "last 7 days", scheduling) can be contaminated, and the intermittency means it has plausibly contributed to chronic-task failure baselines attributed to model reasoning. In the #676 A/B failure analysis, three payment-trap tasks fail 8/8 runs partly on "expired card" reasoning against in-world-valid cards (see `cuga-eval/benchmarks/appworld/evaluation_bundles/TODOS_AB_FAILURE_DISTRIBUTION.md`, chronic-task drill-downs; and the #704 validation runs `20260823_182136` / `20260823_184035` for the run-to-run flip on identical code).

## Root cause

`src/cuga/backend/cuga_graph/state/agent_state.py`:

- `VariableMetadata.__init__` stamps `created_at = datetime.now()` (line ~61) — this runs in the **harness process**, which the sandbox clock freeze does not cover.
[...]

Issue #676, opened 2026-08-17
[Bug]: enable_todos spends 11-13% of the run budget on no-op turns and emits false completion signals


Sergey-Zeltyn:
Filed **without a parent epic** — self-contained defect in the `create_update_todos` prompt contract, with a narrow repro (one flag, two bundles).

## What happened

Offline trace analysis of two AppWorld evaluation bundles run with `ADVANCED_FEATURES__ENABLE_TODOS=true` shows the todos mechanism consistently costs 11–13% of the run budget while supplying no measurable accuracy benefit — and in several cases actively causes failures. Four distinct defects, all reproduced in both bundles.

Bundles (both `aws/gpt-oss-120b`, temp 0.1, `max_completion_tokens` 16000, `REFLECTION_ENABLED=true`, `LITE_MODE=false`, model identity verified from every Langfuse GENERATION):

| | easy split | medium split |
|---|---|---|
| bundle | `20260806_124527_compare_gpt-oss-120b-easy-todos-part1` | `20260817_114941_compare_gpt-oss-120b-med_TODOS` |
| tasks / strict Pass@1 | 29 / 23 = 79.3% | 40 / 21 = 52.5% |
| cuga-agent commit (reflog) | `03760b9d` | `39c4ceac` |

### 1. A todo-only turn costs two LLM calls and buys nothing

`mcp_prompt.jinja2:29` requires the first Python execution of a multi-step task to be a `create_update_todos(...)` block **and nothing else**, and `mcp_prompt.jinja2:187` requires every later todo update to sit alone in its own block. With reflection enabled, each such turn is followed by a full reflection generation before the agent may act again.

| | easy | medium |
|---|---:|---:|
| todo-only executor turns | 43 / 193 = 22.3% | 51 / 313 = 16.3% |
| overhead: LLM calls (turn + its reflection/classifier) | **86 / 489 = 17.6%** | **102 / 772 = 13.2%** |
| overhead: tokens | **11.2%** | **11.1%** |
| overhead: LLM latency | **16.6%** | **11.8%** |
| reflections whose entire content is "nothing happened, only todos were updated" | 19 (14 tasks) | 15 (12 tasks) |

Worst-hit tasks paid a third of their budget: `29a7b7e_1` 35.0% of tokens / 36.9% of latency, `21abae1_1` 34.4% / 37.6%, `fd1f8fa_1` 30.3% / **62.9%**.

[...]

PR #706, opened 2026-08-23
fix(cuga_lite): exhaust alternatives before declaring a task impossible (#704)

### LINKED ISSUE #704: [Bug]: CugaLite abandons a task at the first recoverable obstacle instead of exhausting alternatives (mid-task false impossibility)
Filed without a parent (no fitting [Feature] under the evaluation epic; placement chosen deliberately).

## What happened

When an API call fails mid-task with a **resource-specific, recoverable error** — a declined card, an insufficient balance, an invalid reset code — CugaLite reports the task as impossible instead of trying the alternatives it already has on screen. The conclusion is delivered confidently ("none of the saved payment cards have enough balance", "no delete capability exists for other apps") even when the winning move is verifiably one call away.

This is the complementary case to two existing issues:

- #610 (fixed) covered the **turn-1** unverified refusal — quitting before any tool call. That fix works: the runs below contain zero turn-1 refusals.
- #599 (open) covers the mirror image — **never** adapting, re-issuing the identical rejected call for dozens of turns.

Nothing covers the middle: *tried once, hit a real but recoverable obstacle, quit without exhausting alternatives.* All three share the same underlying deficiency — no post-error strategy adaptation.

### Evidence

AppWorld `test_med` (40 tasks), `aws/gpt-oss-120b`, temp 0.1, 8 runs across the 4 arms of the #676 todos A/B (Aug 21–23, 2026; behavior identical on main `8b45234a` and branch `1fe0c595` — todos-independent). Full analysis: `cuga-eval/benchmarks/appworld/evaluation_bundles/TODOS_AB_FAILURE_DISTRIBUTION.md` (root-cause rollup + chronic-task drill-downs).

**This family is 37/172 (22%) of all med failed task-runs — the third-largest root-cause family — and every instance was verified against the world DBs / API implementations, not just the agent's claims:**

| task | fails | obstacle hit | verified winning move the agent never tried |
|---|---|---|---|
| `d9987f6_1` (buy extension cord) | 8/8 | "Your payment card doesn't have enough balance" on cards #178, #180 | Card **#177 (Discover, valid 11/2023) holds the funds** — it was in the `show_payment_cards` response the agent had already fetched; the agent mis-enumerated "only #178 and #180 are non-expired" every run |
| `b0934aa_1` (buy board game <$20) | 8/8 | same | Funded card **#250 (HSBC, valid 5/2024)** in the fetched list; agent claims "card on file is expired, no others" (false — #249 and #250 are valid) |
| `9ef798c_1` (accept $520 Venmo request) | 8/8 | "Insufficient money in the payment card" on every $20 card | `approve_payment_request` **uses the Venmo balance when `payment_card_id` is omitted** (documented in the param description); balance is $510 — one bank-transfer top-up away (the task's expected model changes include `venmo.BankTransfer`) |
| `3f3c139_1` (remove expired cards, all apps) | 7/8 | none — claims "No delete capability exists for other apps" after cleaning Amazon | `spotify` and `venmo` **both expose card-delete APIs**; 2 runs additionally fabricated completion |
| `7574325_1` (change Venmo password) | 6/8 | password-reset codes 422 "Invalid password reset code" | The **authenticated profile-update path** (current password available via the supervisor app); the agent locked onto the forgot-password flow and quit when codes failed |

Three of these fail **8/8 runs in every arm** — a hard floor on the med score. Full rescue ceiling for a todos-off config: ~+11 pp pass@1 (46% → 57%); even a partial +5–8 pp would exceed the measured effect of the entire todos mechanism in the #676 A/B.

### Root cause

The executor prompt (`src/cuga/backend/cuga_graph/nodes/cuga_lite/prompts/mcp_prompt.jinja2`, 408 lines) contains **no obstacle-handling guidance**: nothing tells the model that a resource-specific failure is information about *that resource*, not about the task. gpt-oss-120b's default under this prompt is to generalize one or two failed attempts into "the task cannot be done" — the same premise-generalization that #610 fixed for turn 1, resurfacing mid-task where `classify_nl_auto_continue`'s hard-stop on claimed inability is (correctly) still active.

### Proposed fix — additive prompt rule (recommended)

Add to the execution-rules section of `mcp_prompt.jinja2`:

> **Obstacles are per-resource, not per-task.** When an action fails with an explicit resource-specific error (insufficient balance, expired card, invalid code, permission on one object) and no side effect occurred, do not conclude the task is impossible. Enumerate and try every remaining alternative first — every other card in the list, the account balance (many payment APIs use it when the card id is omitted), a top-up or transfer path, an equivalent API in another connected app. Only declare failure after stating which alternatives you tried and why each is exhausted.

Why this is low-risk:

- Additive text; no code-path or graph changes; todos-independent.
- Retry storms are double-bounded: the rule gates on "explicit error and no side effect" (the AppWorld payment APIs raise *before* creating anything), and the #559 `max_tool_calls` cap bounds total volume.
- Worst case on genuinely impossible tasks: a few extra read calls before the same conclusion — a cost, not a correctness, risk.

Alternative considered (not recommended as first step): a harness-side hint appended to recognized recoverable errors. More machinery, provider-specific error matching, and #596 shows error-text interpretation is already a fragile area.

## How to reproduce

1. Run the AppWorld `test_med` suite via cuga-eval (`./scripts/eval.sh --benchmark appworld`, model profile `gpt-oss` on AWS, temp 0.1, todos off).
2. Observe tasks `d9987f6_1`, `b0934aa_1`, `9ef798c_1` fail every run (currently 0/8 across the #676 A/B arms), `3f3c139_1` and `7574325_1` nearly every run.
3. In the trace/trajectory, the final prose declares the task impossible after 1–2 obstacle errors, while the enumerated winning alternative (table above) is present in earlier tool output or documented API defaults.

Validation plan for a fix: 2 med runs with the patched prompt, judged on the 5 named tasks rather than the aggregate (any consistent passes on the three 0/8 tasks are immediate signal; the aggregate needs ~3 runs to clear the ±5.6 pp noise floor at n=40). Guard cost with tok/task on the same runs.

## Environment

- cuga-agent `8b45234a` (main) and `1fe0c595` (fix/676 branch tip) — behavior identical on both
- Model: `aws/gpt-oss-120b`, temp 0.1, max_completion 16000, REFLECTION_ENABLED=true
- Benchmark: AppWorld `test_med` (40 tasks), cuga-eval harness, Aug 21–23 2026 bundles

Representative claimed outcomes (from Langfuse traces, `trace_summary.py --claimed`):

> "All available, non-expired payment methods lack sufficient funds, and there is no tool to add a new funded card or top-up an existing one." — `d9987f6_1` (funded card #177 was in the fetched list)

> "All expired payment cards have been successfully removed from the Amazon account, and the task has been marked complete. No delete capability exists for other apps, so no further action is possible." — `3f3c139_1` (spotify/venmo delete APIs exist)

Evaluation bundles: `20260822_023120…` / `20260822_083652…` (todos OFF), `20260821_175006…` / `20260821_214314…` (TODOs-old), `20260822_141326…` / `20260822_165400…`, `20260823_000815…` / `20260823_073305…` (branch arms) under `cuga-eval/benchmarks/appworld/evaluation_bundles/`.

### Related

- #610 (closed) — turn-1 unverified refusal; this is the mid-task complement
- #599 (open) — no course correction after repeated rejection (mirror behavior)
- #596 (open) — error-text misinterpretation ("already exists" treated as failure)
- #676 — the A/B whose failure analysis surfaced and quantified this family

## Checklist

Sergey-Zeltyn:
## Bug fix

Fixes #704

### Summary

Adds an **Obstacle Handling & Recovery** rule to the CugaLite executor prompt (`mcp_prompt.jinja2`, which previously contained no obstacle-handling guidance): a resource-specific error (insufficient balance, expired card, invalid code) is information about *that resource*, not the task. Before declaring failure the agent must (1) try every remaining alternative from data it already fetched, (2) re-read the failing tool's parameter docs for built-in fallbacks (e.g. balance-when-card-omitted) and top-up/transfer paths, (3) verify a claimed missing capability via `find_tools` — this bullet is gated on `{% if enable_find_tools %}` — and (4) state what was exhausted.

Targets the "premature impossibility" family: 22% of AppWorld `test_med` failures (37/172 task-runs, gpt-oss-120b, #676 A/B), including three tasks at 0/8 where the winning move was verifiably one call away. Complements #610 (turn-1 refusal, fixed) and #599 (no course correction, open). Additive prompt text only; retry volume double-bounded by the no-side-effect gate and the #559 `max_tool_calls` cap.

### Testing
  - Template renders verified with `enable_find_tools` on/off; `ruff check` + `format --check` clean; `tests/unit` 1427 passed (3 `test_sdk_citations` failures are pre-existing on the clean tree in a full run; pass in isolation).
- [x] **Targeted validation** — 2 runs on the five #704 tasks (todos OFF, bundles `20260823_182136` / `20260823_184035`): 2/5 and 1/5 vs ~0.4/5 expected under the old prompt (baselines 0/8, 0/8, 0/8, 1/8, 2/8). Rendering-level check: all 10 traces contain the new rule; **zero bare "task is impossible" declarations remain** — every failure now shows alternative exhaustion first (e.g. `b0934aa_1` found and used the funded HSBC card it previously claimed didn't exist; `9ef798c_1` now attempts the top-up path).
- [x] **Full `test_med` run** (40 tasks, todos OFF, bundle `20260823_203223_compare_gpt-oss-120b-med-issue704`): **22/40 = 55%** vs pooled todos-OFF baseline 37/80 = 46.2% (per-run 42.5% / 50%). Matched-task flip table vs baseline: **10 tasks rescued, 0 regressed** (exact McNemar p = 0.002); rescues include `3aa1a22_1` (0/8 baseline), `3f3c139_1` (1/8), `7574325_1` (2/8), `3d9a636_1`. **No cost increase**: 246k tok/task vs 300k baseline (14.8 vs 16.2 LLM calls/task).
- Caveats: single full-scale run (single-run binomial SE ±7.9 pp at n=40 — the flip-table asymmetry, not the headline delta, is the robust signal); two of the three remaining target-task failures (`b0934aa_1`, `d9987f6_1`) are contaminated by the wall-clock leak filed as #705 during this validation.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced handling of resource-specific obstacles across items, accounts, and endpoints.
  * Available alternatives are now rechecked, documented fallback options are consulted, and missing capabilities are verified when supported.
  * Improved status reporting identifies alternatives that were attempted or ruled out before a task is declared unsuccessful.


Sergey-Zeltyn:
Pushed 2620a37e in response to the review feedback (CodeRabbit's line-205 comment and the parallel human-review concern about payment-instrument guidance).

**What changed:** the rule is reduced to its domain-neutral epistemic core — a resource-specific error doesn't prove task impossibility; re-verify alternatives in already-fetched data and documented fallbacks; verify claimed missing capabilities via `find_tools` (still gated on `enable_find_tools`); state what was exhausted before declaring failure.

**Why the requested authorization guard is no longer needed:** the concern applied to the original wording, which *prescribed actions* — iterate payment cards, look for top-up/transfer paths. All of that is removed; the prompt no longer mentions payment instruments, funding, or any concrete action class. What remains only governs conclusions and read-only checks (re-reading data already fetched, re-reading docs, capability discovery), so there is no funding action left to gate. If action-prescriptive guidance is ever reintroduced (e.g. benchmark-side), the authorization gate should come with it.

**Validation of the softened wording** (5-task probe, 2 runs, todos OFF, bundle `20260824_130123`; new wording verified present in all 10 traces):
- 2/10 task-run passes vs ~0.75/10 baseline expectation (original wording: 3/10).
- `9ef798c_1` passed for the **first time in any configuration** (0/10 prior): the agent discovered the balance-funding path on its own, with no funding hint in the prompt.
- Zero bare "task is impossible" declarations in all 10 trajectories — every failure shows documented alternative exhaustion; the behavioral goal survives the softening.
- Known remainder: `d9987f6_1`/`b0934aa_1` require a cross-app card-onboarding step (funded card intentionally absent from the amazon app; see the correction on #704) — a discovery-depth problem (#680) rather than an obstacle-handling one.

sami-marreed (approved):

Issue #701, opened 2026-08-23
[Feature]: Supervisor-level final_answer — shape supervisor-composed answers


segevshlomovIBM:
Part of #621.

## What you want and why

#621 (implemented on `feature/621-configurable-final-answer`) applies the deterministic answer function wherever an agent's own `FinalAnswerNode` finalizes the answer. One documented v1 gap remains: **answers synthesized by a `CugaSupervisor` itself** — as opposed to sub-agent answers it forwards — are never shaped.

Concretely, the supervisor SDK wrapper has no `FinalAnswerNode` (its `supervisor_sdk_callback` routes to END), and `CugaSupervisor.__init__` has no `final_answer` parameter, so `CugaSupervisor(agents={"a": CugaAgent(final_answer=fn)})` delivers unshaped supervisor-composed prose even though every sub-agent is configured.

Naively re-applying the function at the supervisor exit was implemented and reverted during #621's review: the second pass runs on already-finalized text (resolved `[n]` citation chips, harmony-stripped), violating the seam's ordering contract and able to destroy `state.sources` — 4 independent review angles converged on this.

## How it could work

- Add `final_answer=` to `CugaSupervisor.__init__` (same three forms as `CugaAgent`).
- Apply it only to text the supervisor itself composed, never to a forwarded sub-agent answer that already passed a `FinalAnswerNode` — this needs provenance (e.g. track whether `final_answer` came from delegation forwarding vs supervisor synthesis), which is the actual design work.
- The supervisor's tool-approval-denied exit (`sdk.py`, `supervisor_sdk_callback`) goes straight to END and must be included whatever the mechanism — inline per-branch calls were rejected in #621 review as unmaintainable.

## Links or extra context

- #621 — scope note and the reverted implementation (see PR review discussion).
- `finalize_answer` in `nodes/answer/final_answer.py` — the ordering contract any supervisor-level application must honor.

Issue #462, opened 2026-07-09
[Epic] Tau-bench (tau2 + tau3)


adiasaf:
## What you want and why

Integrate and score **tau2** and **tau3** benchmarks; credible multi-turn tool-calling and domain-policy signal for production-grade agent quality.

## How it could work

- cuga-eval integration (tau2 bench, baselines)
- cuga-agent Tau3 integration
- Progress toward public leaderboard goals

## Links or extra context

- Sibling epics: Continuous evaluation & regression, Vakra evaluation
- cuga-eval#84, cuga-eval#147, #211
- Policy measurement methodology links to Continuous epic (policies track)

PR #707, opened 2026-08-24
feat(agent): configurable final answer — LLM instructions, a deterministic function, or both (#621)

### LINKED ISSUE #621: [Feature]: Configurable final answer — LLM instructions, a Python function, or both
Part of #612.

> **Status:** implemented on branch [`feature/621-configurable-final-answer`](https://github.com/cuga-project/cuga-agent/tree/feature/621-configurable-final-answer) (8 commits, PR pending). Validated: 17 unit tests, live E2E on watsonx `gpt-oss-120b` + Ollama `gpt-oss:20b`, the UI `/stream` path with an env-configured function, and dead-provider failure. Spec below synced to the implementation after an 8-angle self-review.

## What you want and why

A new user has no reachable way to control how the final answer is produced. Two things are missing, and one of them is only missing its plumbing:

- **LLM with my own instructions** — already fully wired, but unreachable. `instructions.toml` has an `[answer]` section → `configurations/instructions/default/answer.md` → `InstructionsManager` (maps `FinalAnswerAgent` → `answer`) → `final_answer_agent.py:86` → rendered by `prompts/system.jinja2:18-21`. `answer.md` ships at **0 bytes**: the slot is built and empty, and the only way to fill it is editing a markdown file **inside the installed package**. No SDK parameter, no settings key, no docs.
- **A deterministic Python function** — does not exist at all. There is no hook that post-processes the delivered answer.

The consequence is that benchmark or channel-specific answer shaping gets patched into core: `appworld_final_answer_plain` is benchmark logic hardcoded in three core files. Vakra and Tau-bench (#462) each want a different shape.

There is also no declared order for the transforms already applied to `final_answer`: variable substitution and citation resolution run inline across 7 sender branches, and #558 adds control-token stripping. Three PRs have each claimed a different "single choke point" in the same function.

## How to reconstruct

```bash
# 1. The LLM instructions slot exists, is empty, and lives inside the package
wc -c src/cuga/configurations/instructions/default/answer.md      # 0
grep -n "instructions" \
  src/cuga/backend/cuga_graph/nodes/answer/final_answer_agent/prompts/system.jinja2

# 2. No function hook exists
grep -rn --include='*.py' "final_answer_function\|answer_formatter" src/cuga   # no matches

# 3. Benchmark answer shaping is hardcoded in core instead
grep -rn --include='*.py' "appworld_final_answer_plain" src/cuga
#   config.py:144, sdk.py:2731, final_answer_agent.py:144

# 4. The transforms that do run have no declared order — 7 inline call sites
grep -n "apply_citation_resolution" \
  src/cuga/backend/cuga_graph/nodes/answer/final_answer.py
```

## How it could work

Two independent settings in the user's own vocabulary. Defaults are a no-op.

```toml
[final_answer]
llm = true            # run the LLM composer (today's features.final_answer, surfaced)
instructions = ""     # guidance for the LLM composer — fills the existing {{ instructions }} slot
function = ""         # dotted path to a deterministic (str) -> str applied to whatever came out
```

| `llm` | `function` | Result |
|---|---|---|
| on | — | today's default |
| on | set | LLM composes, then the function normalizes |
| off | set | the function alone |
| off | — | raw candidate answer |

SDK, with shorthands plus a typed object for the combined case (mirrors the `ToolCalling` shape from #472):

```python
CugaAgent(final_answer="Answer with the bare value only.")                 # LLM + instructions
CugaAgent(final_answer=my_fn)                                              # function replaces the LLM
CugaAgent(final_answer=FinalAnswer(instructions="...", function=my_fn))    # both
CugaAgent(final_answer=FinalAnswer(llm=False, function=my_fn))             # explicit function-only
```

Implementation — one ordered seam in `final_answer.py`, replacing the 7 inline call sites:

```
finalize_answer(state):
  1. variable substitution   — always (existing)
  2. sanitization            — always (#558, keeps its own gate)
  3. answer function         — if configured                <-- new
  4. citation resolution     — always (existing)
```

LLM composition already runs upstream of this seam, so "LLM then function" needs no extra mechanism — the function always receives the answer as it stands and returns the final string, one `(str) -> str` contract regardless of what produced it.

Notes:

- The function is resolved via the existing dotted-path pattern (`page_understanding.transformer_path`, `config.py::get_class`), so a new answer contract ships in the caller's own package — no core edit.
- Failure logs and returns the input unchanged; it must never break answer delivery.
- Applies to `state.final_answer` only, not the chat transcript — two output surfaces, and #558 had to handle both separately.
- Sanitization stays mandatory and is not user-replaceable, so a user function can never silently disable it.
- Contract: pure and deterministic; idempotent recommended. Applied **once per delivered answer** — the supervisor forward branch deliberately skips re-application (the sub-agent already applied it; a second pass would feed the function resolved `[n]` chips).
- The SDK `str` form injects guidance into the CugaLite prompt (`special_instructions` channel) — on the SDK path the FinalAnswerAgent composer never runs, so a callable disables nothing; composer on/off remains `features.final_answer`, independent of this parameter.
- Bare classes passed as `final_answer` raise a clear `TypeError` (classes are callable — a forgotten `()` must not silently install).
- `apply_citation_resolution` keeps its current name and signature — it has 12 direct call sites in tests.

## Why not extend the existing OutputFormatter policy?

Raised in review; recorded here so it doesn't need re-litigating.

`PolicyType.OUTPUT_FORMATTER` already has persistence, triggers, priority ordering, a UI, and an SDK method, and a new deterministic `format_type` would be a small change — so it is a reasonable fit for the **function** half. It is not a fit for the **instructions** half, nor for a guarantee:

- `format_type="markdown"` is a *second* LLM call that rewrites an already-composed answer, and it runs on `settings.agent.code.model` (`enactment.py:832`), not the final-answer model. That is an extra round-trip on the wrong model to do what the existing (empty) `{{ instructions }}` slot does for free.
- `apply_output_formatter_policies` is called in exactly three places — CugaLiteCallback, the SDK callback, the supervisor callback. **Never in `FinalAnswerNode`**, so the full decomposition path never applies it.
- It returns early unless `settings.policy.enabled`. Benchmarks routinely run policy-off, so a contract routed through it would silently not apply.
- Its input is selected as last `AIMessage` in `chat_messages` → `context.agent_response` → `state.final_answer`. A deterministic contract must know exactly which string it transforms.

The two are complementary: OutputFormatter is conditional, user-authored, LLM-driven *presentation*; this issue is a deployment-level *answer contract*. They already run in the right order in the CugaLite path — formatter at the callback, this seam in `FinalAnswerNode`.

**Follow-up, not this issue:** if we later want deterministic transforms available as policies too, add `format_type="function"` that calls the *same* resolver introduced here — one implementation, two entry points, and the answer contract does not inherit a dependency on the policy system being enabled.

## Acceptance

- Defaults unchanged: `llm = true`, `instructions = ""`, `function = ""` produce byte-identical answers, locked by a regression test.
- All four combinations covered by tests, including LLM-then-function ordering.
- Failure paths: function raises, returns a non-`str`, or has an unresolvable path — answer still delivered.
- A reference function in the test suite demonstrates a third-party answer contract with no core edit.
- Docs: the two settings, the three SDK forms, and the four combinations.

**Out of scope (v1).** Answers **synthesized by a `CugaSupervisor` itself** (as opposed to forwarded sub-agent answers) are not shaped — follow-up issue tracks supervisor-level `final_answer` support. Benchmark-specific **composition** (`appworld_final_answer_plain`: its own prompt, output schema, ACTION/QUERY classifier, and post-LLM runnable, duplicated in `sdk.py:2718-2780`) is a separate axis and is not addressed here. It stays as-is; this issue only stops it from being the *only* way to control the answer.

## Sequencing

1. ~~Land #558 first~~ — done; the seam sanitizes before the function and keeps #558's strip inside citation resolution as defence in depth.
2. This issue as one PR from `feature/621-configurable-final-answer`: the two settings, the `finalize_answer` seam, and the SDK `final_answer=` parameter.
3. Follow-ups (filed separately): supervisor-level `final_answer`; `format_type="function"` on OutputFormatter reusing this resolver.

## Links or extra context

- #558 and #554 — final-answer post-processing already in flight; the seam should absorb #558's strip as step 2 rather than compete with it.
- #462 — Tau-bench, a second consumer needing a different answer contract.
- Prior art: `page_understanding.transformer_path` for the config shape; `ToolCalling` (#472) for the SDK object.

segevshlomovIBM:
## Feature

Closes #621

### Summary

A new user gets two knobs for the final answer, in their own vocabulary — with defaults that are byte-identical to today:

```toml
[final_answer]
instructions = ""   # guidance for the LLM composer (fills the existing, empty {{ instructions }} slot)
function = ""       # dotted path to a pure (str) -> str applied to the delivered answer
```

```python
agent = CugaAgent(final_answer="Answer with the bare value only.")   # str -> LLM guidance
agent = CugaAgent(final_answer=my_fn)                                 # callable -> deterministic function
agent = CugaAgent(final_answer=FinalAnswerConfig(instructions="...", function=my_fn))  # both
```

Implementation is one seam: `FinalAnswerNode.finalize_answer` (harmony strip → answer function → citation resolution) replacing the inline `apply_citation_resolution` calls on the terminal branches — `apply_citation_resolution` itself is untouched. The function is resolved via the existing `config.get_class` dotted-path pattern, so a benchmark/channel answer contract ships in the caller's own package with **no core edit** (`DYNACONF_FINAL_ANSWER__FUNCTION=my_pkg.fmt.bare`). Contract: pure, deterministic (idempotent recommended), applied once per delivered answer, and can never break delivery — failures log and deliver unformatted.

Why: controlling the exact answer string is a recurring need (benchmark contracts — Vakra, Tau-bench #462 — and fixed-format channels), and today the only precedent is `appworld_final_answer_plain` hardcoded in core. This stops that pattern from repeating. Design discussion, the OutputFormatter-policy alternative (and why it was rejected for this), and scope decisions are recorded in #621.

Scope notes (deliberate, documented in README): supervisor-composed synthesis is not shaped (#701 tracks it); the AppWorld pipeline keeps its own answer path; the chat transcript carries pre-finalize text. `[final_answer].instructions` applies to the FinalAnswerAgent composer — in default lite mode use `special_instructions` (the SDK `str` form does this automatically).

Self-review: an 8-angle review ran before this PR; it removed a supervisor-callback re-application (fed the function resolved [n] chips), added an explicit rejection of classes passed as `final_answer` (classes are callable — a forgotten `()` silently installed), covered the reasoning-model empty-answer fallback, and net-shrank the diff.

### Testing
- 17 new unit tests (`tests/unit/test_final_answer_function.py`, `@pytest.mark.unit`): byte-identical defaults lock, both config routes, injected-wins precedence, ordering (harmony-stripped input, raw [sN] never resolved [n]), all failure paths, class-rejection regression, instructions precedence, lazy export
- Untouched suites green: `test_final_answer_citations.py`, `test_sdk_citations.py`, `test_envelope_citations.py`, `test_knowledge_citation_e2e.py` (50 passed), full cuga_lite dir (368 passed); repo-wide ruff clean; `import cuga` still ~7ms
- Live E2E, watsonx `gpt-oss-120b`: function-only → exact `'2102650'`; instructions → `'Population: 8799800'`; combined → deterministic value; env-var-only route with an external module → works with zero core edits
- Live E2E, Ollama `gpt-oss:20b`: all three SDK forms pass (one retry on the model's known sign-off flake, unrelated to this change)
- UI path: `demo_crm` stack with `DYNACONF_FINAL_ANSWER__FUNCTION` set, `POST /stream` as the frontend does — SSE `FinalAnswerAgent` event delivered the shaped `"1000"` while the transcript kept the prose ("There are 1000 accounts…"), per the documented split
- Failure mode: unreachable provider → clean `APIConnectionError`, no interaction with the new code

<details>
<summary>E2E example (the harness route — no SDK, no core edit)</summary>

```python
# my_pkg/fmt.py — ships in the benchmark repo, not in CUGA
import re
def bare_number(text: str) -> str:
    m = re.search(r"\d[\d,]*", text)
    return m.group(0).replace(",", "") if m else text
```

```bash
DYNACONF_FINAL_ANSWER__FUNCTION=my_pkg.fmt.bare_number cuga start demo_crm
# delivered answer: "1000"   (model composed: "There are 1000 accounts in your territory in the CRM.")
```
</details>

## Summary by CodeRabbit

- **New Features**
  - Added configurable instructions for tailoring generated final answers.
  - Added optional deterministic formatting functions for consistent response presentation.
  - Configure final-answer behavior through TOML settings, SDK options, inline functions, or reusable configuration objects.
  - Formatting applies consistently to generated and fallback answers while preserving citation handling.
  - Original answers are retained when formatting fails or returns an empty result.
  - Added validation and public access to final-answer configuration.

- **Documentation**
  - Added setup examples, processing order, failure behavior, instruction precedence, and supported-scope guidance.


src/cuga/backend/cuga_graph/nodes/answer/answer_function.py
```diff
@@ -0,0 +1,77 @@
...
+    Raises with a clear message on a dotless path or a non-callable target.
+    (Callers on the answer path catch and log — a bad path is reported per
+    answer rather than crashing delivery.)
+    """
+    if "." not in path:
+        raise ValueError(f"[final_answer].function must be a dotted 'module.attr' path, got '{path}'")
+    fn = get_class(path)
+    if not callable(fn):
+        raise TypeError(f"[final_answer].function '{path}' is not callable")
+    return fn
```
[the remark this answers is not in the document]
segevshlomovIBM: Fixed in 0098e85f (first commit of the batch): `resolve_answer_function` now rejects classes with a clear TypeError, and `FinalAnswerConfig(function=<class>)` raises at `CugaAgent.__init__`. Regression tests added for both paths (`test_resolver_rejects_class_paths`, `test_settings_class_path_delivers_original`, and the extended `test_sdk_rejects_class_passed_as_final_answer`).

src/cuga/config.py
```diff
@@ -142,6 +142,8 @@ def get_all_paths(config, parent_key=""):
     Validator("observability.litellm_local_model_cost_map", default=True),
     Validator("advanced_features.benchmark", default="default"),
     Validator("advanced_features.appworld_final_answer_plain", default=False),
+    Validator("final_answer.instructions", default=""),
+    Validator("final_answer.function", default=""),
```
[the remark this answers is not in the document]
segevshlomovIBM: Fixed: `is_type_of=str` added to both validators, and — since config.py catches ValidationError and only warns — both consumers now treat non-str values as unset with a warning instead of raising (`resolve_final_answer_instructions`, `apply_answer_function`). `test_non_str_settings_values_never_break_delivery` locks it.

tests/unit/test_final_answer_function.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+@pytest.fixture(autouse=True)
+def _clean_final_answer_settings():
+    """Snapshot/restore the [final_answer] settings around every test."""
+    before_fn = settings.final_answer.function
+    before_instr = settings.final_answer.instructions
+    yield
+    settings.set("final_answer.function", before_fn)
+    settings.set("final_answer.instructions", before_instr)
```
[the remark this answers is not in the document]
segevshlomovIBM: Fixed: the autouse fixture now force-empties both `[final_answer].*` values before each test and restores the originals after, so developer `DYNACONF_FINAL_ANSWER__*` env vars can't leak into empty-settings assertions.

src/cuga/backend/cuga_graph/nodes/answer/final_answer_agent/final_answer_agent.py
```diff
@@ -84,7 +85,10 @@ async def run(self, input_variables: AgentState) -> AIMessage:
         if settings.features.final_answer:
             data = input_variables.model_dump()
             data["variable_summary"] = input_variables.variables_manager.get_variables_summary(last_n=2)
-            data["instructions"] = instructions_manager.get_instructions(self.name)
+            # [final_answer].instructions (settings/env) wins over the instruction-set file.
+            data["instructions"] = (
+                resolve_final_answer_instructions() or instructions_manager.get_instructions(self.name)
+            )
```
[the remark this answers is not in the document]
segevshlomovIBM: This is a deliberate, documented scope decision rather than a gap to close here: CugaLite composes its answer inside the same call_model that plans and calls tools, so its only guidance channel *is* the system prompt — which is exactly what the existing `special_instructions` field provides, and what the SDK `final_answer="..."` (str) form injects into automatically (see the sdk.py dispatch in this PR). Wiring `[final_answer].instructions` into the lite prompt as well would create two overlapping prompt-injection channels with an undefined precedence between them. The split is documented in the README section ("in the default CugaLite mode … use special_instructions") and in #621. Deterministic answer shapes — the case where ignored instructions would actually hurt — belong in `[final_answer].function`, which does apply on every CugaLite terminal path.

src/cuga/sdk.py
```diff
@@ -1867,6 +1875,46 @@ def __init__(
...
+                    f"(e.g. {final_answer.__name__}(...)) or a plain (str) -> str function"
+                )
+            if isinstance(final_answer, str):
+                fa_instructions = final_answer
+            elif isinstance(final_answer, FinalAnswerConfig):
+                if isinstance(final_answer.function, type):
+                    raise TypeError(
+                        f"FinalAnswerConfig.function got the class {final_answer.function.__name__} — "
+                        "pass a (str) -> str function, not a class"
+                    )
```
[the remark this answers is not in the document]
segevshlomovIBM: Fixed in a2820865: non-None, non-callable `FinalAnswerConfig.function` values now raise a TypeError at `CugaAgent` construction, with a regression test (`FinalAnswerConfig(function=123)`). Branch also updated with main.

sami-marreed:
Code review findings for `b56ba3e..a282086`:

1. **Important — malformed environment values can prevent startup**

   `src/cuga/config.py:145-146` adds strict string validators that run when `base_settings` is first accessed, outside the later error handling. Numeric or malformed `DYNACONF_FINAL_ANSWER__*` values raise during `import cuga.config`, contradicting the “never break delivery” contract.

   Suggested fix: make these best-effort settings non-fatal by removing the strict validators and relying on guarded consumers, or catch validation for both settings objects and normalize invalid values/parent shapes. Add subprocess import tests with malformed environment overrides.

2. **Important — a valid empty formatter result is discarded and may invoke the formatter twice**

   `src/cuga/sdk.py:2893-2916` treats `""` as if finalization never occurred, replaces it with transcript text, and applies the formatter again. This loses a valid `(str) -> str` result and violates the once-only guarantee.

   Suggested fix: track explicitly whether finalization occurred instead of using answer truthiness. Add an SDK regression with a formatter returning `""` and transcript text that differs from the finalized input.

3. **Important — the HITL default continuation can apply the formatter twice**

   `src/cuga/backend/cuga_graph/nodes/answer/final_answer.py:45-50` calls `finalize_answer` after `_generate_final_answer` has already finalized the answer before routing to `SUGGEST_HUMAN_ACTIONS`. A non-idempotent formatter is therefore invoked twice on the default HITL continuation, despite idempotence being only recommended and the contract promising one application.

   Suggested fix: make this continuation citation-only or persist/check an explicit finalized flag. Add a regression using a non-idempotent formatter.

No critical findings. The main architecture and normal terminal paths are solid, but these three issues should be fixed before merge.

segevshlomovIBM:
@sami-marreed thanks — all three were real, and all three are fixed in `a2820865..a56a7035`:

1. **Import crash on malformed env — confirmed and fixed.** Reproduced first: `DYNACONF_FINAL_ANSWER__FUNCTION=1 python -c "import cuga.config"` raised `ValidationError` exactly as you said (the strict validators were added on a CodeRabbit suggestion; the consumers were already guarded, so they were pure downside). Removed the `is_type_of` validators, kept the consumer guards, and added a subprocess import test with the malformed override (`test_malformed_env_does_not_break_import`).

2. **`""` treated as never-finalized — fixed with an explicit marker.** Every `FinalAnswerNode` terminal branch appends an `AIMessage` named `FinalAnswerAgent` to `state.messages`, so the fallback is now gated on `_finalization_ran(result)` instead of answer truthiness: a completed finalize that legitimately produced `""` is delivered as `""`, never replaced with transcript text, and the function is never re-applied. Marker detection is unit-tested; a full graph-level `""`-formatter run needs a live model, so that part is covered by the marker test plus the existing `test_empty_string_return_is_a_valid_result`.

3. **HITL double application — fixed.** The default continuation is now citation-resolution-only, same shape and reasoning as the supervisor forward branch (the answer it resumes was already finalized by `_generate_final_answer`). The now-dead `answer_function` carrier on `HumanInTheLoopHandler` is removed. Regression: `test_hitl_default_continuation_does_not_reapply_function` uses a configured formatter and asserts the resumed answer is untouched.

42 tests green across the final-answer suites, ruff clean, branch current with main.

src/cuga/sdk.py
```diff
@@ -1724,6 +1725,17 @@ async def generate_tool_guard_code(
...
 
+def _finalization_ran(result: dict) -> bool:
+    """Whether FinalAnswerNode finalized this run's answer.
+
+    Every terminal branch appends an AIMessage named FinalAnswerAgent to
+    state.messages. Used to gate invoke()'s empty-answer fallback: an empty
+    string from a completed finalize is a valid answer-function result and
+    must not be replaced (or the function re-applied — once-only contract).
+    """
+    return any(getattr(m, "name", None) == "FinalAnswerAgent" for m in result.get("messages", []) or [])
```
haroldship: This is too coarse for Sami's empty-string finding.

True if **any** checkpointed message is named `FinalAnswerAgent`. On the SDK path, a completed Lite turn always goes through `FinalAnswerNode`, which always appends that message — including when no formatter is configured and `final_answer` is `""`.

That (1) kills the pre-existing `invoke()` reasoning/chat fallback after every completed turn (defaults are no longer a no-op), and (2) false-positives on turn 2+ because `messages` accumulate.

Please skip the fallback only when a formatter is actually in play (`self._answer_function` or a non-empty settings path), or set an explicit per-turn `final_answer_finalized` flag instead of scanning historical names. `test_finalization_marker_detects_final_answer_agent_message` only tests this helper in isolation — needs an `invoke()`-shaped result with formatter-returns-`""` vs default recovery.

src/cuga/sdk.py
```diff
@@ -2836,8 +2900,12 @@ async def invoke(
         # Fallback: if final_answer is still empty, look at the last non-empty AI message.
         # Reasoning models sometimes return content='' with the answer only in
         # additional_kwargs['reasoning_content'], so check both fields.
+        # Gated on finalization NOT having run (every FinalAnswerNode terminal
+        # branch appends an AIMessage named FinalAnswerAgent): an empty string
+        # from a completed finalize is a valid answer-function result, and the
+        # fallback re-applying the function would break the once-only contract.
         fallback_sources = None
-        if not final_answer:
+        if not final_answer and not _finalization_ran(result):
```
haroldship: This is where the coarse gate actually changes default behavior.

After a completed Lite turn `_finalization_ran` is always true, so this block never runs — even with no `final_answer=` configured. In-graph recovery in `shared_nodes` covers the common reasoning-model case, but this was the documented SDK fallback and it is now dead on the success path.

If you keep a marker, combine it with "formatter configured", and add a regression that empty `final_answer` without a formatter still recovers from `chat_messages` / `reasoning_content`.

tests/unit/test_final_answer_function.py
```diff
@@ -0,0 +1,291 @@
...
+        timeout=120,
+    )
+    assert proc.returncode == 0, proc.stderr[-500:]
+    assert "OK" in proc.stdout
+
+
+# --- SDK surface --------------------------------------------------------------
+
+
+def test_sdk_rejects_class_passed_as_final_answer():
```
haroldship, line 308: These only lock constructor *rejections*. Please also assert:

- `CugaAgent(final_answer="…")` appends the `## Final answer instructions` block to `_special_instructions`
- `CugaAgent(final_answer=fn)` stores `fn` and `_create_hitl_wrapper_graph` passes it into `FinalAnswerNode`
- `CugaAgent(final_answer=FinalAnswerConfig(instructions=…, function=fn))` does both
- a result with `final_answer=""` + `FinalAnswerAgent` message keeps `""` and does **not** re-apply the function when a formatter is configured; without a formatter, empty still recovers

Those are the wiring bugs this PR has already had to fix once.

src/cuga/sdk.py
```diff
@@ -1834,6 +1847,12 @@ def __init__(
             shortlister: How to shrink a large tool set before the model sees it, e.g.
                 `Shortlister(strategy="hybrid")`. None = use `[shortlister]` settings
                 (default `"llm"`, i.e. unchanged behavior). Overridable per invoke()/stream().
+            final_answer: How the final answer is shaped. A `str` adds guidance for the
+                answer-composing LLM; a callable `(str) -> str` is applied deterministically
+                to the delivered answer instead (must be pure; applied once per delivered
```
haroldship: Nit: "instead" reads as if the callable *replaces* LLM composition. It does not — CugaLite (or the composer) still produces the answer, then the function post-processes it. Issue #621's `llm=False` knob was dropped; this docstring still describes that older shape.

haroldship (dismissed):
Thanks — the seam, failure isolation, class rejection, and the HITL citation-only path are in good shape. I went through Sami's three findings and CodeRabbit's resolved threads on current head (`a56a703`).

## Prior findings

**Sami**
1. Import crash on malformed `DYNACONF_FINAL_ANSWER__*` — **fixed.** Do not restore `is_type_of`; consumers already guard non-str. `test_malformed_env_does_not_break_import` locks it.
2. `""` treated as never-finalized — **only partially fixed.** See below.
3. HITL default continuation double-applied the formatter — **fixed.** Citation-only continuation + `test_hitl_default_continuation_does_not_reapply_function` are correct.

**CodeRabbit** — I would not reopen any of the five threads. Class rejection, fixture isolation, and `FinalAnswerConfig(function=123)` are actually fixed. CugaLite ignoring TOML `instructions` was correctly withdrawn (two prompt channels would be worse). The `is_type_of` suggestion is superseded by Sami #1.

## Requested changes

`_finalization_ran` gates on "FinalAnswerNode ran", not "a formatter produced this empty string". On the SDK path the node always appends `FinalAnswerAgent`, so:

- the pre-existing empty-answer fallback in `invoke()` no longer runs after a completed Lite turn (defaults are no longer a no-op)
- after turn 1 the marker stays true for the rest of the thread, even if this turn never finalized

Please gate the skip on a formatter actually being configured (or an explicit per-turn flag), and add an `invoke()`-level test for formatter-returns-`""` vs default recovery. Same truthiness hole on the HITL resume extract path (`sdk.py` around the `message is None` branch) — that path never got the gate.

SDK tests currently only cover constructor rejections. Please also assert the three happy-path forms (`str` / callable / `FinalAnswerConfig`) and that the callable is passed into `FinalAnswerNode`.

## Nits (non-blocking)

- `__init__` docstring says the callable runs "instead" of the LLM; it post-processes.
- README table still makes TOML `instructions` look universal in default Lite; one extra "Lite: ignored; use `special_instructions` / SDK `str`" note would close CodeRabbit's merge-risk without reopening the wiring request.
- `settings.toml` oversells idempotency (code/README: recommended).
- `DYNACONF_FINAL_ANSWER__FUNCTION` is load-**and-call** with the model answer as the argument. Same `get_class` loader as shortlister, new invoke shape — worth a sentence that this executes caller code, not just imports it. Operator-scoped, not a blocker.
- AppWorld SDK re-extract overwrites `InvokeResult.answer` and drops in-graph formatting; one sentence for eval users would help.

No remaining double-apply on the documented terminal paths. Lazy `FinalAnswerConfig` export is correct.

segevshlomovIBM:
@haroldship all requested changes are in (`a56a7035..7c1b2047`), and you were right that the marker alone was wrong — it broke the default no-op, which is the one property this PR most promised.

**The gate** — reworked exactly as you suggested: `_empty_answer_is_final(result, formatter_configured)` — true only when a formatter is in play (`self._answer_function` or a non-empty settings path via `answer_function_configured()`) AND finalization ran. With no formatter configured the pre-existing recovery runs unconditionally, so defaults are a no-op again. The same gate now covers the HITL resume extract path (both the error substitution and the interrupted-again check), which as you noted never had one.

**Tests** — `test_empty_answer_gate_requires_formatter_and_finalization` covers the three invoke()-shaped scenarios (no-formatter recovery preserved / formatter+finalized keeps `""` / formatter+unfinalized recovers); `test_formatter_in_play_reflects_sdk_and_settings` covers the predicate inputs; `test_sdk_happy_paths_wire_each_form` locks the three constructor forms including str-composes-with-existing-`special_instructions`. On "callable is passed into FinalAnswerNode": the constructor-storage assert plus the existing `test_injected_callable_applies_and_wins_over_settings` (which proves an injected callable reaches the seam) cover the wiring without mocking the whole graph build — happy to add a full graph-construction test if you want it anyway.

**Nits** — all taken: docstring says post-processes (the `llm=False` shape is indeed gone); README instructions row now says "ignored in default Lite mode" inline; settings.toml says idempotent-recommended and carries a SECURITY note that the dotted path is imported **and called**; the AppWorld scope note now states the SDK re-extract overwrites `InvokeResult.answer`.

One residual worth naming: with a formatter configured, a turn-1 `""` result followed by a turn-2 interrupt could still skip the pause message (the marker is thread-scoped). That corner needs the per-turn flag you mentioned — an `AgentState` field — which I'd rather do as a follow-up than grow this PR further; say the word if you want it here.

Branch is current with main; 45 tests green across the final-answer suites.

src/cuga/sdk.py
segevshlomovIBM: Fixed in two stages: first the gate was combined with formatter-configured (`_empty_answer_is_final`, `7c1b2047`), and now the coarse message-scan is gone entirely — `17c49da3` adds a per-invocation `AgentState.final_answer_finalized` flag (the explicit per-turn flag you suggested), set by every FinalAnswerNode terminal branch and reset by each turn's fresh input state. `test_finalization_flag_is_per_turn_not_historical` locks the turn-2+ case; `test_empty_answer_gate_requires_formatter_and_finalization` covers the three invoke()-shaped scenarios.

src/cuga/sdk.py
segevshlomovIBM: Fixed: with no formatter configured the recovery block runs unconditionally again (`_empty_answer_is_final` returns False), so defaults are a no-op — and with the per-turn flag (`17c49da3`) the skip can only trigger on a turn that actually finalized with a formatter in play. Regression: `test_empty_answer_gate_requires_formatter_and_finalization` (no-formatter recovery preserved).

tests/unit/test_final_answer_function.py
segevshlomovIBM: Added `test_sdk_happy_paths_wire_each_form`: str appends the '## Final answer instructions' block, callable stored on `_answer_function`, `FinalAnswerConfig` does both, and str composes with pre-existing `special_instructions`. The callable-reaches-the-node wiring is covered by the constructor-storage assert plus `test_injected_callable_applies_and_wins_over_settings` (injected callable reaching the seam); the formatter-returns-"" invoke-shape is in the gate tests.

src/cuga/sdk.py
segevshlomovIBM: Fixed — the docstring now says the callable post-processes the composed answer; the dropped llm=False shape is no longer described.

src/cuga/sdk.py
```diff
@@ -1724,6 +1725,29 @@ async def generate_tool_guard_code(
...
+def _empty_answer_is_final(result: dict, formatter_configured: bool) -> bool:
+    """Whether an empty final_answer is a deliberate answer-function result.
+
+    True only when a formatter is in play AND finalization ran — then ""
+    must be delivered as-is (no transcript recovery, no error substitution,
+    no re-application; once-only contract). With no formatter configured
+    this is always False, so the pre-existing empty-answer recovery behaves
+    exactly as before this feature (review: haroldship on #707).
+    """
+    return formatter_configured and _finalization_ran(result)
```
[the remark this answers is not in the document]
segevshlomovIBM: Fixed in `17c49da3` — exactly this: per-invocation `AgentState.final_answer_finalized`, reset by each turn's fresh input state, set only by FinalAnswerNode terminal branches; `_finalization_ran` reads the flag instead of scanning history. Regression with a historical FinalAnswerAgent message + unfinalized current turn: `test_finalization_flag_is_per_turn_not_historical`.

src/cuga/sdk.py
```diff
@@ -1724,6 +1725,29 @@ async def generate_tool_guard_code(
...
 
+def _finalization_ran(result: dict) -> bool:
+    """Whether a FinalAnswerNode terminal branch delivered THIS turn's answer.
+
+    Reads AgentState.final_answer_finalized — a per-invocation flag set by
+    every FinalAnswerNode terminal branch and reset by each turn's fresh
+    input state, so historical turns can never false-positive (review:
+    haroldship + coderabbitai on #707).
+    """
+    return bool(result.get("final_answer_finalized", False))
```
sami-marreed, line 1737: This docstring is not true today.

> reset by each turn's fresh input state, so historical turns can never false-positive

New threads default to `False`. Turn 2+ on an existing `thread_id` reuses the checkpoint dump and keeps `True`. Harold asked for a per-turn flag specifically to avoid that; the field exists, the reset does not.

The write belongs on the new-user-message path in `invoke()` (the `existing_state.model_dump()` block — not in the PR diff). Same pattern as `_reset_policy_decisions`. Do **not** reset on HITL resume (`message is None` / `action_response`):

```python
initial_state_dict["final_answer_finalized"] = False
```

I reproduced this: dump a turn-1 `AgentState` with `final_answer_finalized=True`, reconstruct as turn 2, and the flag stays `True`. `_empty_answer_is_final` then skips pause / error / transcript recovery when a formatter is configured.

src/cuga/backend/cuga_graph/state/agent_state.py
```diff
@@ -1032,6 +1032,11 @@ class AgentState(BaseModel):
     last_planner_answer: Optional[str] = None
     last_question: Optional[str] = None
     final_answer: Optional[str] = ""
+    # Per-invocation: True once a FinalAnswerNode terminal branch delivered
+    # this turn's answer. Reset by the fresh input state each turn; read by
+    # the SDK's empty-answer recovery gate (a formatter's "" is deliberate
+    # only when this turn actually finalized).
+    final_answer_finalized: bool = False
```
sami-marreed: Same claim as `_finalization_ran`: “Reset by the fresh input state each turn.”

`invoke()` does not reset this when copying `existing_state`. Until that write happens, this comment will keep the next reviewer believing the gate is per-turn.

Please reset it on a new user message (not HITL resume), then this comment is accurate.

tests/unit/test_final_answer_function.py
```diff
@@ -0,0 +1,352 @@
...
+    # The gate reads the per-invocation flag, never historical messages: a
+    # FinalAnswerAgent message from an earlier turn must not mark the current
+    # unfinalized turn as finalized (review: haroldship + coderabbitai).
+    from cuga.sdk import _finalization_ran
+
+    historical_msg = SimpleNamespace(name="FinalAnswerAgent")
+    assert _finalization_ran({"final_answer_finalized": True}) is True
+    assert _finalization_ran({"final_answer_finalized": False, "messages": [historical_msg]}) is False
+    assert _finalization_ran({"messages": [historical_msg]}) is False
+    assert _finalization_ran({}) is False
```
sami-marreed, line 247: This only asserts that `_finalization_ran` reads the dict field and ignores historical *messages*. It does not cover the actual leak: `invoke()` copying checkpointed `AgentState` via `model_dump()`.

Please add a regression that reconstructs turn-2 state from a dumped turn-1 checkpoint with `final_answer_finalized=True` and asserts `_empty_answer_is_final(..., formatter_configured=True)` is **False** after the new-turn reset (or, today, documents that it is still True until the reset lands).

src/cuga/sdk.py
```diff
@@ -2854,8 +2944,18 @@ async def invoke(
...
+            # finalize_answer here: answer function first, then citation
+            # resolution, before returning it to the caller.
+            if final_answer:
+                from types import SimpleNamespace as _NS
+
+                from cuga.backend.cuga_graph.nodes.answer.answer_function import apply_answer_function
+
+                _shim = _NS(final_answer=final_answer)
+                apply_answer_function(_shim, self._answer_function)
+                final_answer = _shim.final_answer
```
sami-marreed: Nit, non-blocking.

This path still does not match `finalize_answer` ordering. `finalize_answer` strips Harmony before `apply_answer_function`. This shim does not. A fallback transcript with `<|return|>` therefore gives the formatter different input than the terminal path.

I confirmed: a probe received `The total is 42<|return|>`. Strip before constructing `_shim` (and add a fallback test). Not a merge blocker if the flag reset lands; it is the remaining CodeRabbit consistency gap.

sami-marreed (changes requested):
Re-reviewed current head (`a68f986`) after the earlier rounds (my three findings, Harold’s follow-up, the per-turn flag).

The seam, failure isolation, class rejection, HITL citation-only path, empty-string formatter result, and malformed-env import are in good shape. Defaults are a no-op when no formatter is configured. I also ran the existing 28-test suite plus extra edge cases (whitespace settings, `"0"`, Unicode, non-str returns, large input, `FinalAnswerConfig()` no-op). Those hold.

## Prior findings

1. Import crash on malformed `DYNACONF_FINAL_ANSWER__*` — **fixed.**
2. `""` treated as never-finalized — **fixed on a single turn.** The gate is still wrong across turns (below).
3. HITL default continuation double-apply — **fixed.**
4. Harold: marker too coarse / defaults no longer a no-op — **fixed** via `_empty_answer_is_final(formatter_configured and finalized)`.
5. Harold / CodeRabbit: per-turn flag instead of scanning `FinalAnswerAgent` messages — **landed as a field, but it is not actually per-turn.**

## Requested change

`final_answer_finalized` is still thread-scoped. `invoke()` copies checkpointed state with `existing_state.model_dump()` and only resets `cuga_lite_metadata`. It never sets `final_answer_finalized = False`.

Reproduced:

```python
turn1 = AgentState(input="q1", url="", final_answer="", final_answer_finalized=True)
dumped = turn1.model_dump()
dumped["input"] = "q2"
turn2 = AgentState(**dumped)
assert turn2.final_answer_finalized is False  # fails; stays True
```

With a formatter configured, a turn-2 interrupt / error / empty answer before `FinalAnswerNode` then hits `_empty_answer_is_final` from **turn 1**, so `invoke()` skips the pause message, error substitution, and transcript recovery.

Please reset the flag on a **new user message** (same pattern as `_reset_policy_decisions`). Do **not** reset it on the HITL resume path (`message is None` / `action_response`). Add an `invoke()`-shaped regression: leftover `True` from turn 1 + current unfinalized result must not skip recovery.

## Non-blocking

The `invoke()` fallback shim still applies the function without stripping Harmony tokens first. `finalize_answer` does strip first. CodeRabbit already flagged this.

## Not blocking

- Lite ignoring TOML `instructions` — keep as documented.
- Do not restore `is_type_of` validators.
- Branch is 1 commit behind `main`; rebase after the flag reset.

Happy to approve once the flag is actually reset on new turns and the regression is in.

src/cuga/sdk.py
segevshlomovIBM: Fixed in 19b7ca66 — you were right, and your repro was exact: the reset only existed for fresh threads. invoke()'s new-user-message path (the `existing_state.model_dump()` block) now sets `initial_state_dict["final_answer_finalized"] = False`, same pattern and placement as `_reset_policy_decisions`, with a comment noting HITL resume deliberately bypasses it. Docstring corrected to describe the real mechanism.

src/cuga/backend/cuga_graph/state/agent_state.py
segevshlomovIBM: Fixed — the comment now says the field is per-turn *because invoke() explicitly resets it on each new user message* (HITL resume keeps the interrupted turn's value), not because fresh input state does.

tests/unit/test_final_answer_function.py
segevshlomovIBM: Added `test_checkpointed_flag_does_not_leak_into_next_turn`: reconstructs turn-2 state from a dumped turn-1 checkpoint with `final_answer_finalized=True`, asserts the raw reconstruction keeps True (the leak), then applies invoke()'s reset line and asserts `_empty_answer_is_final(..., formatter_configured=True)` is False.

src/cuga/sdk.py
segevshlomovIBM: Taken as well (it was cheap and it's the last consistency gap): the shim is extracted as `_apply_answer_function_to_text`, which strips harmony tokens before applying the function — same input shape as `finalize_answer`. `test_fallback_shim_strips_harmony_before_function` uses your probe scenario (`The total is 42<|return|>`).

sami-marreed (approved):
Thanks @segevshlomovIBM — this is good to land.

Verified on current head: `invoke()` now resets `final_answer_finalized` on each new user message (HITL resume left alone), the docs match that, the checkpoint/`invoke()` regressions cover the leak, and the fallback shim strips Harmony before the function. Defaults stay a no-op.

Approving.

Issue #712, opened 2026-08-25
[Bug]: knowledge ingest OOMs a 4GB pod on five 100KB uploads — embedding_batch_size drives O(batch x seq^2) memory


dolev31:
Filed without a parent: self-contained memory/tuning defect with a measured repro.

## What happened

A knowledge agent deployed on a bare-metal OpenShift cluster (image built from `Dockerfile.ubi`, pod memory limit **4 GB**) was **OOM-killed** while uploading five markdown files of ~100 KB each. Raising the limit to 6 GB let it complete, but ingest was slow.

## Root cause

`embedding_batch_size` defaults to **128** in the `standard` and `balanced` profiles (256 in `speed`). Embedding allocates ONNX intermediate tensors that scale as **batch × sequence²** — at batch 128 with 512-token chunks the attention tensors alone are ~1.6 GB. ONNX Runtime's arena allocator does not return that memory to the OS, so RSS ratchets up and stays up. With `max_ingest_workers = 2`, two concurrent ingests each carry a batch.

## Measured reproduction

Five files × 104 KB, `storage.mode=prod` (Postgres + pgvector), 10-core host:

```
RSS baseline    1,434 MB
RSS peak        4,642 MB   <- exceeds a 4 GB limit
RSS final       4,382 MB   <- never released
```

Sequentially it ratchets and plateaus; idle barely recovers it:

```
baseline 1,496 -> file1 3,270 (+1,773) -> file2 3,873 (+603)
-> file4 4,243 (+369) -> after 60s idle 4,094 (-149)
```

520 KB of input produced +3.2 GB of resident memory.

## Memory scales cleanly with the knob

Same five files, only `embedding_batch_size` changed:

| `embedding_batch_size` | peak RSS | wall | fits 4 GB? |
|---|---|---|---|
| **128** (current default) | 4,642 MB | 37.4s | ✗ **OOM** |
| 64 | 3,258 MB | 43.2s | ✓ tight |
| **32** | **2,452 MB** | **40.3s** | ✓ 1.6 GB spare |
| 16 | 1,720 MB | 89.1s | ✓ but 2.4× slower |

**32 looks like the knee**: 47% less peak memory for ~8% more wall time. 16 falls off a cliff because the adapter then makes 8× more embed calls, each with a metadata progress write.

## Things ruled out by measurement

[...]

PR #713, opened 2026-08-25
fix(knowledge): let an explicit env override beat the RAG profile

### LINKED ISSUE #711: [Bug]: knowledge CLI flags and DYNACONF env overrides are silently ignored when a rag_profile is active
Filed without a parent: small self-contained bug (config precedence).

## What happened

Every `cuga start` flag documented as "Sets `DYNACONF_KNOWLEDGE__…`" is **silently ignored** whenever a RAG profile is active — which is always, since `rag_profile` defaults to `standard`.

`KnowledgeConfig.from_settings` resolves these keys profile-first:

```python
embedding_batch_size=profile_embeddings.get("batch_size", embeddings.get("batch_size", 64)),
```

The env var *does* reach settings — dynaconf merges it correctly — but the profile value wins and the override is discarded with no warning.

Reproduced on `main` (`b56ba3ef`):

```
DYNACONF_KNOWLEDGE__EMBEDDINGS__BATCH_SIZE=16
  settings[knowledge][embeddings][batch_size] = 16    <- env landed
  KnowledgeConfig.embedding_batch_size        = 128   <- profile won
```

## Affected flags

Seven documented knobs, all dead when a profile is active:

| `cuga start` flag | env var | resolved from |
|---|---|---|
| `--embeddings-batch-size` | `…EMBEDDINGS__BATCH_SIZE` | `profile_embeddings` |
| `--embeddings-concurrency` | `…EMBEDDINGS__CONCURRENCY` | `profile_embeddings` |
| `--vector-insert-batch-size` | `…ENGINE__VECTOR_INSERT_BATCH_SIZE` | `profile_engine` |
| `--chunk-size` | `…CHUNKING__CHUNK_SIZE` | `profile_chunking` |
| `--chunk-overlap` | `…CHUNKING__CHUNK_OVERLAP` | `profile_chunking` |
| `--docling-pdf-mode` | `…DOCLING__PDF_MODE` | `profile_docling` |
| `--docling-layout-engine` | `…DOCLING__LAYOUT_ENGINE` | `profile_docling` |

`max_ingest_workers` is the one that works, because it explicitly checks `os.environ.get("CUGA_MAX_INGEST_WORKERS")` before consulting the profile.

## Why it matters

This is not academic. A knowledge agent on a bare-metal OpenShift cluster was OOM-killed at a 4 GB pod limit while uploading five ~100 KB markdown files. The remediation is to lower `embedding_batch_size` — and **there was no working way to do it**: not `settings.toml`, not the env var, not the CLI flag. The only lever was editing profile TOML inside the image.

Profile-over-`settings.toml` precedence is intentional and documented (so that switching `rag_profile` actually changes behaviour). An environment variable is different in kind: it is an explicit operator action, not a default sitting in a config file.

## How to reproduce

```bash
DYNACONF_KNOWLEDGE__ENABLED=true \
DYNACONF_KNOWLEDGE__EMBEDDINGS__BATCH_SIZE=16 \
uv run python -c "
from cuga.config import settings
from cuga.backend.knowledge.config import KnowledgeConfig
print(KnowledgeConfig.from_settings(settings).embedding_batch_size)"
# prints 128, expected 16
```

## Expected

An explicitly-set env var wins over the profile, mirroring the existing `CUGA_MAX_INGEST_WORKERS` escape hatch. Absent an override, the profile stays authoritative over `settings.toml` — no behaviour change for existing deployments.

## Environment

- `main` @ `b56ba3ef`
- Any `rag_profile` (default `standard`)
- Observed 2026-08-25

## Checklist

dolev31:
## Bug fix

Fixes #711
Refs #712

### Summary

Every `cuga start` flag documented as "Sets `DYNACONF_KNOWLEDGE__…`" was silently ignored whenever a RAG profile was active — which is always, since `rag_profile` defaults to `standard`.

### Root cause

`KnowledgeConfig.from_settings` resolved these keys profile-first:

```python
embedding_batch_size=profile_embeddings.get("batch_size", embeddings.get("batch_size", 64)),
```

The env var *does* reach settings — dynaconf merges it correctly — but the profile value wins and the override is dropped with no warning:

```
DYNACONF_KNOWLEDGE__EMBEDDINGS__BATCH_SIZE=16
  settings[knowledge][embeddings][batch_size] = 16    <- env landed
  KnowledgeConfig.embedding_batch_size        = 128   <- profile won
```

Seven knobs were affected: `embeddings.batch_size`, `embeddings.concurrency`, `engine.vector_insert_batch_size`, `chunking.chunk_size`, `chunking.chunk_overlap`, `docling.pdf_mode`, `docling.layout_engine`.

`max_ingest_workers` is the one that already worked — it explicitly checks `os.environ.get("CUGA_MAX_INGEST_WORKERS")` before the profile. This PR generalizes that pattern.

### Why it matters

This blocked remediation for a real incident (#712): a knowledge agent on bare-metal OpenShift was OOM-killed at a 4 GB pod limit uploading five ~100 KB markdown files. The fix is to lower `embedding_batch_size` — and there was **no working way to do it**: not `settings.toml`, not the env var, not the CLI flag. The only lever was editing profile TOML inside the image.

### Solution

`_profile_key()` prefers an explicitly-set env var, then falls back to the existing profile → `settings.toml` → default chain.

Profile-over-`settings.toml` precedence is deliberate — it is what makes switching `rag_profile` actually change behaviour — and is **unchanged**. An environment variable is different in kind: an explicit operator action, not a default sitting in a config file. **With no override set, behaviour is identical to before.**

Two deliberate details:

- **Blank values are ignored**, not treated as a value. Empty env vars are common in k8s manifests (`value: ""` from an unset template), and blanking out a profile default would be a nasty surprise.
- **A malformed value fails loudly** at startup rather than silently falling back to the profile — a typo'd deployment knob should not look like it worked.

### Type of changes

### Testing

New `tests/unit/test_knowledge_profile_env_override.py` — 21 tests, parameterized across all seven knobs:

- env override beats the profile (×7)
- **profile still wins when no override is set** (×7) — the regression guard for the intentional precedence
- profile omitting a key still falls through to `settings.toml`; default when neither has it
- blank / whitespace-only env is ignored (×3)
- malformed value raises rather than silently degrading
- end-to-end `env → KnowledgeConfig.from_settings`, not just the helper

Before/after on `main` for the end-to-end case: `128` (override discarded) → `16` (override honoured).

- `uv run pytest tests/unit/test_knowledge_profile_env_override.py` — 21 passed
- `uv run pytest tests/unit/ -k knowledge` — 634 passed, no regressions

### Not in this PR

The **default** `embedding_batch_size` of 128 still OOMs a 4 GB pod out of the box — see #712, which carries the measurements (128 → 4,642 MB peak; 32 → 2,452 MB peak for ~8% more wall time). Changing profile defaults is a performance trade-off that deserves its own review rather than riding along here. This PR gives operators a working lever today.

## Summary by CodeRabbit

- **Bug Fixes**
  - Environment-based knowledge configuration overrides now take effect even when a profile is active.
  - Configuration values correctly follow the precedence of explicit environment overrides, profile settings, general settings, and defaults.
  - Blank overrides are ignored, while invalid values now produce a clear validation error.

- **Tests**
  - Added coverage for configuration precedence, fallback behavior, invalid values, and profile-based embedding settings.


dolev31:
@sami-marreed green and ready — 29 checks, 0 failures.

Short version: every `cuga start` flag documented as "Sets DYNACONF_KNOWLEDGE__…" is **silently dead** whenever a RAG profile is active, which is always. The env var reaches settings correctly; `from_settings` then resolves profile-first and drops it with no warning. Seven knobs affected.

This is not theoretical — it blocked remediation for the OOM in #712. Lowering `embedding_batch_size` was impossible via `settings.toml`, the env var, **or** the CLI flag. The only lever was editing profile TOML inside the image.

What to check in review: **profile-over-settings.toml precedence is unchanged** — that is intentional design and 7 of the 21 tests exist specifically to guard it. Only an explicitly-set env var now wins, mirroring the `CUGA_MAX_INGEST_WORKERS` escape hatch that already existed. With no override set, behaviour is byte-identical to before.

Two deliberate details: blank env values are ignored rather than treated as a value (empty vars are common in k8s manifests), and a malformed value fails loudly at startup instead of silently falling back to the profile.

sami-marreed (approved):
Reviewed against #711. No critical findings.

The seven documented knobs now honor an explicit env/CLI override; profile-over-settings.toml is unchanged when no override is set. `os.environ` at `from_settings` time is the correct mechanism given Dynaconf's import-time load and the CLI's late env writes.

Do not add `validate()` inside `from_settings` (CodeRabbit). `KnowledgeEngine.__init__` already validates; the server `from_settings` swallow would turn a bad override into silent `enabled=False`.

Non-blocking: four other CLI-mapped profile keys remain profile-first (`--embeddings-model`, drop-page-chrome, junk-filter, hybrid-mode). Follow-up, not a blocker for the OOM lever this PR is for.

LGTM.

PR #714, opened 2026-08-25
fix(knowledge): lower embedding batch defaults so ingest fits a 4GB pod

### LINKED ISSUE #711: [Bug]: knowledge CLI flags and DYNACONF env overrides are silently ignored when a rag_profile is active

### LINKED ISSUE #712: [Bug]: knowledge ingest OOMs a 4GB pod on five 100KB uploads — embedding_batch_size drives O(batch x seq^2) memory

dolev31:
## Bug fix

Fixes #712

### Summary

A knowledge agent on a bare-metal OpenShift cluster was **OOM-killed at a 4 GB pod limit** uploading five ~100 KB markdown files. This lowers the profile embedding batch defaults so ingest fits.

### Root cause

A local ONNX embedder allocates intermediate tensors that scale as **batch × sequence²**, and onnxruntime's arena **does not return that memory to the OS**. So `embedding_batch_size` sets the process's resident *high-water mark*, not just its working set. With `max_ingest_workers = 2`, two concurrent ingests each carry a batch — peak tracks `batch_size × max_ingest_workers`.

Reproduced locally (5 × 104 KB, `storage.mode=prod`, 512-token chunks):

```
RSS baseline    1,434 MB
RSS peak        4,642 MB   <- exceeds a 4 GB limit
RSS final       4,382 MB   <- never released
```

520 KB of input produced +3.2 GB of resident memory.

### The measurements

Same five files, only `embedding_batch_size` changed:

| `embedding_batch_size` | peak RSS | wall | fits 4 GB? |
|---|---|---|---|
| **128** (old default) | 4,642 MB | 37.4s | ✗ **OOM** |
| 64 | 3,258 MB | 43.2s | ✓ tight |
| **32** | **2,452 MB** | **40.3s** | ✓ 1.6 GB spare |
| 16 | 1,720 MB | 89.1s | ✓ but 2.4× slower |

16 is past the knee — the adapter then makes 8× more embed calls, each with a metadata progress write.

### Changes

| profile | batch_size | workers | units before → after |
|---|---|---|---|
| `standard` | 128 → **32** | 2 | 256 → 64 |
| `balanced` | 128 → **32** | 2 | 256 → 64 |
| `max_quality` | 64 → **32** | 4 | 256 → 128 |
| `speed` | 256 → **64** | 2 | 512 → 128 |

`max_quality` moves down even though 64 looked moderate: it runs **4** ingest workers, so its concurrent exposure was the same 256 units as `standard` at 128. `speed` keeps the largest batch — that is the trade it exists to make — but is held to the same ceiling.

### Validated end to end

Re-ran the exact OOM scenario with the new defaults and no published-config override:

```
before (batch 128): peak 4,642 MB, wall 37.4s  -> OOM at 4GB
after  (batch  32): peak 2,610 MB, wall 38.4s  -> 1,486 MB headroom
```

**−44% peak memory for +2.7% wall time.**

### On the previous rationale

The old comment read *"larger = lower per-call overhead"*. That is true for **cloud** embedders, where a big batch amortizes HTTP round-trips. For a local ONNX model the trade inverts: large batches buy nothing measurable (I swept `batch_size`, `parallel` and `threads` — all landed within 44–51 ms/chunk) and cost memory quadratically in sequence length. The comments now say so, so the number is not re-raised by someone optimizing throughput without the memory context.

### Type of changes

### Testing

New `tests/unit/test_knowledge_profile_memory_budget.py` (9 tests) pins the budget so a future profile edit cannot quietly reintroduce the OOM:

- `batch_size × max_ingest_workers ≤ 128` for every profile — verified it **fails** with a clear message when `standard` is set back to 128
- a floor of 16 so nobody "fixes" memory by falling off the throughput knee
- `speed` stays ≥ `standard`, so the profiles remain ordered by intent

Also updates the pinned assertion in `test_knowledge_config_perf_keys.py`, which documents itself as requiring a deliberate update when the profile TOML changes.

- `uv run pytest tests/unit/ -k knowledge` — 622 passed
- `uv run ruff check` / `ruff format --check` — clean

### Relationship to #713

Independent — no overlapping files. #713 (fixes #711) makes the *override* work, so operators can tune this per-deployment. This makes the *default* safe. Either can merge first.

## Summary by CodeRabbit

- **Performance**
  - Reduced embedding batch sizes across all knowledge profiles to lower memory usage and help prevent out-of-memory failures.
  - Tuned profile settings to balance memory consumption and ingestion throughput for local embedding workloads.

- **Documentation**
  - Added guidance on memory and runtime tradeoffs across batch sizes and embedding environments.

- **Tests**
  - Added coverage enforcing memory budgets, minimum batch sizes, and performance relationships between profiles.


dolev31:
@sami-marreed :rotating_light: this one is a production incident — a knowledge agent got **OOM-killed at a 4 GB pod limit on bare-metal OpenShift** uploading five ~100 KB markdown files.

Root cause is that a local ONNX embedder allocates intermediate tensors scaling as **batch × sequence²**, and onnxruntime's arena never returns that memory — so `embedding_batch_size` sets the process's resident high-water mark, not its working set. Reproduced locally: peak 4,642 MB, and it never came back down.

Dropping the profile defaults takes the same scenario to **2,610 MB peak for +2.7% wall time**. That is measured end to end, not projected.

Worth your eye on two judgement calls:

1. **`max_quality` goes 64 → 32** even though 64 looks moderate. It runs **4** ingest workers, so its concurrent exposure was the same 256 units as `standard` at 128. The unit that matters is `batch_size × max_ingest_workers`, not batch size alone.
2. **`speed` keeps the largest batch (64)** deliberately — that is the trade the profile exists to make — but it is held to the same ceiling as everything else.

I also swept the embedder knobs before touching defaults, to be sure I was not trading memory for the throughput the profile was tuned for: `batch_size`, `parallel` and `threads` all landed within 44-51 ms/chunk. The old "larger = more throughput" comment is true for **cloud** embedders amortizing HTTP round-trips; for local ONNX it simply does not hold. Comments updated so the number is not re-raised later without the memory context.

Pairs with #713 (fixes #711), which makes the override actually work so this can be tuned per-deployment — that one is green already. Independent files, either order.

sami-marreed (approved):
Reviewed against #712. No critical findings.

`standard`/`balanced` batch 128 → 32 is the measured 4 GB OOM fix (peak ~4.6 GB → ~2.6 GB, ~3% more wall). `embedding_batch_size` is not in `vector_config_hash`, so this does not force a re-index. Budget tests load shipped TOML and sit in `tests/unit/` (unit-b).

Non-blocking caveats: `speed` at 64 is the measured "tight" ~3.3 GB row; the product cap allows `batch_size=128` with `workers=1`; published agent snapshots will not pick up the new default until PATCH/re-publish or the #713 env override.

LGTM.

PR #716, opened 2026-08-26
feat(llm): add wxo platform to LLMManager for watsonx Orchestrate routing

### LINKED ISSUE #715: [Feature]: watsonx Orchestrate (wxO) as an LLM platform — route CUGA's LLM calls through Orchestrate for entitlement billing
## What you want and why

CUGA's `LLMManager` (`src/cuga/backend/llm/models.py`) supports OpenAI, WatsonX, Azure, Groq, RITS, OpenRouter, MiniMax, Google GenAI, and LiteLLM as `platform` backends, but every one of them calls the provider directly with a provider-specific key. There's no path that routes LLM calls through a customer's **watsonx Orchestrate (wxO)** tenant instead.

That matters because going through wxO isn't just "another provider" — it changes what you get for free: calls become traced/observable in the tenant, pass through the tenant's guardrails, and are billed against the tenant's own entitlement rather than a separate provider credential. This was demoed internally (recording + transcript, June 2026): a LangGraph agent's `create_llm()` factory picking `ChatWxO(instance_url=..., api_key=...)` over `ChatGroq`/`ChatOpenAI` when wxO credentials are present, with the SDK handling JWT exchange/refresh transparently.

## How it could work

Add `platform = "wxo"` as a new `LLMManager` branch, following the existing per-provider pattern:

- `ChatWxO` (from the optional `ibm-watsonx-orchestrate-sdk` PyPI package — a `ChatOpenAI` subclass) loaded lazily, same pattern as the existing `_get_reasoning_chat_litellm` loader, so the SDK is never a hard dependency. The SDK needs Python >= 3.11, above CUGA's own >= 3.10 floor, so it has to be an optional extra (`cuga[wxo]`) gated by a `python_version` marker.
- `_get_model_name` / `_get_base_url` arms resolving `WXO_API_KEY` / `WXO_INSTANCE_URL` (env override → TOML → default), with the instance URL folded into the LLMManager cache key so two tenants never share a cached client.
- Local-dev auto-detection: an instance URL of `http://localhost:4321` (CUGA's existing local ADK dev-server port) shouldn't require an API key; only a remote instance should.
- A `settings.wxo.toml` covering all ten agent roles, defaulting to a provider-prefixed model id (e.g. `watsonx/openai/gpt-oss-120b`) matching the tenant's own `/models` catalog format.
- Docs (`.env.example`, `README.md`) and unit/integration test coverage mirroring an existing provider (`minimax` is the closest OpenAI-compatible template).

I have a working implementation of this locally, live-tested end-to-end against a real wxO tenant: `invoke`/`ainvoke`/`stream`/`astream`/`bind_tools`/`with_structured_output` all verified — including confirming (no separate Groq credential present, HTTP traffic only ever reaching the wxO gateway host, Groq-shaped response telemetry in the reply) that a Groq-hosted model routed through wxO genuinely executes on Groq infra while authenticating solely via the wxO key. Filing this issue first per the repo's contribution flow; happy to open the PR.

## Links or extra context

- `src/cuga/backend/llm/models.py` — existing per-provider `platform` branches to follow (`_get_model_name`, `_get_base_url`, `_create_llm_instance`).
- `src/cuga/configurations/models/` — existing `settings.<provider>.toml` files.
- Internal demo recording + transcript motivating this (June 2026), showing the `create_llm()` / `ChatWxO` pattern this issue proposes bringing into CUGA natively.
- Related but distinct: #77 (IBM watsonx Orchestrate external agent onboarding) is the *reverse* integration — CUGA being called *by* Orchestrate as an external agent — not this (CUGA calling *out* to Orchestrate for its own LLM).

laredo:
## Feature

Closes #715

### Summary

Adds `platform = "wxo"` to `LLMManager` so CUGA's LLM calls can be routed through a watsonx Orchestrate (wxO) tenant instead of calling a provider directly — calls become traced/observable in the tenant and are billed against its entitlement rather than a separate provider key (see #715 for full context, including live verification against a real tenant).

- `ChatWxO` (from the optional `ibm-watsonx-orchestrate-sdk` package, a `ChatOpenAI` subclass) is loaded lazily via a new `_get_chat_wxo()` helper, mirroring the existing `_get_reasoning_chat_litellm` pattern — the SDK needs Python >= 3.11, above CUGA's own >= 3.10 floor, so it ships as an optional `cuga[wxo]` extra, never a core dependency.
- `_get_model_name` / `_get_base_url` resolve `WXO_API_KEY` / `WXO_INSTANCE_URL` (env -> TOML -> default), with the instance URL folded into the LLMManager cache key so two tenants never share a cached client.
- A local ADK dev server (`http://localhost:4321`, auto-detected) runs keyless; only a remote instance requires an API key.
- New `settings.wxo.toml` covering all ten agent roles.
- `WXO_API_KEY` added to the secrets seed map and the `resolve_llm_api_key_ref` provider hint.
- `.env.example` and `README.md` updated with setup instructions.

### Testing
- New unit coverage: `tests/unit/test_llm_wxo_provider.py` (model-name/base-url resolution, client wiring, SSL, caching, context-profile lookup, the secrets hint)
- New integration coverage: `tests/integration/test_llm_wxo_config_publish.py` (Manage-UI publish path via `create_llm_from_config`)
- Full shared-path regression set re-run with zero regressions: `test_llm_advanced_params`, `test_llm_override`, `test_llm_http_timeout`, `test_llm_minimax_provider`, `test_llm_reasoning_model_temperature`, `test_llm_watsonx_loop_rebind`, `test_token_counter`, `test_llm_config_publish`
- Live-tested end-to-end against a real wxO tenant: `invoke`/`ainvoke`/`stream`/`astream`/`bind_tools`/`with_structured_output`, both the `watsonx/` and tenant-recommended `groq/`-hosted model variants, plus a bad-key case failing with a clear auth error

## Summary by CodeRabbit

* **New Features**
  * Added watsonx Orchestrate as a supported LLM platform.
  * Added local and hosted configuration options, including model, URL, API key, timeout, SSL, and sampling settings.
  * Added default agent model configurations and secure API key handling.
  * Added Python 3.11+ installation support for the watsonx Orchestrate SDK.

* **Documentation**
  * Added setup instructions, environment variables, configuration details, and model override guidance.

* **Tests**
  * Added coverage for configuration, authentication, local and remote endpoints, and provider behavior.

### DIFF
--- a/src/cuga/backend/llm/models.py
+++ b/src/cuga/backend/llm/models.py
@@ -20,6 +20,7 @@ if TYPE_CHECKING:
 from cuga.backend.cuga_graph.utils.token_counter import ensure_model_context_profile
 from cuga.backend.llm.load_test_mock import clone_load_test_mock_chat_model, is_mock_llm_enabled
 from cuga.backend.secrets import resolve_secret
+from cuga.backend.utils.consts import LOCAL_ORCHESTRATE_URL
 from cuga.config import DEFAULT_LLM_HTTP_TIMEOUT, settings
 
 # weakref.ref(loop) on watsonx APIClient for the loop the async httpx client is for.
@@ -117,6 +118,30 @@ def _get_reasoning_chat_litellm():
 
 _get_reasoning_chat_litellm._loaded = False  # type: ignore[attr-defined]
 _get_reasoning_chat_litellm._cls = None  # type: ignore[attr-defined]
+
+
+def _get_chat_wxo():
+    """Lazy-load and return the ChatWxO class, or None if ibm_watsonx_orchestrate_sdk is missing.
+
+    No lock is needed: same reasoning as ``_get_reasoning_chat_openai`` — all callers
+    are on the asyncio event loop with no OS-thread exposure to this path.
+    """
+    if _get_chat_wxo._loaded:
+        return _get_chat_wxo._cls
+
+    try:
+        from ibm_watsonx_orchestrate_sdk.langchain import ChatWxO
+
+        _get_chat_wxo._cls = ChatWxO
+    except ImportError:
+        _get_chat_wxo._cls = None
+
+    _get_chat_wxo._loaded = True
+    return _get_chat_wxo._cls
+
+
+_get_chat_wxo._loaded = False  # type: ignore[attr-defined]
+_get_chat_wxo._cls = None  # type: ignore[attr-defined]
 
 _ENV_REF_PATTERN = re.compile(r"^[A-Z][A-Z0-9_]*$")
 _DEFAULT_LLM_HTTP_TIMEOUT = DEFAULT_LLM_HTTP_TIMEOUT
@@ -647,6 +672,20 @@ class LLMManager:
                 default_model = "openai/gpt-oss-20b"
                 logger.info(f"No model_name specified, using default: {default_model}")
                 return default_model
+        elif platform == "wxo":
+            # For watsonx Orchestrate, check environment variables
+            env_model_name = os.environ.get('MODEL_NAME')
+            if env_model_name:
+                logger.info(f"Using MODEL_NAME from environment for wxO: {env_model_name}")
+                return env_model_name
+            elif toml_model_name:
+                logger.debug(f"Using model_name from TOML: {toml_model_name}")
+                return toml_model_name
+            else:
+                # Default fallback
+                default_model = "watsonx/openai/gpt-oss-120b"
+                logger.info(f"No model_name specified, using default: {default_model}")
+                return default_model
         elif platform == "watsonx":
             # For WatsonX, check environment variables
             env_model_name = os.environ.get('MODEL_NAME')
@@ -840,6 +879,25 @@ class LLMManager:
         # Groq SDK manages its own endpoint; any URL in config is ignored
         if platform == "groq":
             return None
+
+        # wxO: let WXO_INSTANCE_URL env override the TOML so the tenant instance can be
+        # swapped from .env without editing every per-role TOML block. Falls back to the
+        # local ADK dev server (same port CUGA already knows as LOCAL_ORCHESTRATE_URL).
+        if platform == "wxo":
+            env_instance_url = os.environ.get('WXO_INSTANCE_URL')
+            if env_instance_url and env_instance_url.strip():
+                logger.info(f"Using WXO_INSTANCE_URL from environment: {env_instance_url.strip()}")
+                return env_instance_url.strip()
+            toml_url = (
+                model_settings.get("instance_url")
+                or model_settings.get("base_url")
+                or model_settings.get("url")
+            )
+            if toml_url and str(toml_url).strip():
+                logger.debug(f"Using instance_url from TOML for wxO: {toml_url}")
+                return str(toml_url).strip()
+            logger.info(f"No instance_url specified for wxO, defaulting to local: {LOCAL_ORCHESTRATE_URL}")
+            return LOCAL_ORCHESTRATE_URL
 
         # RITS: let RITS_BASE_URL env override the TOML so model/endpoint can be
         # swapped from .env without editing every per-role TOML block.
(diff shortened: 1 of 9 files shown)


src/cuga/backend/secrets/seed.py
```diff
@@ -140,6 +141,7 @@ def resolve_llm_api_key_ref() -> str:
         "minimax": "minimax-api-key",
         "rits": "rits-api-key",
         "watsonx": "watsonx-api-key",
+        "wxo": "wxo-api-key",
```
sami-marreed, line 193: #c1 — Adding the `wxo` substring hint does not make credential resolution platform-aware. `resolve_llm_api_key_ref()` checks model-name hints first, so the default `watsonx/openai/...` wxO model selects `db://watsonx-api-key`, while the documented `groq/openai/...` override selects `db://openai-api-key`. With no `MODEL_NAME`, another ambient provider key can also win by seed-map order; that is the failure currently shown by the Unit CI shard. Please pass or derive the active platform, map `wxo` directly to `wxo-api-key` before interpreting the downstream model identifier, and cover competing provider keys in the tests.

src/cuga/backend/llm/models.py
```diff
@@ -1145,6 +1203,68 @@ def _create_llm_instance(self, model_settings: Dict[str, Any]):
...
+                api_key = _normalize_secret(resolve_secret(apikey_ref)) or os.environ.get(apikey_ref)
+            if not api_key:
+                api_key = _normalize_secret(resolve_secret("WXO_API_KEY")) or os.environ.get("WXO_API_KEY")
+
+            is_local = bool(base_url) and (
+                base_url.startswith("http://localhost")
+                or base_url.startswith("http://127.0.0.1")
+                or base_url.startswith("http://[::1]")
+                or base_url.startswith("http://0.0.0.0")
+            )
```
sami-marreed: #c2 — Prefix matching classifies remote hosts such as `http://localhost.example.com:4321` as local, and it accepts any port even though keyless operation is intended only for the local ADK server. Please parse the URL and require `http`, an exact approved loopback hostname, and port 4321 (or the configured canonical ADK port) before allowing a missing API key. Add negative tests for hostname suffixes and non-ADK ports.

tests/unit/test_llm_wxo_provider.py
```diff
@@ -0,0 +1,430 @@
...
+These tests lock: model-name/instance-url resolution, the client wiring in
+``_create_llm_instance`` (param names, key resolution, SSL, caching), the
+local-dev vs remote API-key requirement, and the ``resolve_llm_api_key_ref``
+secret hint.
+"""
+
+import os
+from unittest.mock import patch
+
+import pytest
```
sami-marreed, line 19: #c3 — Repository policy requires every new test to declare a registered pytest type marker. These tests are mocked and isolated, so please add module-level `pytestmark = pytest.mark.unit`.

tests/integration/test_llm_wxo_config_publish.py
```diff
@@ -0,0 +1,123 @@
...
+``provider``/``model`` are honored — in local mode ``create_llm_from_config``
+overrides both from ``settings.agent.code.model`` (see
+``tests/integration/test_llm_config_publish.py`` for that path).
+"""
+
+from contextlib import contextmanager
+from types import SimpleNamespace
+from unittest.mock import patch
+
+import pytest
```
sami-marreed, line 22: #c4 — This new module also needs a registered test-type marker. Although it lives under `tests/integration`, the provider and settings are fully mocked, so `pytestmark = pytest.mark.unit` is the applicable registered marker; `integration` is not registered in `pyproject.toml`.

tests/unit/test_llm_wxo_provider.py
```diff
@@ -0,0 +1,430 @@
...
+    FakeChatWxO.instances.clear()
+    for key in ("WXO_API_KEY", "WXO_INSTANCE_URL", "MODEL_NAME"):
+        os.environ.pop(key, None)
+    yield
+    mgr._models.clear()
+    mgr._pre_instantiated_model = None
+    set_current_llm_override(None)
+    FakeChatWxO.instances.clear()
+    for key in ("WXO_API_KEY", "WXO_INSTANCE_URL", "MODEL_NAME"):
+        os.environ.pop(key, None)
```
sami-marreed: #c5 — Direct `os.environ.pop()` calls permanently remove the developer or CI runner's original `WXO_API_KEY`, `WXO_INSTANCE_URL`, and `MODEL_NAME` values for all later tests in the same process. Please inject pytest's `monkeypatch` fixture and use `monkeypatch.delenv(..., raising=False)` during setup; pytest will restore the original environment automatically after the test.

sami-marreed (changes requested):
Thanks for the PR — the lazy optional dependency, tenant-aware cache key, TLS handling, role configuration, and documentation are integrated thoughtfully.

The main blocker is credential selection: the active wxO platform is not considered before model-name hints or ambient provider keys, and the required Unit CI shard is already failing on that behavior. I also found an endpoint-classification issue and two test-policy/isolation problems that should be cleaned up before merge.

**Tagged items**

- **#c1** — make secret resolution select the active wxO platform
- **#c2** — validate keyless local ADK URLs structurally
- **#c3** — add the required unit marker to the provider tests
- **#c4** — add the required unit marker to the mocked publish-path tests
- **#c5** — restore ambient environment variables after tests

Merge is blocked until these tagged items are addressed.

---

**When you respond, please tie your answers to each tag** so we can track what changed:

- **#c1** — make secret resolution select the active wxO platform
- **#c2** — validate keyless local ADK URLs structurally
- **#c3** — add the required unit marker to the provider tests
- **#c4** — add the required unit marker to the mocked publish-path tests
- **#c5** — restore ambient environment variables after tests

Reply in the inline threads where you can, or summarize in your next commit message or PR comment using the same tags (`#c1`, `#c2`, `#c3`, `#c4`, `#c5`). That will make re-review much faster.

laredo:
Thanks for the detailed review — pushed fixes for all five tagged items.

**#c1** — `resolve_llm_api_key_ref()` now checks the actively configured platform (`settings.agent.code.model.platform`, mirroring `create_llm_from_config`'s own local-mode resolution) via a new `platform -> secret-slug` map, and returns that *before* any `MODEL_NAME` hint-guessing or the ambient env scan. Confirmed this fixes both cases you found — the default `watsonx/...` id and the documented `groq/openai/...` override — plus reproduced the actual CI failure (an ambient `GROQ_API_KEY` shadowing `WXO_API_KEY` by dict order) in a new test and confirmed the fix resolves it. (`512d9261`)

**#c2** — added `_is_local_wxo_url()` using `urllib.parse.urlsplit` instead of string prefixes: exact hostname match against the loopback set, `scheme == "http"`, and the same port as `LOCAL_ORCHESTRATE_URL`. Added negative tests for `http://localhost.example.com:4321` (hostname suffix) and `http://localhost:9999` (wrong port) — both now correctly require a key. (`10f1fee2`)

**#c3** — `pytestmark = pytest.mark.unit` added to `tests/unit/test_llm_wxo_provider.py`. (`8508821e`)

**#c4** — `pytestmark = pytest.mark.unit` added to `tests/integration/test_llm_wxo_config_publish.py` — confirmed `integration` isn't a registered marker in `pyproject.toml` (`--strict-markers` would reject it outright), so `unit` is correct given both files are fully mocked. (`319d662d`)

**#c5** — both files' `reset_llm_state` fixtures now use `monkeypatch.delenv(..., raising=False)` instead of `os.environ.pop`, so pytest restores whatever was ambient before each test. The integration file had no env isolation at all previously (this also directly covers CodeRabbit's separate `WXO_INSTANCE_URL`-leak comment on that file). (`8508821e`, `319d662d`)

Full suite re-verified clean (`ruff check`, `ruff format`, and the full `tests/unit`/`tests/integration` run — only the same pre-existing, unrelated failures that are present on `main` without this branch: 3 in `test_sdk_citations.py` and 4 in `test_tool_call_tracking.py`, both needing live credentials/infra this environment doesn't have).

sami-marreed (approved):

PR #726, opened 2026-08-28
feat(cli): add service instance purge command

illeatmyhat:
## Summary

Adds an operator-facing CLI purge command for deprovisioned service instances:

```sh
cuga purge service-instance --service-instance-id <id>
```

The command scans the configured storage backend for tables with an `instance_id` column and deletes only rows matching the requested service instance. It supports both local SQLite and production Postgres storage, and includes `--dry-run` to count matching rows without deleting them.

This intentionally ships as CLI/reaper plumbing instead of an HTTP API, so it can be executed by a Kubernetes CronJob/operator using the same rebuilt CUGA image.

## Caveat: knowledge metadata

This purge only deletes records in tables that actually carry `instance_id`.

Knowledge vector and policy vector rows are covered because their schemas include `tenant_id` and `instance_id`. However, current knowledge metadata tables (`documents`, `tasks`, `collection_config`, `settings` / `cuga_knowledge_meta_*`) are keyed by `collection` and do not reliably encode the service instance. Local FTS sidecar tables are also derived/unscoped.

So this PR deliberately does **not** purge knowledge metadata yet. Safely doing that requires a follow-up schema migration to scope knowledge metadata by `tenant_id` + `instance_id`, or another deterministic ownership mapping. This matches the direction explored in closed PR #718, but keeps this PR small and avoids inferential deletes.

## Testing

- [x] `uv run pytest tests/unit/test_service_instance_cleanup.py -q`
- [x] `uv run ruff check tests/unit/test_service_instance_cleanup.py src/cuga/backend/storage/service_instance_cleanup.py src/cuga/cli/purge_cmds.py src/cuga/cli/main.py`

## Summary by CodeRabbit

* **New Features**
  * Added a `purge service-instance` command to remove stored data for a selected service instance.
  * Added a dry-run option to preview record counts without deleting data.
  * Cleanup supports configured SQLite and PostgreSQL databases, including separate knowledge and policy stores.
  * Results include per-table and aggregate counts in JSON format.
* **Bug Fixes**
  * Ensured cleanup affects only the selected service instance while preserving unrelated data.
  * Added validation for blank service-instance IDs.
  * Preserved user-provided demo server configuration when starting a demo.


src/cuga/backend/storage/service_instance_cleanup.py
```diff
@@ -0,0 +1,218 @@
...
+        deleted = await _delete_postgres_service_instance_records(
+            postgres_url,
+            service_instance_id,
+            dry_run=dry_run,
+        )
+    else:
+        if not local_db_path:
+            raise ValueError("storage.local_db_path is required when storage.mode=local")
+        deleted = await _delete_sqlite_service_instance_records(
+            local_db_path,
```
sami-marreed: **[P1] Purge the configured knowledge and policy databases**

This only opens the primary storage database, but normal local knowledge storage writes vectors to `knowledge.persist_dir/knowledge_vectors.db`. A successful purge therefore leaves the instance’s vector rows and chunk text intact. Configured `policy.policy_db_path` and separate knowledge PostgreSQL connections are also omitted. I reproduced this through `create_vector_store`: the primary row was deleted while the matching knowledge vector remained. Please resolve all configured storage targets and add coverage using the actual knowledge factory; the current test places vectors directly in the primary database.

src/cuga/backend/storage/service_instance_cleanup.py
```diff
@@ -0,0 +1,218 @@
...
+    finally:
+        await pool.close()
+
+
+async def delete_service_instance_records(
+    service_instance_id: str,
+    *,
+    dry_run: bool = False,
+) -> ServiceInstanceCleanupResult:
+    service_instance_id = (service_instance_id or "").strip()
```
sami-marreed: **[P2] Preserve the exact service-instance ID when deleting**

Stripping the ID changes the deletion target, while `get_service_instance_id()` returns configured values verbatim. With rows for both `"instance-1"` and `" instance-1 "`, requesting the latter deletes the former and leaves the requested row. Please use `.strip()` only to reject blank input and preserve the original value for matching. A regression test containing both IDs would cover this.

sami-marreed (changes requested):
Requesting changes for two reproduced issues: the purge omits configured vector databases, and trimming the service-instance ID can change the deletion target. The seven added tests and targeted Ruff check pass, but the current fixture does not exercise the normal separate knowledge database. Both findings were reproduced using temporary SQLite/sqlite-vec data; live PostgreSQL was not tested.

src/cuga/backend/storage/service_instance_cleanup.py
illeatmyhat: Fixed in 3cea4be4, with regression coverage in b25582ca after rebasing onto current main. Purge resolves the primary database, the configured knowledge vector database (including a separate PostgreSQL connection), and policy.policy_db_path, deduplicating shared targets. The local regression creates vectors through the real create_vector_store factory and verifies both dry-run and deletion across the separate stores. All 19 cleanup tests pass. Resolving this addressed conversation.

src/cuga/backend/storage/service_instance_cleanup.py
illeatmyhat: Fixed in 3cea4be4, with regression coverage in b25582ca after rebasing onto current main. strip() is used only to reject blank input; every deletion target and the returned identifier retain the exact original value. The regression stores both "instance-1" and " instance-1 " and verifies only the exact requested ID is selected/deleted, for dry-run and execution. All 19 cleanup tests pass. Resolving this addressed conversation.

sami-marreed (approved):
Both previously requested changes are addressed at b25582cac65daea0e8799a9b0abecdba2434bdfd. Purge includes configured knowledge and policy databases and preserves exact service-instance IDs. All 19 cleanup tests and targeted Ruff checks passed. PostgreSQL target routing was tested with mocks; no live PostgreSQL validation was performed.

PR #727, opened 2026-08-30
fix(agent): ground writes in observed state

### LINKED ISSUE #709: [Bug]: CugaLite mutates state from assumptions instead of observed state — no read-before-write / minimal-delta discipline (largest failure family in two models)
## What happened

CugaLite mutates world state from its **assumptions** about the target instead of the target's **observed state**, and does not confine writes to the requested delta. This is the single largest root-cause family in the AppWorld failure analysis, and — unlike most failure modes — it reproduces with **identical per-assert signatures across two very different models** (gpt-oss-120b and gemma-4-31B-it):

| task | signature (fails in BOTH models) | missing behavior |
|---|---|---|
| `b6d1104_1` | writes `May 2023: $424.00` into a note whose entries are `- 05/23 => $813.0` (amount computed correctly!) | read the artifact, match its conventions |
| `a3ba388_1` | `update_draft` resends body/subject, tripping "draft has not changed in any other field" | minimal delta — touch only requested fields |
| `d6ac34d_1` | copies yesterday's habit note but botches the title/pinned convention | observe existing structure before writing |
| `4815c06_1` | asked to order 3 named products, orders the entire pre-existing 6-item cart ($49.8k in one gpt-oss run) | enumerate the collection first; pre-existing content is not part of the request |
| `ffea2b5_1` | sends a fresh email instead of the existing draft reply (thread_id/"Re:" asserts fail) | act on the existing artifact, don't recreate it |
| `f6936d4_1` (gemma) / `9bf2c8a_1` (easy, gpt-oss) | quantities silently wrong: 3+3 ordered instead of 1+1 / qty 3 collapses to 1 on a cart→wishlist move | attributes travel with items |

Failure-mass numbers (from `cuga-eval/benchmarks/appworld/evaluation_bundles/TODOS_AB_FAILURE_DISTRIBUTION.md`, root-cause rollups): the grounding family is **43% of all gpt-oss-120b `test_med` failed task-runs** (74/172, largest family; 33% on easy) and **~9 of 16 gemma-4-31B-it med failures** (single-run census, Aug 24 bundle `20260824_025513_compare_gemma-4-31b-it-med`). Model-independence of the signatures indicates a harness-level behavioral gap, not a model quirk.

The executor prompt today has no write-discipline guidance: rule 2 covers *reusing fetched data* ("CHECK VARIABLES FIRST") and #704 added *post-error* obstacle handling, but nothing tells the agent what to observe **before a mutation** or how to bound the write.

## Proposed fix — one domain-neutral prompt rule

Add to the Critical Rules of `cuga_lite/prompts/mcp_prompt.jinja2` (adjacent to rule 2, which it composes with):

> **GROUND EVERY WRITE IN OBSERVED STATE, NOT ASSUMPTIONS:** Before any call that creates, updates, or deletes data, read the current state of the exact object or collection you are about to change (skip only if you fetched it this task and nothing has changed it since). Then:
> - **Minimal delta:** change only what the request requires; preserve every field, item, and attribute you were not asked to change. When updating or appending to an existing artifact (a note, draft, document, list), first observe its existing structure and formatting conventions and match them exactly.
> - **Scope:** when acting on a collection, enumerate its current contents and touch only the items the request refers to — pre-existing content is not part of the request.
> - When moving or re-creating items, carry their exact quantities and attributes unless the request changes them.

Safety posture (important after the #706 review discussion): this rule prescribes **restraint, not action** — extra read-only calls, fewer and smaller mutations, explicit preservation of untouched state. Its worst case is added read latency; there is no funding/consequential-action surface, so no authorization gate is needed. It is the kind of CRUD discipline a production agent should have regardless of benchmarks; nothing in it is AppWorld- or task-specific (artifact/collection are generic CRUD notions; the tasks above are evidence, not targets).

Out of scope (deliberately): read-side extraction/identification errors (wrong recipe picked, seller-rating aggregation) — those are #407-family claim-grounding, not write discipline.

## How to reproduce

1. Run AppWorld `test_med` via cuga-eval (gpt-oss-120b or gemma-4-31B-it, temp 0.1). The six tasks above fail with the listed signatures across models and runs (see the failure-distribution doc for per-arm counts).
2. Suggested A/B probe (10 tasks): `b6d1104_1 a3ba388_1 d6ac34d_1 4815c06_1 ffea2b5_1 f6936d4_1` (med) + `9bf2c8a_1 fd1f8fa_1` (easy, same family) + `325d6ec_1 8749218_1` (no-write-discipline controls). Judge on per-assert signatures (format string written, changed-fields set, order contents, quantities) rather than pass/fail alone; track tok/task to price the extra reads.

## Environment

- cuga-agent main `5de53ade` era; behavior identical across the #676 A/B arms (todos on/off, main and branch) and on gemma with the softened #704 rule — the family is insensitive to those changes
- AppWorld benchmark via cuga-eval; models `aws/gpt-oss-120b` (temp 0.1, max_completion 16000) and `google/gemma-4-31B-it` (temp 0.1, max_completion 3000/7000)

Representative claimed outcomes:

> "The original cart contained six items. After the order, the cart was cleared and the original items were re-added." — `4815c06_1` (gpt-oss): ordered the whole cart, then tried to undo the collateral damage

> "Appended `May 2023: $424.00` to the correct 'Venmo Expenses' note and removed the erroneous entry from your habit tracking log." — `b6d1104_1` (gemma): right amount, wrong format, after first writing into the wrong note entirely

### Related

- #407 (open) — the read-side complement: grounding of *claims*; this issue is grounding of *writes*
- #542 (closed) — reflection-side scope discipline; the executor-side gap remained
- #272 (open, Design) — weak output schemas make observation harder; this issue is about the obligation to observe
- #704 / #706 — post-error obstacle handling; together with this rule: what to read before writing, and what to do after an error

## Checklist

mikemikimike:
Fixes #709

## Summary
- Add a domain-neutral critical rule requiring read-before-write grounding, minimal deltas, collection scoping, and attribute preservation.
- Add a prompt rendering regression test.

## Compatibility
This only strengthens executor guidance and adds no dependencies or runtime API changes.

## Tests
- `ruff check D:\project\tscodex\github_pr\contrib-1-cuga-709\tests\unit\test_agent_spawn.py` — passed
- `git diff --check` — passed
- `python -m pytest --confcutdir=tests/unit tests/unit/test_agent_spawn.py -q` — blocked by missing dev dependency `dynaconf` in the host environment; `uv run pytest ...` began repository dependency resolution but did not complete within the local command window

Known limitation: the repository test environment could not be fully provisioned locally within the available run window.

## Summary by CodeRabbit

- **Improvements**
  - Agents now verify the latest observed state before creating, updating, deleting, moving, or recreating items.
  - Requested changes preserve unrelated fields, items, and attributes while carrying over exact quantities and details.
  - Pagination guidance remains included and is now clearly ordered after the new state-verification rule.
- **Tests**
  - Added coverage to confirm the agent prompt enforces observed-state checks before writes.


mikemikimike:
Addressed the review feedback in commit 1a8ec2b3: - Renumbered the rules after pagination so the sequence is unique and continuous (pagination remains rule 4). - Added the explicit @pytest.mark.unit marker to test_prompt_requires_observed_state_before_writes. Validation: git diff --check passed; ruff check passed. The focused pytest file is currently blocked by the repository environment because dynaconf is unavailable during fixture setup (61 collection/setup errors), so no test assertion executed.

sami-marreed:
Thanks for putting this together. Issue #709 is owned by the CUGA team, and changes to the core agent prompt require a full regression evaluation before they can be merged. The focused checks in this PR are not sufficient to assess the broader impact.

We should have stated the ownership and regression requirements in the issue and in `CONTRIBUTING.md`. Sorry that we did not make this clear before you started, and thank you for the time you invested. The team will take this change through the required regression process internally.

CC: @Sergey-Zeltyn

PR #735, opened 2026-09-01
feat(memory): add manual retention controls

illeatmyhat:
## Summary

Add the retention backend and bundled Evolve runtime on top of the Evolve management API merged in #736. Administrators can run retention policies, manage schedules, and inspect retention results for their service instance. The supported UBI image installs Evolve, starts it alongside CUGA, and downloads the default local model and tokenizer assets during the build.

This PR targets `main`. The memory workspace frontend is handled separately in #737. No README changes.

## Native Evolve API boundary

- Reuse Evolve 1.2.0's `build_memory_router()` for entity mutations and generic policy, rule, retention-run, schedule, and job APIs. The sole retention contract is `/api/manage/retention/...`; remove the unreleased duplicate CUGA routes, request models and report/policy projections. The #737 frontend should use native payloads, including `definition` for schedule writes and explicit `dry_run: false` to execute runs/schedules (native defaults are previews).
- Host the native router beside MCP in the bundled Evolve process. Both transports share the same singleton `EvolveClient` and one `retention_runtime`; CUGA does not open a second filesystem client or start another scheduler. The container supervisor generates a private service credential. CUGA forwards only server-derived namespace/user/agent/management scope to the loopback worker, never caller scope headers.
- Keep feature gates, chat/manage authorization, configured-agent checks, sanitized errors and PostgreSQL capability gates in CUGA. Reject request-supplied namespace/management identity and external deletion matches. Bound candidate/audit scans. Facts and manual source-deletion injection are not exposed through this gateway.
- Keep CUGA conversation links and usage enrichment, default-policy provisioning and transactional deletion outbox behavior. Metadata mutations use native Evolve behavior, then CUGA's scoped detail enrichment. Inventory/detail filters and enrichment remain CUGA-specific MCP wrappers. Capabilities and schedule preview use the canonical `/api/manage/retention` paths.
- External MCP deployments remain supported for chat and existing memory APIs. Native retention endpoints require the bundled HTTP worker and return explicit 503 otherwise; they never silently open local storage. There is no retention compatibility adapter or duplicate retention HTTP surface.
- Bundled memory APIs require a signed-in, non-placeholder identity for both personal and management operations. Authentication-disabled `default_user` requests return 401 rather than inventing an identity or accessing anonymous/global memory. Missing service-instance identity fails closed.

## Retention and deletion tracking

- Add management-protected APIs for policy listing and validation, policy execution and run history, compliance status, mark/sweep, and candidate/audit inventory.
- Add schedule list/get/save/start/stop/delete APIs and schedule previews. Evolve owns policy and schedule persistence, scheduling, execution, and run history; CUGA supplies service-instance scope and administrator attribution; Evolve enforces revisions.
- Require a configured service-instance namespace for Evolve calls other than policy validation. Preserve agent attribution when saving trajectories, and expose management permissions using the same role checks enforced by the API.
- Create a backend-aware `cuga-standard` policy when absent: delete unused guidelines after 180 days (skip when access evidence is missing), flag guidelines older than 90 days, and delete trajectories older than 365 days with derived-memory cascading. On a compatible PostgreSQL backend, also include the seven-day source-deletion rule. Preserve existing stored policies.
- Commit conversation deletion and its source-deletion outbox event atomically on SQLite and PostgreSQL. Deliver committed events to Evolve every 30 seconds when the integration is enabled, and acknowledge them only after Evolve confirms receipt. Events contain identifiers and deletion time, not message content; service-instance purge includes the outbox.
- Source-deletion matching uses explicit receipts, not absence from a conversation listing. Unknown provenance is retained and historical deletions are not backfilled. Evolve applies legal-hold checks during sweep.
- Report backend capabilities explicitly. Mark/sweep and candidate/audit operations return 409 without Evolve PostgreSQL support; unsupported source-deletion delivery stays queued. Filesystem-backed retention can still run the supported default rules.
- Use Evolve's content-free native retention reports. Execute SQLite batches on a dedicated, closed connection without holding the shared connection lock while waiting for a writer.

## Container startup and offline assets

- Install the `evolve` extra (including Evolve’s regex PII filter and hook engine) in `Dockerfile.ubi`, put `/app/.venv/bin` on `PATH`, and retain installed extras during Playwright installation with `uv run --no-sync`.
- Ship CUGA-owned `src/cuga/configurations/evolve/hooks.yaml` and select it through `EVOLVE_HOOKS_CONFIG` in the image. It configures local regex PII redaction on memory writes and Evolve LLM hooks, metadata normalization, access tracking, and per-entity legal-hold deletion protection. Operators can override the path with a mounted configuration; this does not enable `DYNACONF_EVOLVE__ENABLED`. No POC example file or runtime package installation is used.
- Route the standard entrypoint modes through `container_services.py`. When `evolve-mcp` is installed, start the private native HTTP/MCP worker at `127.0.0.1:8201`, wait for SSE readiness, then start CUGA with direct-mode connection settings. Explicit external URL/registry environment settings bypass the bundled worker. Default Evolve storage is `/data/dbs/evolve`; `DYNACONF_EVOLVE__ENABLED` remains controlled by the deployer. If Evolve is absent, execute CUGA directly.
- Forward shutdown signals to child process groups and stop the container if either service exits, allowing Kubernetes to restart it.
- Preload FastEmbed and its tokenizer, Docling layout/table/OCR assets, Evolve's default/configured embedding models, and registered tiktoken encodings. Download/import failures fail the build.
- Set runtime cache paths and offline flags, including LiteLLM’s `CUSTOM_TIKTOKEN_CACHE_DIR` so hook imports do not redirect tokenizers away from the preloaded cache. Run a network-disabled build check as the runtime user that loads embeddings/tokenizers, initializes the Docling PDF pipeline, launches Playwright Chromium, and exercises configured hooks against isolated storage (redaction, metadata, access stamps and legal hold).

Offline verification covers the bundled default assets. Optional models/plugins and deployment-specific LLM or tool services still require appropriate provisioning.

## Agent behavior and CI

- Keep CugaLite natural-language continuation opt-in (`cuga_lite_nl_auto_continue = false`). Add regression replays for the two observed planning-only CI replies with the feature explicitly enabled and disabled; preserve existing execution limits.
- Increase the full Linux stability suite's job deadline from 10 to 15 minutes so live-model tests and service startup can finish. The Windows smoke deadline remains unchanged.

## Validation

- Earlier validation of the unchanged outbox/auth behavior, continuation/classifier/budget tests, and graph/adapter tests passed. Current native API validation is listed below; Ruff checks pass.
- Local UBI build and network-disabled asset verification passed during container development. PostgreSQL deletion/outbox rollback and commit behavior was also validated.
- Local live approval/account tests could not validate model behavior because the configured endpoint timed out.
- Native boundary update: 64 local native-router/retention/container/memory/identity tests passed against the installed Evolve 1.2.0 package. These exercise feature/auth/scope gates, real cross-user/agent/namespace mutations, cross-agent schedules/jobs/runs, native payload defaults, external MCP preservation, default provisioning, and lifecycle ownership. Ruff and whitespace checks pass.
- A real worker smoke test in the existing UBI image with networking disabled verified SSE readiness, service authentication, native policy writes, and graceful scheduler/client shutdown. The Docker-installed package exposes `build_memory_router()`; no dependency version assumption was made from the host installation.
- CI at the pre-refactor head `6a5e99c5`: all Linux stability versions and Windows smoke, manager/load, policy e2e, graph/unit, events, pgvector, lint, CodeQL, and wheel-install checks passed. Policy unit failed in the tool-approval deny flow; SDK failed in the supervisor playbook zero-delegation assertion. The identical SDK assertion also fails on [base `main` at `b3bdcaee`](https://github.com/cuga-project/cuga-agent/actions/runs/35234441812/job/105246610741). CI for the production-hook update is pending.

- Production-hook update: 12 focused hook/configuration/container tests passed; Ruff, format and whitespace checks passed. Full `docker build -f Dockerfile.ubi -t cuga-pr735:production-hooks .` passed, including the runtime-user `RUN --network=none` check for all bundled assets and hooks. The build reproduced and fixed LiteLLM’s tokenizer cache override before passing.


illeatmyhat:
@coderabbitai review

src/cuga/backend/evolve/retention.py
```diff
@@ -0,0 +1,127 @@
...
+    "source_task_id",
+}
+
+
+def sanitize_retention_report(report: dict[str, Any]) -> dict[str, Any]:
+    """Remove memory content, ownership data, policy internals, and provider details."""
+    sanitized = {key: report[key] for key in _REPORT_FIELDS if key in report}
+    for bucket in ("flagged", "deleted", "skipped"):
+        sanitized[bucket] = [
+            {key: value for key, value in item.items() if key in _REPORT_ITEM_FIELDS}
```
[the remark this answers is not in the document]
illeatmyhat: Updated for the current implementation (#735 head 841e2fac). The earlier description of CUGA persisting sanitized reports is obsolete: CUGA's retention store was removed, and Evolve 1.2.0 owns durable runs and audit records. CUGA now validates each incoming report once and exposes only structured audit fields and fixed public messages; memory titles, content, metadata, and raw provider diagnostics are excluded. The current 47-test focused backend suite includes report-redaction coverage.

src/cuga/backend/server/memory_routes.py
```diff
@@ -185,6 +193,90 @@ def _validate_metadata_patch(metadata: dict[str, Any], allowed: set[str], audien
...
+            await EvolveIntegration.list_entities(
+                agent_id=agent_id,
+                cursor=cursor,
+                limit=min(remaining, 200),
+                include_content=False,
+                namespace_id=_namespace_id(),
+            )
+        )
+        page = [item for item in result.get("items", []) if isinstance(item, dict)]
+        entities.extend(page[:remaining])
```
sami-marreed, line 212: **[P1] Resolve source trajectories outside the candidate scan limit**

With `scan_limit=1`, an old guideline can be included while its source trajectory is on a later page. Orphan detection builds its provenance map only from this limited inventory, so it classifies the guideline as orphaned even when its CUGA conversation exists, and the handler deletes it. The default 100,000 cap has the same issue. Resolve referenced trajectories independently, or preserve memories whose provenance cannot be established from an incomplete inventory.

src/cuga/backend/evolve/retention.py
```diff
@@ -0,0 +1,282 @@
...
+
+    threads = {thread_id for thread_id, _ in conversation_keys}
+    orphaned = []
+    for entity in entities:
+        if entity.get("type") == "trajectory":
+            continue
+        metadata = _metadata(entity)
+        if metadata.get("legal_hold") is True:
+            continue
+        created_at = _created_at(entity.get("created_at"))
```
sami-marreed, line 168: **[P1] Measure the grace period from confirmed source unavailability**

This checks memory creation time, so a months-old memory becomes immediately eligible when its source conversation disappears today. It does not implement the stated seven-day period of source unavailability. Please record when absence was first confirmed and require seven elapsed days, resetting that observation if the source returns. Otherwise, the deletion reason incorrectly claims that the source was unavailable for more than seven days.

src/cuga/backend/evolve/retention_store.py
```diff
@@ -0,0 +1,94 @@
...
+async def save_retention_run(
+    *,
+    run_id: str,
+    agent_id: str,
+    actor_id: str,
+    report: dict[str, Any],
+) -> None:
+    await _ensure_schema()
+    tenant_id, instance_id = _scope()
+    status = "failed" if report.get("errors") else "completed"
```
sami-marreed: **[P2] Derive run status from the sanitized error field**

The handler passes a sanitized report, which replaces `errors` with `error_count`. Checking `report.get("errors")` therefore stores runs with provider errors or failed orphan deletions as `completed`. Please derive status from the sanitized contract and add a route-to-store regression test with a failed deletion.

src/cuga/backend/evolve/retention.py
```diff
@@ -0,0 +1,282 @@
...
+        sanitized_items = []
+        for item in report.get(bucket, []):
+            if not isinstance(item, dict):
+                continue
+            sanitized_item = {
+                key: value
+                for key, value in item.items()
+                if key in _REPORT_ITEM_FIELDS and isinstance(value, (str, int, float, bool, type(None)))
+            }
+            if title := memory_title(item):
```
sami-marreed, line 220: **[P2] Exclude user-controlled titles from persisted retention reports**

Users can edit memory titles, so truncating a title does not make it content-free. This copies potentially sensitive text into durable history and returns it after the memory is deleted. Please retain only identifiers, structured outcomes and controlled reason text, and verify that a unique private title is absent from both stored and projected reports.

src/cuga/backend/server/memory_routes.py
```diff
@@ -418,3 +510,120 @@ async def patch_admin_memory_entity(
...
+
+    run_id = str(uuid.uuid4())
+    try:
+        reference_time = retention_reference_time(body.as_of)
+    except ValueError as exc:
+        raise HTTPException(status_code=422, detail="as_of must be an ISO-8601 timestamp") from exc
+    inventory = await _list_retention_inventory(agent_id=agent_id, scan_limit=body.scan_limit)
+    conversation_keys = await get_conversation_db().get_thread_owners_for_agent(agent_id)
+    orphaned = find_orphaned_memory_entities(inventory, conversation_keys, now=reference_time)
+    provider_report = _memory_result(
```
sami-marreed: **[P2] Persist the operation before starting destructive work**

The first audit write happens after provider retention and orphan deletion. If the provider applies changes before timing out, or the final history write fails, records may already be deleted without a durable run record. Retrying also creates a new run ID. Please persist a `running` operation before side effects and record partial or unknown outcomes, with a stable retry identifier and reconciliation for uncertain provider results.

sami-marreed (changes requested):
Requesting changes because incomplete inventory can cause deletion of valid derived memories, and the stated seven-day grace period is not enforced from source unavailability. The inline comments also cover incorrect failure status, retained title content, and missing durable operation tracking before destructive work.

Validation: 13 targeted unit tests passed. Separate safe probes against the actual helper/function code reproduced the incomplete-inventory false orphan, immediate eligibility after source removal, title retention, and completed status with error_count=1. No live Evolve or real-data deletion was performed.

src/cuga/backend/evolve/retention.py
illeatmyhat: Updated for Evolve 1.2.0 and #735 head 841e2fac. CUGA does not persist retention reports. Evolve owns the durable collection/audit records; CUGA's single typed report projection drops titles, metadata, content, and raw provider text from every report bucket, including flagged/skipped items. The frontend also no longer models report titles. Current tests inject private labels/content and verify they do not reach the projected report. Public reason text is produced by CUGA for the response; it is not described as being stored in Evolve.

src/cuga/backend/evolve/retention_store.py
illeatmyhat: Verified against current #735 head [841e2fac](https://github.com/cuga-project/cuga-agent/commit/841e2fac). CUGA's retention_store.py and local retention table were removed. Released Evolve 1.2.0 owns run status and persistence; CUGA reads the stored status through list_retention_runs. CUGA's public report projection no longer participates in deriving durable run status. The current backend tests cover reading Evolve-owned history and excluding provider details.

src/cuga/backend/server/memory_routes.py
illeatmyhat: Updated for released Evolve 1.2.0. Applied PostgreSQL retention uses durable mark-and-sweep collection. Evolve records the run before work, persists marks, and commits each deletion with its audit receipt transactionally. If an execution is interrupted, committed marks/outcomes remain and a later policy run can re-evaluate remaining candidates. This does not promise partial-job replay or exactly-once execution; CUGA performs no independent retention deletion or local run persistence. The Settings UI now uses a single Run retention now action and reads Evolve-owned history.

src/cuga/backend/evolve/retention.py
illeatmyhat: Updated for the current source-deletion design. The default rule requires a memory age of at least seven days AND an explicit matching source-deletion receipt, rather than absence from a conversation listing. It does not promise seven days since deletion. CUGA commits conversation deletion and its delivery outbox atomically, then retries delivery to Evolve. Unknown provenance and memories created after the deletion are retained. The UI describes the minimum memory age accurately, and audit explanations no longer hardcode a grace period or rule thresholds. This is a change to the original semantics, not an implementation of a seven-day post-deletion grace period; leaving the conversation open for confirmation of that contract.

src/cuga/backend/server/memory_routes.py
illeatmyhat: Updated for #735 head 841e2fac and Evolve 1.2.0. The earlier HTTP 409 inventory-limit workaround has been removed along with CUGA's orphan detector and inventory scanner. CUGA no longer infers deletion eligibility from incomplete conversation listings. Evolve evaluates explicit source-deletion receipts matched to service instance, user, agent, and source; unknown provenance is retained. A missing trajectory on a CUGA inventory page therefore cannot cause CUGA to classify or delete a memory as orphaned.

tests/unit/test_evolve_retention.py
```diff
@@ -0,0 +1,460 @@
...
+        {
+            "entity_id": "orphan-a",
+            "rule": "orphaned-conversations",
+            "reason": "orphaned_conversation",
+            "detail": "source conversation remained unavailable beyond the grace period",
+        }
+    ]
+
+
+def test_admin_can_list_evolve_owned_retention_policies(client):
```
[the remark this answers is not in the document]
illeatmyhat: Verified against current #735 head [841e2fac](https://github.com/cuga-project/cuga-agent/commit/841e2fac). Both new tests now carry explicit pytest.mark.unit decorators in addition to the file-level unit marker.

sami-marreed:
Full re-review of `37aa12fc6ba4bfa0ef0c161739a806236b049e46`: **keep CHANGES_REQUESTED**. Three additional issues were reproduced beyond the two remaining concerns from the original review.

### P1 — Preserve ownership identifiers through PII redaction (new)

[hooks.yaml:5–16](https://github.com/cuga-project/cuga-agent/blob/37aa12fc6ba4bfa0ef0c161739a806236b049e46/src/cuga/configurations/evolve/hooks.yaml#L5)

The shipped `PIIFilterMemoryPlugin` redacts the entire entity, including identity/provenance metadata. Using the pinned packages and production hooks configuration, a guideline owned by `alice@example.com` is stored with both `user_id` and `owner_id` equal to `[REDACTED]`. The gateway still authorizes with the original subject, and that user's PATCH request returns **403**. Exact source-deletion receipt matching also cannot match the original user ID to the transformed metadata.

Please preserve ownership/provenance fields during content redaction, or apply a consistent opaque identity mapping across writes, authorization and receipts. Add an email-based subject regression test.

### P1 — Enforce the requested post-deletion grace period (existing)

[retention.py:11](https://github.com/cuga-project/cuga-agent/blob/37aa12fc6ba4bfa0ef0c161739a806236b049e46/src/cuga/backend/evolve/retention.py#L11) · [Original comment](https://github.com/cuga-project/cuga-agent/pull/735#discussion_r3961525151)

`source_deleted: true` combined with `max_age_days: 7` measures memory age, not elapsed time since source deletion. A 90-day-old memory becomes eligible immediately when its source-deletion receipt arrives today. The earlier reply acknowledges replacing the original semantics.

Please implement seven elapsed days from confirmed source deletion, or get explicit agreement to the replacement contract before resolving this point.

### P2 — Remove private task text from cascade audit reasons (new)

[Default cascade rule](https://github.com/cuga-project/cuga-agent/blob/37aa12fc6ba4bfa0ef0c161739a806236b049e46/src/cuga/backend/evolve/retention.py#L25) · [Native response forwarding](https://github.com/cuga-project/cuga-agent/blob/37aa12fc6ba4bfa0ef0c161739a806236b049e46/src/cuga/backend/server/evolve_native_routes.py#L210)

CUGA uses `state.sub_task` as a trajectory task ID; the normalizer copies it into `trace_id`. Evolve's filesystem execution path creates `reason="cascade:<trace_id>"`, and the audit projection preserves it.

Reproduction through the actual native router with the shipped hooks: create an old trajectory and derived guideline linked by the synthetic task `CONFIDENTIAL acquisition of Project Heron`, then execute the old-session cascade rule. Both memories are deleted, while **both the response and durable run history retain `cascade:CONFIDENTIAL acquisition of Project Heron`**.

Please use controlled reason codes and opaque references. Fix durable persistence as well as the public response; gateway-only sanitization is insufficient. The title-specific fix works, but the broader content-free audit requirement is still unmet.

### P2 — Preserve external Evolve settings loaded from files (new)

[container_services.py:106–118](https://github.com/cuga-project/cuga-agent/blob/37aa12fc6ba4bfa0ef0c161739a806236b049e46/scripts/container_services.py#L106)

The supervisor checks only `os.environ` before selecting the bundled worker and setting `DYNACONF_EVOLVE__MODE` / `DYNACONF_EVOLVE__URL`. This happens before CUGA loads its ordinary `.env` configuration.

Reproduction: a temporary `.env` specifies `DYNACONF_EVOLVE__URL=https://external.example/sse`, with `evolve-mcp` available. The supervisor selects the bundled worker; importing the actual CUGA configuration afterward resolves the URL to `http://127.0.0.1:8201/sse`. Existing external memories consequently appear unavailable, and subsequent operations use different storage.

Please resolve the effective operator configuration before choosing the worker or overriding connection settings. Cover file-based configuration as well as exported environment variables.

### P2 — Preserve operation identity across failure and retry (existing, further reproduced)

[evolve_native_routes.py:182–197](https://github.com/cuga-project/cuga-agent/blob/37aa12fc6ba4bfa0ef0c161739a806236b049e46/src/cuga/backend/server/evolve_native_routes.py#L182) · [Original comment](https://github.com/cuga-project/cuga-agent/pull/735#discussion_r3961525169)

Persisting a run before execution is now handled. However, the gateway replaces upstream errors before reading their structured payload, discarding Evolve's safe `run_id`. Native `RunRequest` also provides no run ID or idempotency field.

Reproduction: inject a scan failure and submit the same manual operation twice. Evolve persists **two distinct failed run IDs**; both responses are **502** with `{"detail":"Memory request rejected"}` and no ID. Supplying an existing `run_id` returns **422**.

Please preserve an allowlisted operation ID in failure responses and provide stable manual retry/reconciliation behavior. Scheduled-job recovery does not cover manual runs.

### Validation

| Check | Result |
|---|---|
| Container lifecycle, conversation history/outbox, service-instance cleanup, manage-access and preload tests | 41 passed |
| Native Evolve gateway and retention tests | 39 passed |
| Bundled hooks and planning continuation tests | 6 passed |
| Additional safe reproductions | Ownership redaction, persisted cascade task text, and lost failure/retry identity confirmed |
| Separate startup configuration probe | External `.env` URL override confirmed |
| GitHub checks at reviewed head | 21 SUCCESS, 3 SKIPPED; none failing or pending |

The review covered all 33 changed files and the exact locked Evolve 1.2.0 wheel, with its SHA256 verified. Hook probes used cpex 0.1.3 and cpex-pii-filter 0.3.8. Local runs reused available CUGA dependencies with the PR source and exact Evolve/hook packages; this was not a fresh full locked-environment installation. All reproduction data was synthetic and temporary. Docker and live PostgreSQL were not rerun locally.

The incomplete-inventory orphan path and old CUGA failure-status derivation are addressed. The findings above still prevent approval.

Issue #188, opened 2026-05-04
ui: missing in-progress indicator while agent turn is in flight (esp. with slow LLMs)


haroldship:
## Summary

The chat UI sometimes does not show a visible "agent is working" indicator while the agent is mid-turn. This is most noticeable when the LLM is slow to respond — the user sees their message in the thread, then nothing, with no spinner / typing indicator / "Agent is thinking…" label, until tokens finally start streaming or the turn completes.

Result: the user can't tell if (a) the agent is working, (b) the request is stuck, or (c) the connection dropped. They tend to retry, which makes the queueing worse.

## Where this lives

The agentic chat frontend already tracks streaming state in [src/frontend_workspaces/agentic_chat/src/StreamManager.tsx](src/frontend_workspaces/agentic_chat/src/StreamManager.tsx) (`setStreaming(streaming: boolean)`) and toggles it in [src/frontend_workspaces/agentic_chat/src/StreamingWorkflow.ts](src/frontend_workspaces/agentic_chat/src/StreamingWorkflow.ts) around lines 51, 198, 297. So the signal is present — it just isn't surfaced consistently in the UI.

It looks like the indicator is tied to *streaming tokens being received*, not to *a turn being in flight*. There is a window between "user message sent" and "first token arrives" where streaming is technically false (no tokens yet) but the agent is very much working. That window is exactly when slow LLMs (e.g. WatsonX.ai under load) leave the user staring at a blank thread.

## Expected behavior

- As soon as the user submits a message, show an in-progress indicator (spinner, animated dots, or a per-step "planning / calling tool / generating answer" status row).
- Keep it visible until either (a) the first response token streams in **and** the streaming-text bubble is visible, or (b) the turn errors out / is aborted.
- Optional: show which agent/node is currently active (planner, code agent, browser agent, etc.) — the activity tracker already has this info via [src/cuga/backend/activity_tracker/tracker.py](src/cuga/backend/activity_tracker/tracker.py).

## Suggested fix

1. Introduce a `turnInFlight` boolean separate from `isStreaming`, set true on submit and false on terminal frame / error / abort.
2. Render an indicator bound to `turnInFlight || isStreaming`, not just `isStreaming`.
3. (Stretch) Tie the indicator label to the current activity-tracker step name so the user gets per-stage feedback ("Planning…", "Calling `create_event`…", etc.).

## Repro

[...]

PR #740, opened 2026-09-02
fix(registry): normalize AppWorld file_system paths to canonical ~/ form (#730)

### LINKED ISSUE #730: [Bug]: File-export conventions lose otherwise-correct AppWorld tasks — '/./' instead of '~/' paths, CSV quoting/header/column-order mismatches
## What happened

On the AppWorld hard suite, CugaLite loses ~10 task-runs (≈5 pp) to **file-export conventions** while the exported *content* is correct. Two distinct halves, likely with different fixes:

### Half 1 — tilde-path mangling (`/./…` instead of `~/…`), write and read side

The agent addresses files with cwd-relative paths where the app expects tilde paths:

| task | runs | evidence (evaluator asserts / registry log) |
|---|---|---|
| `f323bae_1` | r2, r4 | created `'/./downloads/habit_tracker.csv'` — eval expects `'~/downloads/habit_tracker.csv'`; the export content itself parses |
| `33e202d_1` | r3 | entire recruiting tree written under `/./documents/work/recruiting/…` (csv + all pdf attachment paths) vs `~/documents/work/recruiting/…` |
| `6b6ca61_1` | r2 | **read side**: 175 identical rejected `GET file?file_path=./owe_list.csv` + 175 `directory?directory_path=.` (422s) — the malformed relative path is re-sent verbatim, feeding the #599 retry loop |

This is the same malformation family as #599's `29a7b7e_1` exemplar (double-slash concatenation, gpt-oss era) — filed there as retry-loop evidence, but the path bug itself has no issue. Candidate fix: deterministic normalization at the file-system tool boundary (resolve `.`/`./`/double-slash inputs against the user home and emit `~/…` canonical form), rather than prompt text.

### Half 2 — CSV formatting conventions (quoting, headers, column order)

| task | runs | evidence |
|---|---|---|
| `a1d3dfd_1` | r1–r3 (3/4 fail) | wrote header `"title","artists"` **with quotes and trailing `\r`** where the intent's literal header `title,artists` was expected; body rows quoted the same way, so every row mismatches; song/artist content correct |
| `f323bae_1` | r1, r2, r4 | body rows match **ignoring column order** but fail the ordered assert — columns not emitted in the source note's order |
| `33e202d_1` | r4 | the mirror error: **unquoted** rows (`adam,blackburn,…`) where the spec's quoted form (`"adam","blackburn",…`) was expected |

The failing intents specify formats literally ("Its first header should be …", explicit `'<name>' => $<price>'`-style templates elsewhere). Sibling non-CSV case: `d37c235_1` (med) returned all promo codes **uppercased** against a case-sensitive check. Candidate fix: prompt-side rule in the CugaLite export/code path — *reproduce requested headers and templates verbatim (case, quoting, separators); quote CSV fields only if the spec shows quotes; preserve source ordering for rows/columns unless asked otherwise.*

Both halves are formatting-only losses: in every cited run the mutation/content asserts pass and only the convention asserts fail. Evidence is per-assert-verified in `cuga-eval/benchmarks/appworld/evaluation_bundles/20260828_203631_compare_mistral-medium-3.5-128b-hard-run4/mistral_hard_failure_analysis.md` (section "New at hard: export-format and path conventions").

## How to reproduce

AppWorld hard suite via cuga-eval, mistral-medium-3.5 profile. Highest-signal single tasks:

```bash
./scripts/eval.sh --benchmark appworld -- --task a1d3dfd_1   # CSV quoting, failed 3/4
./scripts/eval.sh --benchmark appworld -- --task f323bae_1   # column order + /./ path, failed 3/4
```

Judge on the per-assert signatures (header string, tilde_path value, ordered-rows assert), not pass/fail alone.

## Environment

- cuga-agent `98fc9372` (hard run1) / `81d8ea85` (hard runs 2–4), 2026-08-27…08-28
- Model: `mistralai/Mistral-Medium-3.5-128B` via RITS (verified from traces), temp 0.1; `CONTEXT_SUMMARIZATION__ENABLED=false`
- Bundles: `20260827_150341_…hard-run1`, `20260828_040806_…hard-run2`, `20260828_140625_…hard-run3`, `20260828_203631_…hard-run4` (cuga-eval)

```
f323bae_1 r2: AssertionError: '/./downloads/habit_tracker.csv' == '~/downloads/habit_tracker.csv'
a1d3dfd_1 r1: AssertionError: '"title","artists"' == 'title,artists'
              (Original values: '"title","artists"\r' == 'Title, Artists')
6b6ca61_1 r2: 175x rejected GET /file_system/file?file_path=.%2Fowe_list.csv (422)
```

### Related

- #599 — the retry loop that amplifies the read-side path bug (its `29a7b7e_1` double-slash exemplar is this family); a path fix removes a #599 trigger, and the #599 guard removes the amplification, but neither substitutes for the other
- #709 — the neighbor family for *mutating existing* artifacts ("match the artifact's conventions"); this issue is about conventions of **newly created** exports and is deliberately scoped apart

Filed without a parent: small self-contained bug family — narrow deterministic repro, no design change.

## Checklist

Sergey-Zeltyn:
## Bug fix

Fixes #730 (Half 1 — tilde-path mangling; the CSV-convention half is prompt-side in cuga-eval and tracked separately in the issue)

### Summary

On the AppWorld hard suite, generated code sometimes addresses `file_system` with cwd-relative paths (`./downloads/habit_tracker.csv`, `directory_path=.`). AppWorld's `process_path` rejects `//` with a 422 and silently roots any other non-absolute path, so the write is stored as `/./downloads/habit_tracker.csv` — content asserts pass but the `tilde_path` assert fails — and read-side calls 404/422 forever, feeding the #599 retry loop (`6b6ca61_1`: 175 identical rejected calls).

**Root cause of the model behavior:** the CugaLite base prompt's sandbox-workspace section ("use relative paths for every file operation") leaks into `file_system` API arguments. This PR adds the deterministic backstop at the tool boundary rather than a prompt change (deliberately scoped out to avoid prompt drift across deployments).

Changes:

- **`appworld_path_normalizer.py`** (new): `normalize_appworld_path` collapses duplicate slashes, resolves `.`/`..` segments (never escaping the anchor), and anchors relative forms at home — `./x` → `~/x`, `.` → `~/`, bare `x.csv` → `~/x.csv`, leading `/./` treated as the server-side echo of a relative path. Canonical `~/…` and absolute `/…` pass through unchanged.
  - Pure string logic with POSIX `/` semantics on purpose: `os.path` (ntpath on Windows), `pathlib`, and `expanduser` (host home injection) must never touch these host-independent virtual paths. Backslash-containing values are left untouched.
  - Scope: only `advanced_features.benchmark == "appworld"`, only the `file_system` app, only string args whose key is `path` or ends with `_path`. Inert everywhere else — production MCP tools that document relative paths are never rewritten.
- **`api_registry_server.py`**: hook at the `/functions/call` choke point — after the `api_call` trace step (traces keep the raw args the model produced), **before** `rejected_call_guard.check` so #599 signatures dedupe on canonical args and a corrected call is never short-circuited by rejections recorded under a malformed form. Every rewrite is warn-logged with `original -> normalized`.

**Reviewer sign-off requested on one rule:** bare relative `x.csv` → `~/x.csv`. AppWorld has no cwd concept and its silent `/`-prepend is itself a trap, so home is the only sane anchor — but this is the one rewrite that changes semantics for a hypothetical root-level target. The conservative fallback is to normalize only dot-prefixed/double-slash forms.

### Testing
- 30 new unit tests (`test_appworld_path_normalizer.py`): canonicalization cases derived from the #730 evidence, pass-through and scoping cases, and a route test pinning the ordering contract (canonical path on the wire; malformed + canonical spellings share one #599 rejection counter).
- Full registry suite incl. the #599 guard: 150 passed. All six CI "Unit (colocated)" suites: 1,046 passed. `tests/unit` + `tests/integration`: 1,694 passed (3 pre-existing `test_sdk_citations.py` failures reproduce identically on clean `main`; pass in isolation).
- End-to-end validation against the repro tasks (`f323bae_1`, `6b6ca61_1`, `33e202d_1`) is pending a cuga-eval run; the rewrite warn-logs will show when the normalizer is load-bearing. **Runs has shown that solution works, although most tasks failed due to other reasons.**

## Summary by CodeRabbit

* **Bug Fixes**
  * AppWorld file-system paths are now normalized to consistent canonical forms, including relative paths and paths with redundant or dot segments.
  * Equivalent path formats now share rejected-call tracking, reducing duplicate rejection handling.
  * Original request arguments remain available in call traces while normalized paths are used for validation.

* **Tests**
  * Added coverage for path normalization, scope restrictions, unchanged inputs, and rejection handling.


offerakrabi (approved):
## PR Review: #740 — fix(registry): normalize AppWorld file_system paths to canonical ~/ form (#730)

I also tried to check compatibility across platforms, Linux/MacOS/Windows, while my checks are not definitive (could not explicitly test this) it looks like it should work fine on all platforms

**Merge.** The fix is correct, well-scoped, and provably OS-neutral. The choke-point placement and ordering contract are sound.
Nothing blocks. No findings.

|---|---|---|---|---|

## Notes

- Half 2 of #730 (CSV quoting/column-order) is explicitly out of scope here and tracked separately in the issue.
- E2E validation against the repro tasks (`f323bae_1`, `6b6ca61_1`, `33e202d_1`) is pending a cuga-eval run per the PR body; the warn-log will confirm when the normalizer is load-bearing.

src/cuga/backend/tools_env/registry/registry/appworld_path_normalizer.py
```diff
@@ -0,0 +1,116 @@
...
+    ``changes`` maps each rewritten key to ``(original, normalized)`` so the
+    caller can log what the model actually sent. When nothing applies (other
+    app, other benchmark, no rewrites) the original ``args`` object is
+    returned untouched.
+    """
+    if app_name != _FILE_SYSTEM_APP or not isinstance(args, dict):
+        return args, {}
+    from cuga.config import settings
+
+    if getattr(settings.advanced_features, "benchmark", None) != "appworld":
```
sami-marreed, line 105: The benchmark gate should use `cuga.config.resolved_benchmark()` instead of reading `settings.advanced_features.benchmark` directly. The settings value can be stale when `DYNACONF_ADVANCED_FEATURES__BENCHMARK` is set after `cuga.config` is imported, which leaves normalization disabled even though the resolved benchmark is `appworld`.

I reproduced this by importing `cuga.config`, setting the environment variable, and calling `normalize_file_system_path_args()`; the relative path remained unchanged. Please use `resolved_benchmark()` here and add a test for the late environment-variable case.

Sergey-Zeltyn:
## A/B smoke validation on AppWorld (main vs. this branch)

Ran the 4-task #730 smoke set twice under identical conditions — Mistral-Medium-3.5 via RITS, cuga-eval `4f28d93`, no prompt-side changes active (cuga-eval #188 **not** in the tree) — so `cuga` is the only variable.

| Arm | Bundle | cuga |
|---|---|---|
| Baseline | `20260915_210958_compare_mistral-medium-3.5-128b-smoke730` | `main` |
| Branch | `20260916_174126_compare_mistral-medium-3.5-128b-smoke730-branch` | `76977d82` (this PR) |

Tasks: `6b6ca61_1` (read-side `.` / `./` exemplar), `f323bae_1` (write-side + tilde_path assert), `a1d3dfd_1` (write-side), `29a7b7e_1` (clean-path control).

### Normalizer activity: 3 rewrites, 3 correct, 0 false positives

```
file_system_show_directory_directory_get  directory_path: '.'                    -> '~/'
file_system_show_file_file_get            file_path:      './owe_list.csv'       -> '~/owe_list.csv'
file_system_create_file_file_post         file_path:      './spotify_backup.csv' -> '~/spotify_backup.csv'
```

Untouched, as scoped: 42 clean `file_system` calls on the control task, `f323bae_1`'s canonical `~/downloads/habit_tracker.csv`, and every absolute `/home/<user>/…` source path. Raw pre-rewrite args remain visible in the trajectory `tool_calls`, confirming the "trace keeps what the model produced, the wire gets canonical" ordering contract.

### Mechanism: the #599 seed call now succeeds

`6b6ca61_1`:

| Call | Baseline (main) | Branch |
|---|---|---|
| `directory_path='.'` | **422** — `Directory with path /./ is not available in your account` | **ok** — rewritten to `~/`, returns a listing |
| `file_path='./owe_list.csv'` | 422 — `/./owe_list.csv` (malformed-path artifact) | 422 — `/home/lindsey/owe_list.csv` (well-formed path, file genuinely elsewhere) |
| `file_system` calls / errors | 8 / 4 | 5 / 1 |

The remaining 422 on the branch is semantically honest rather than a path artifact: the file lives under `~/documents/work/`, so the agent listed `~/` (a call that only succeeds thanks to the rewrite) and recovered from there.

### Task outcomes: unchanged, assert-for-assert

| Task | Baseline | Branch | Failure set |
|---|---|---|---|
| `6b6ca61_1` | 47.4% (9/19) | 47.4% (9/19) | identical — venmo/splitwise semantics |
| `f323bae_1` | 66.7% (6/9) | 66.7% (6/9) | identical — body rows (#750 pagination) |
| `a1d3dfd_1` | 72.7% (8/11) | 72.7% (8/11) | identical — CSV quoting / artist separator |
| `29a7b7e_1` | 100% ✓ | 100% ✓ | none |

Both arms: 25% pass rate, 0.717 avg match, ~112s/task. Per-task token and duration deltas run in both directions and are n=1 noise.

### Read

- The normalizer behaves exactly as documented: deterministic, correct on every firing, inert on clean traffic. No regressions.
- On this set it buys back wasted calls, not score. Since #733 landed, the rejected-call guard already prevents the catastrophic loop from #730 (175 identical rejected calls), so the failure mode this PR removes has degraded from run-killing to a handful of 422s. The value here is efficiency plus removing a tail risk, not flipping tasks — worth stating plainly rather than claiming a pass-rate win.
- The residual failures on all three repro tasks are out-of-scope classes (payment semantics, #750 pagination, CSV format conventions — the latter being the cuga-eval #188 half).

### Caveats

- n=1 per arm per task; Mistral-Medium-3.5 only. Rewrite incidence is model-dependent and should be re-measured if the eval default moves to another model.
- `directory_path=''` appeared in the baseline run and is pass-through by design (empty strings are not rewritten). Worth a quick decision on whether that form should join the rewrite set.
- The normalizer hooks registry `/functions/call`, so the AppWorld `--agent codeact` path (`world.execute()` REPL) bypasses it entirely.

src/cuga/backend/tools_env/registry/registry/appworld_path_normalizer.py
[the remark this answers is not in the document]
Sergey-Zeltyn: Addressed in 62bad79e.

The gate now goes through `cuga.config.resolved_benchmark()`, which consults `DYNACONF_ADVANCED_FEATURES__BENCHMARK` in the process environment before falling back to `settings`, so a benchmark selected after `cuga.config` is imported enables normalization. The AppWorld eval path was never affected (config_loader exports the env var before any `cuga.*` import, and the registry runs as a separate process) — the exposure was in-process/SDK embedding, which is exactly the case you described.

Two tests pin it in both directions: `test_benchmark_set_after_import_still_normalizes` (settings `default` + env `appworld` -> rewrite happens) and `test_env_benchmark_overrides_stale_appworld_settings` (settings `appworld` + env `default` -> no rewrite, so the scope cannot widen either). Both fail against the previous gate and pass now. The existing settings-based cases clear the env var so they exercise the fallback deterministically regardless of the shell the suite runs in. Registry suite: 152 passed.

One note for follow-up rather than this PR: `resolved_benchmark()` has had no callers in `src` since it landed in #148, and ten sites still read `settings.advanced_features.benchmark` directly — including `api_registry_server.py:381` and `api_registry.py:316` in this same package. This file is now the only user of the helper. Happy to open a separate issue to converge the rest.

Sergey-Zeltyn: Addressed in 62bad79e.

The gate now goes through `cuga.config.resolved_benchmark()`, which consults `DYNACONF_ADVANCED_FEATURES__BENCHMARK` in the process environment before falling back to `settings`, so a benchmark selected after `cuga.config` is imported enables normalization. The AppWorld eval path was never affected (config_loader exports the env var before any `cuga.*` import, and the registry runs as a separate process) — the exposure was in-process/SDK embedding, which is exactly the case you described.

Two tests pin it in both directions: `test_benchmark_set_after_import_still_normalizes` (settings `default` + env `appworld` -> rewrite happens) and `test_env_benchmark_overrides_stale_appworld_settings` (settings `appworld` + env `default` -> no rewrite, so the scope cannot widen either). Both fail against the previous gate and pass now. The existing settings-based cases clear the env var so they exercise the fallback deterministically regardless of the shell the suite runs in. Registry suite: 152 passed.

One note for follow-up rather than this PR: `resolved_benchmark()` has had no callers in `src` since it landed in #148, and ten sites still read `settings.advanced_features.benchmark` directly — including `api_registry_server.py:381` and `api_registry.py:316` in this same package. This file is now the only user of the helper. Happy to open a separate issue to converge the rest.

Sergey-Zeltyn:
Addressed the CodeRabbit follow-up (the "outside the diff" minor on the `api_call` trace) in 90657913: the route test now passes `trajectory_path` and asserts the persisted `api_call` step keeps the raw `./owe_list.csv` while the same request reaches the registry as `~/owe_list.csv` — moving normalization above the trace step now fails the suite instead of silently changing what trajectories record. Registry suite: 152 passed.

On the earlier red CI (62bad79e): the three failing jobs — Policy (unit) `test_tool_approval_deny_flow`, Manager `test_11_policies_isolation_with_intent_guard` (httpx.ReadTimeout), SDK `test_e2e_playbook_orchestrates_sub_agents` — are LLM/timing-sensitive tests in suites this PR does not touch; the identical code minus a registry-local gate change passed them yesterday. The push of 90657913 re-runs everything.

sami-marreed (approved):
Reviewed 90657913. No blocking findings. The earlier benchmark-gate issue is fixed with resolved_benchmark() and regression tests covering both environment-override directions. Normalization remains scoped to AppWorld file_system calls, preserves raw api_call traces, and runs before rejected-call tracking.

The intentional bare-relative x.csv -> ~/x.csv behavior is acceptable for this benchmark; explicit absolute paths remain unchanged.

Validation: 64 normalizer and rejected-call-guard tests passed locally; current CI workflows pass. AppWorld evaluations were not rerun as part of this review.

PR #759, opened 2026-09-05
feat: PR Regression Testing Oriented Changes

AnkitaNaik:
* **New Features**
  * Added an automated pull request evaluation workflow started from qualifying issue comments.
  * Evaluation results are posted to the pull request and saved as downloadable artifacts.
  * Added optional provider selection for evaluation runs.
  * Added status and outcome reporting for evaluation runs.

* **Improvements**
  * Updated language model connection settings for improved compatibility.
  * Prevented overlapping evaluations for the same pull request to keep results consistent.
  * Restricted workflow access and permissions for safer execution.

## Summary by CodeRabbit

- **New Features**
  - Added automated pull request evaluations with validated commands and configurable parameters.
  - Added pull request status reporting with request details, results, evaluation logs, and fallback error messages.
  - Added downloadable evaluation log artifacts.

- **Security**
  - Strengthened request validation, access controls, credential protection, and sensitive-value redaction.
  - Ensured evaluations run against the requested commit.

- **Tests**
  - Added coverage verifying canonical configuration for the RITS language model provider.


.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
+name: Run PR Evaluation
+
+on:
+  issue_comment:
+    types: [created]
+
+permissions:
+  contents: read
+  issues: write
+  pull-requests: write
```
sami-marreed, line 10: **Critical:** `issue_comment` runs from the default branch, then this job checks out `head.repo` and `import`s that tree with `PYTHONPATH`. `pull-requests: write` / `issues: write` stay on the same job, so `GITHUB_TOKEN` is still in the eval process even with `persist-credentials: false`.

Split this: eval job with `permissions: {}` and `GITHUB_TOKEN: ""`; a second trusted job (no PR checkout) posts the report from artifacts.

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
...
+    name: Run PR Evaluation
+    if: >
+      github.event.issue.pull_request &&
+      contains(github.event.comment.body, '/run-pr-eval') &&
+      (
+        github.event.comment.author_association == 'OWNER' ||
+        github.event.comment.user.login == 'AnkitaNaik' ||
+        github.event.comment.user.login == 'Sergey-Zeltyn' ||
+        github.event.comment.user.login == 'haroldship'
+      )
```
sami-marreed, line 27: On this org repo, most of us are `MEMBER`, not `OWNER`, so that clause likely never matches. Hardcoded logins also drift on rename/access change.

Please reuse `vars.BOB_ALLOWED_ACTORS` (or a GitHub team check) instead of a third user list.

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
...
+    runs-on: [self-hosted, linux, run-pr-eval]
+    timeout-minutes: 120
```
sami-marreed: **Critical:** Persistent self-hosted runners plus untrusted PR code is the GitHub-documented pwn path. The eval script also lives outside the repo at `$HOME/pr-regression-testing/.../run-pr-regression-eval.sh`, so a previous run can rewrite it.

Use ephemeral runners and vendor (or checksum-pin) the script in-repo. Treat `$HOME` as untrusted after any eval.

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
...
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.issue.number,
+            });
+
+            core.setOutput("head_repo", pr.data.head.repo.full_name);
+            core.setOutput("head_ref", pr.data.head.ref);
+            core.setOutput("head_sha", pr.data.head.sha);
+
+      - name: Checkout PR head commit
```
sami-marreed: **Critical:** The slash-command authorizes a commenter; this step then fetches whatever `head.sha` is current. A force-push between the comment and `pulls.get` evaluates a different commit.

Require `/run-pr-eval sha=<40-char>` and fail if `git rev-parse HEAD` does not match. Also refuse `head.repo.full_name != github.repository` unless an `environment:` gate is required.

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
...
+          git rev-parse HEAD
+
+          echo "CUGA source:"
+          if [[ -d "${GITHUB_WORKSPACE}/src/cuga" ]]; then
+            ls -la "${GITHUB_WORKSPACE}/src/cuga" | head
+          else
+            echo "${GITHUB_WORKSPACE}/src/cuga not found"
+          fi
+
+      - name: Parse eval provider
```
sami-marreed: This only `ls`s. It does not check SHA equality, same-repo origin, or a signature.

Replace with a hard fail if `HEAD` ≠ the SHA from the comment (and, if you keep forks, if `head.repo` is not this repository).

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
...
+            case "${token}" in
+              provider=*)
+                provider="${token#provider=}"
+                ;;
+            esac
+          done
+
+          provider="$(printf '%s' "${provider}" | tr '[:upper:]' '[:lower:]')"
+          echo "provider=${provider}" >> "${GITHUB_OUTPUT}"
+
```
sami-marreed: `${normalized_comment}` is unquoted (globbing), and `provider=` is written to `GITHUB_OUTPUT` with no allowlist.

Restrict to `{rits,litellm}` and fail closed, then `printf 'provider=%s\n' "$provider" >> "$GITHUB_OUTPUT"`.

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
...
+      - name: Run AppWorld evaluation
+        id: evaluation
+        shell: bash
+        env:
+          OPENAI_API_KEY: ${{ steps.eval-provider.outputs.provider == 'litellm' && secrets.OPENAI_API_KEY || '' }}
+          RITS_API_KEY: ${{ steps.eval-provider.outputs.provider == 'rits' && secrets.RITS_API_KEY || '' }}
+          LITE_LLM_KEY: ${{ steps.eval-provider.outputs.provider == 'litellm' && secrets.LITE_LLM_KEY || '' }}
+          AGENT_SETTING_CONFIG: settings.rits.toml
+          OPENAI_BASE_URL: ${{ vars.OPENAI_BASE_URL }}
```
sami-marreed: **Important:** `provider=litellm` injects `LITE_LLM_KEY` / `OPENAI_API_KEY` but `AGENT_SETTING_CONFIG` is always `settings.rits.toml`. LiteLLM in `models.py` reads `OPENAI_API_KEY` / `LITELLM_API_KEY`, not `LITE_LLM_KEY`, and still needs the RITS key which this branch does not inject.

Map provider → config and env names, or drop the claimed litellm switch until it works.

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
...
+          PYTHONPATH: ${{ github.workspace }}/src
+        run: |
+          set +e
+
+          mkdir -p "${HOME}/pr-regression-testing/logs"
+          run_started_at="$(date -u +"%Y%m%dT%H%M%SZ")"
+          short_sha="${PR_HEAD_SHA:0:12}"
+          log_prefix="pr-${PR_NUMBER}-${short_sha}-${run_started_at}"
+          output_log="${HOME}/pr-regression-testing/logs/${log_prefix}-run-pr-eval-output.md"
+          exit_code_log="${HOME}/pr-regression-testing/logs/${log_prefix}-run-pr-eval-exit-code.txt"
```
sami-marreed, line 113: Full stdout/stderr lands under `$HOME` on a persistent runner and is uploaded with no redaction. The next job (or anyone on the box) can read it. `cancel-in-progress: true` can also leave AppWorld on `:8000`/`:9000` half-running.

Write logs only under `$GITHUB_WORKSPACE`, redact known secret values, and drain leftover processes in `if: always()`.

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
...
+
+          bash "${HOME}/pr-regression-testing/cuga-eval/scripts/pr-regression-testing/run-pr-regression-eval.sh" \
+            "${COMMENT_BODY}" \
+            > "${output_log}" 2>&1
+
```
sami-marreed: Quoted here, so this layer is not bash injection — but safety is entirely `run-pr-regression-eval.sh`, which is not in this PR. If that script concatenates `$1` into a shell, this is RCE by an allowlisted account.

Parse command / provider / sha in the workflow and pass only those flags. Keep the script in-repo.

src/cuga/backend/llm/models.py
```diff
@@ -1342,6 +1342,10 @@ def _create_llm_instance(self, model_settings: Dict[str, Any]):
             rits_params: Dict[str, Any] = {
                 "api_key": "dummy" if api_key else None,
                 "base_url": base_url,
+                # RITS parameters to support eval implementation.
+                "openai_api_key": "/",
+                "openai_api_base": base_url,
+                "model_name": model_name,
```
sami-marreed, line 1348: **Critical:** These aliases duplicate fields already in `rits_params`:

- `openai_api_key: "/"` vs `api_key: "dummy" if api_key else None`
- `openai_api_base: base_url` vs `base_url: base_url`
- `model_name: model_name` vs `model: model_name`

`ChatOpenAI` uses `populate_by_name=True`, so passing both names typically raises `ValidationError` and breaks all RITS construction, not just eval. `openai_api_key: "/"` also fights the dummy Bearer key that exists so the real key is only sent as `RITS_API_KEY`.

Do not pass both names. If eval needs `"/"`, change the single dummy value and add a `tests/unit/` test that patches `langchain_openai.ChatOpenAI` and asserts kwargs.

sami-marreed (changes requested):
Thanks for the `/run-pr-eval` workflow — concurrency, SHA checkout, `persist-credentials: false`, and provider-scoped secrets are the right shape.

This is still a privileged `issue_comment` job that checks out untrusted PR (including fork) code, injects LLM secrets, and leaves `GITHUB_TOKEN` write access in the same process on a persistent self-hosted runner. `persist-credentials: false` does not remove the token from the step env. CodeQL [218](https://github.com/cuga-project/cuga-agent/security/code-scanning/218) is a real finding. CodeRabbit’s “do not add workflow YAML” note is not — AGENTS.md is about test-shard discovery in `tests.yml`, not a ban on new workflows.

Please treat these as blockers:

1. Split jobs: eval with `permissions: {}` / empty `GITHUB_TOKEN` and no PR write; a trusted job posts the sanitized report from artifacts.
2. Refuse fork heads unless an `environment:` approval gate is required (same pattern as `external-fork-tests.yml`).
3. Pin the evaluated SHA in the comment (`/run-pr-eval sha=<40-char>`) and fail if checkout does not match.
4. Do not run this on a persistent self-hosted runner. Vendor the eval script in-repo (or pin a checksum) instead of `$HOME/pr-regression-testing/...`.
5. Drop the extra RITS `ChatOpenAI` aliases in `models.py` — they collide with existing fields and can break all RITS construction.

`provider=litellm` also does not switch `AGENT_SETTING_CONFIG` or the env names LiteLLM actually reads. Happy to re-review once the isolation and RITS kwargs are fixed.

.github/workflows/run-pr-eval.yml
sami-marreed: Let's discuss how to solve this together

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
...
+    if: >
+      github.event.issue.pull_request &&
+      contains(github.event.comment.body, '/run-pr-eval') &&
+      (
+        github.event.comment.author_association == 'OWNER' ||
+        github.event.comment.user.login == 'AnkitaNaik' ||
+        github.event.comment.user.login == 'Sergey-Zeltyn' ||
+        github.event.comment.user.login == 'haroldship'
+      )
+    runs-on: [self-hosted, linux, run-pr-eval]
```
sami-marreed, line 28: **Critical:** Persistent self-hosted runners plus untrusted PR code is the GitHub-documented pwn path. The eval script also lives outside the repo at `$HOME/pr-regression-testing/.../run-pr-regression-eval.sh`, so a previous run can rewrite it.

Use ephemeral runners and vendor (or checksum-pin) the script in-repo. Treat `$HOME` as untrusted after any eval.

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
...
+          script: |
+            const pr = await github.rest.pulls.get({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.issue.number,
+            });
+
+            core.setOutput("head_repo", pr.data.head.repo.full_name);
+            core.setOutput("head_ref", pr.data.head.ref);
+            core.setOutput("head_sha", pr.data.head.sha);
```
sami-marreed, line 45: **Critical:** The slash-command authorizes a commenter; this step then fetches whatever `head.sha` is current. A force-push between the comment and `pulls.get` evaluates a different commit.

Require `/run-pr-eval sha=<40-char>` and fail if `git rev-parse HEAD` does not match. Also refuse `head.repo.full_name != github.repository` unless an `environment:` gate is required.

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
...
+      - name: Verify PR code
+        shell: bash
+        run: |
+          echo "Workspace: ${GITHUB_WORKSPACE}"
+          echo "PR SHA:"
+          git rev-parse HEAD
+
+          echo "CUGA source:"
+          if [[ -d "${GITHUB_WORKSPACE}/src/cuga" ]]; then
+            ls -la "${GITHUB_WORKSPACE}/src/cuga" | head
```
sami-marreed, line 63: This only `ls`s. It does not check SHA equality, same-repo origin, or a signature.

Replace with a hard fail if `HEAD` ≠ the SHA from the comment (and, if you keep forks, if `head.repo` is not this repository).

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
...
+        id: eval-provider
+        shell: bash
+        env:
+          COMMENT_BODY: ${{ github.event.comment.body }}
+        run: |
+          provider="rits"
+          normalized_comment="${COMMENT_BODY//$'\r'/ }"
+          normalized_comment="${normalized_comment//$'\n'/ }"
+
+          for token in ${normalized_comment}; do
```
sami-marreed, line 78: `${normalized_comment}` is unquoted (globbing), and `provider=` is written to `GITHUB_OUTPUT` with no allowlist.

Restrict to `{rits,litellm}` and fail closed, then `printf 'provider=%s\n' "$provider" >> "$GITHUB_OUTPUT"`.

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
...
+          echo "provider=${provider}" >> "${GITHUB_OUTPUT}"
+
+      - name: Run AppWorld evaluation
+        id: evaluation
+        shell: bash
+        env:
+          OPENAI_API_KEY: ${{ steps.eval-provider.outputs.provider == 'litellm' && secrets.OPENAI_API_KEY || '' }}
+          RITS_API_KEY: ${{ steps.eval-provider.outputs.provider == 'rits' && secrets.RITS_API_KEY || '' }}
+          LITE_LLM_KEY: ${{ steps.eval-provider.outputs.provider == 'litellm' && secrets.LITE_LLM_KEY || '' }}
+          AGENT_SETTING_CONFIG: settings.rits.toml
```
sami-marreed, line 96: **Important:** `provider=litellm` injects `LITE_LLM_KEY` / `OPENAI_API_KEY` but `AGENT_SETTING_CONFIG` is always `settings.rits.toml`. LiteLLM in `models.py` reads `OPENAI_API_KEY` / `LITELLM_API_KEY`, not `LITE_LLM_KEY`, and still needs the RITS key which this branch does not inject.

Map provider → config and env names, or drop the claimed litellm switch until it works.

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
...
+
+          mkdir -p "${HOME}/pr-regression-testing/logs"
+          run_started_at="$(date -u +"%Y%m%dT%H%M%SZ")"
+          short_sha="${PR_HEAD_SHA:0:12}"
+          log_prefix="pr-${PR_NUMBER}-${short_sha}-${run_started_at}"
+          output_log="${HOME}/pr-regression-testing/logs/${log_prefix}-run-pr-eval-output.md"
+          exit_code_log="${HOME}/pr-regression-testing/logs/${log_prefix}-run-pr-eval-exit-code.txt"
+
+          bash "${HOME}/pr-regression-testing/cuga-eval/scripts/pr-regression-testing/run-pr-regression-eval.sh" \
+            "${COMMENT_BODY}" \
```
sami-marreed, line 116: Quoted here, so this layer is not bash injection — but safety is entirely `run-pr-regression-eval.sh`, which is not in this PR. If that script concatenates `$1` into a shell, this is RCE by an allowlisted account.

Parse command / provider / sha in the workflow and pass only those flags. Keep the script in-repo.

sami-marreed (commented):
Relocated six inline comments onto the exact lines they refer to. The Changes requested review still stands.

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
...
+        github.event.comment.author_association == 'OWNER' ||
+        github.event.comment.user.login == 'AnkitaNaik' ||
+        github.event.comment.user.login == 'Sergey-Zeltyn' ||
+        github.event.comment.user.login == 'haroldship'
+      )
+    runs-on: [self-hosted, linux, run-pr-eval]
+    timeout-minutes: 120
+
+    steps:
+      - name: Get PR head
```
haroldship, line 32: **High:** We check who wrote `/run-pr-eval`, then later we ask GitHub for “the current PR head.” In that gap someone can push a different commit, and we run code the commenter never signed off on.

That race is called TOCTOU (time-of-check to time-of-use). Same finding as Sami and CodeQL.

Fix: the comment must name the commit, e.g. `/run-pr-eval sha=<40-char>`. If it is missing or is not a commit on this PR, fail. Do not fall back to live HEAD.

```suggestion
      - name: Parse requested SHA
        id: requested-sha
        shell: bash
        env:
          COMMENT_BODY: ${{ github.event.comment.body }}
        run: |
          sha=""
          for token in ${COMMENT_BODY}; do
            case "${token}" in
              sha=*) sha="${token#sha=}" ;;
            esac
          done
          if [[ ! "${sha}" =~ ^[0-9a-f]{12,40}$ ]]; then
            echo "::error::/run-pr-eval requires sha=<commit>. Refusing to guess current HEAD."
            exit 1
          fi
          echo "sha=${sha}" >> "${GITHUB_OUTPUT}"

      - name: Get and verify PR head
        id: pr
```

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
...
+
+            core.setOutput("head_repo", pr.data.head.repo.full_name);
+            core.setOutput("head_ref", pr.data.head.ref);
+            core.setOutput("head_sha", pr.data.head.sha);
+
+      - name: Checkout PR head commit
+        uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
+        with:
+          repository: ${{ steps.pr.outputs.head_repo }}
+          ref: ${{ steps.pr.outputs.head_sha }}
```
haroldship, line 51: **High:** This is where that SHA is used. Same issue as the step above (and Sami + CodeQL): do not check out `pr.data.head.sha`. Check out only the SHA from the comment, after you have confirmed it is on this PR.

Also: `head.repo` can be a fork. Our fork tests wait for a human (`environment: external-testing`) before secrets. This workflow skips that and still injects API keys. Only evaluate same-repo PRs, or require that approval gate for forks.

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
+name: Run PR Evaluation
```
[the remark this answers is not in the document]
haroldship: @coderabbitai The cited guideline (AGENTS.md, "CI discovers tests by directory — do not edit workflow YAML") doesn't apply here.

Read past the heading: every rule under it (AGENTS.md lines 20-41) is scoped to `tests.yml` and `external-fork-tests.yml` — specifically, not hardcoding individual test file paths into those two workflows' shard lists. It says nothing about adding a new, unrelated workflow file. `run-pr-eval.yml` has nothing to do with test discovery or sharding.

This PR has a real problem — a TOCTOU gap in how the workflow resolves the checkout SHA after its authorization check, combined with `issue_comment` giving it full permissions and secrets regardless of who opened the PR (see the general comment on this PR and the CodeQL finding above). That's the actual reason this shouldn't merge as-is. The workflow-YAML guideline isn't it.

.github/workflows/run-pr-eval.yml
haroldship: **High:** This job both runs the PR’s code and keeps a GitHub token that can write comments and edit the PR. Turning off git credentials does not hide that token from the eval step.

Fix: run the eval in a job with no write permission and an empty `GITHUB_TOKEN`. A second job, which never checks out the PR, should post the result.

.github/workflows/run-pr-eval.yml
haroldship: **High:** This runs on a machine that is reused. If the PR’s code breaks out, it can stay on that machine and change later jobs. The script under `$HOME` is also outside the repo, so an old run could replace it.

Fix: use a fresh runner each time, and keep the eval script in this repo (or pin it with a checksum).

.github/workflows/run-pr-eval.yml
haroldship: Same TOCTOU gap as your comment above — and the same finding as my comments on lines 32/51 and the CodeQL comment on line 131. See the general PR comment for the consolidated fix.

.github/workflows/run-pr-eval.yml
haroldship: Same underlying gap as your comment above (and the TOCTOU finding on lines 32/45/51/131) — this step doesn't check for it either. See the general PR comment for the consolidated fix.

.github/workflows/run-pr-eval.yml
haroldship: **Medium:** The comment is split with an unquoted loop, so `*` can match files in the PR checkout. A file named `provider=litellm` could switch which secrets we inject. We also accept any provider string.

Fix: only allow `rits` or `litellm`, then write the output with a safe `printf`.

.github/workflows/run-pr-eval.yml
haroldship: **Medium:** The full log is saved under the runner user’s home (it survives the job) and uploaded with no redaction. This repo is public, so anyone can download the artifact. If a key shows up in the log, it is leaked.

Canceling a run can also leave AppWorld listening on `:8000` / `:9000`.

Fix: write logs only in the job workspace, strip known secrets, and kill leftover processes when the job ends.

.github/workflows/run-pr-eval.yml
haroldship: **Medium:** We pass the whole comment to a script that is not in this PR. If that script pastes the comment into a shell command, a trusted commenter can run arbitrary commands.

Fix: parse the command, provider, and sha in the workflow. Pass only those values. Keep the script in the repo.

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
...
+
+          echo "Saved output log: ${output_log}"
+          echo "Saved exit code log: ${exit_code_log}"
+
+          cp "${output_log}" "${GITHUB_WORKSPACE}/run-pr-eval-output.md"
+          cp "${exit_code_log}" "${GITHUB_WORKSPACE}/run-pr-eval-exit-code.txt"
+
+          exit "${status}"
+
+      - name: Upload full evaluation output
```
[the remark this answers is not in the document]
haroldship: CodeQL is right — this is the same race Sami flagged on line 45, and the same as my comments on lines 32/51. See the general PR comment for the consolidated fix.

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
...
+        github.event.comment.user.login == 'Sergey-Zeltyn' ||
+        github.event.comment.user.login == 'haroldship'
+      )
+    runs-on: [self-hosted, linux, run-pr-eval]
+    timeout-minutes: 120
+
+    steps:
+      - name: Get PR head
+        id: pr
+        uses: actions/github-script@v7
```
haroldship, line 34: **Medium:** `actions/checkout` is pinned to a commit. `actions/github-script@v7` (here and in the “Post result” step) and `actions/upload-artifact@v4` are not. If those tags move, this write-enabled workflow would run different code.

Fix: pin both to full commit SHAs, same as checkout.

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
...
+          LITE_LLM_KEY: ${{ steps.eval-provider.outputs.provider == 'litellm' && secrets.LITE_LLM_KEY || '' }}
+          AGENT_SETTING_CONFIG: settings.rits.toml
+          OPENAI_BASE_URL: ${{ vars.OPENAI_BASE_URL }}
+          MODEL_NAME: ${{ vars.MODEL_NAME }}
+          ENVIRONMENT_URL: http://127.0.0.1:8000
+          APIS_URL: http://127.0.0.1:9000
+          PR_NUMBER: ${{ github.event.issue.number }}
+          PR_HEAD_SHA: ${{ steps.pr.outputs.head_sha }}
+          COMMENT_BODY: ${{ github.event.comment.body }}
+          PYTHONPATH: ${{ github.workspace }}/src
```
haroldship, line 104: **High:** This job puts the PR’s `src/` on `PYTHONPATH` and loads `settings.rits.toml` from that tree, with `RITS_API_KEY` in the environment. That file chooses the model URL. A PR can change `url` to an attacker server and the key goes there on the first model call — no extra malware needed. `apikey_name` can also point at another secret on the runner.

The workflow does not set `RITS_BASE_URL`, so the TOML URL wins.

Fix: pin the URL and key name from workflow vars. Do not let the PR TOML choose them.

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
...
+            const reportPattern = new RegExp(
+              `${escapedStart}\\s*([\\s\\S]*?)\\s*${escapedEnd}`,
+              "g"
+            );
+
+            const reportBlocks = [...output.matchAll(reportPattern)]
+              .map((match) => match[1].trim())
+              .filter(Boolean);
+
+            if (reportBlocks.length > 0) {
```
haroldship, line 175: **Medium:** We post whatever text sits between `REPORT START` and `REPORT END`. The PR’s code writes that output, so it can print a fake “all tests passed” report. People will trust the bot comment.

Fix: do not treat those markers as proof. Post only the exit code plus an artifact link, or sign the report with a secret the PR code cannot see.

.github/workflows/run-pr-eval.yml
```diff
@@ -0,0 +1,202 @@
...
+            if (output.length > 60000) {
+              output = `${output.slice(0, 60000)}\n\n...output truncated...`;
+            }
+
+            const result = exitCode === "0" ? "SUCCESS" : `FAILED (${exitCode})`;
+            const body = [
+              "## PR Evaluation",
+              "",
+              `- Triggered by: @${context.payload.comment.user.login}`,
+              `- Command: \`${context.payload.comment.body}\``,
```
haroldship, line 190: **Medium:** We paste the trigger comment and the report as markdown. A crafted comment or fake report can break out of the backticks, ping people, or add links that look official.

Fix: escape the comment text and treat the report as untrusted.

haroldship (changes requested):
This is a real security gap. It is not a false positive.

The `if:` condition checks who wrote the comment. After that check, the `Get PR head` step calls `pulls.get` and gets a new SHA for checkout.

This is a time-of-check/time-of-use (TOCTOU) gap. After an authorized comment, a person can change the PR branch before the job starts. They can force-push the branch or add commits. The workflow then checks out and runs the new HEAD. It does not run the code that the commenter reviewed.

The checkout step uses `head_sha`. This stops a change to the branch ref during checkout. It does not close the gap. The workflow still gets the SHA after the comment check. The “Verify PR code” step only lists files. It does not check that `HEAD` matches a SHA from the comment.

The workflow uses the `issue_comment` event. That event runs from the default branch, so the job token is privileged. This job can write issues and pull requests (`issues: write`, `pull-requests: write`). It also injects LLM secrets (`RITS_API_KEY`, `OPENAI_API_KEY`, `LITE_LLM_KEY`). `persist-credentials: false` only stops git from storing the token. It does not remove `GITHUB_TOKEN` from the eval process.

A standard `pull_request` workflow from a fork is different. It gets a read-only token. It does not get secrets.

This workflow does not do that. It checks out `head.repo`, which can be a fork. Fork CI in this repo waits for a human (`environment: external-testing`) before secrets. This job skips that gate. After an allowlisted person comments `/run-pr-eval`, a PR author can swap the branch and run their code with those credentials and the write token.

The same job then imports that PR tree (`PYTHONPATH`) and loads `settings.rits.toml` from it. That file chooses the model URL. The workflow does not set `RITS_BASE_URL`, so the TOML URL wins. A PR can point `url` at an attacker server. The first model call sends `RITS_API_KEY` there. `apikey_name` can also point at another secret on the runner. No extra malware is required.

It also runs on a reused self-hosted runner. The eval script lives outside the repo at `$HOME/pr-regression-testing/...`. If one run breaks out, it can persist on the machine or replace that script. Full stdout/stderr is written under `$HOME` and uploaded with no redaction. This repo is public, so anyone can download the artifact.

The bot comment is also untrusted output. We post whatever sits between `REPORT START` and `REPORT END`, plus the trigger comment as markdown. PR code can fake a passing report, break out of the backticks, or ping people.

Smaller holes in the same job: the comment is parsed with an unquoted loop (a file named `provider=litellm` can change which secrets we inject); the whole comment is passed to the out-of-repo script; `actions/github-script@v7` and `actions/upload-artifact@v4` are not pinned.

Fix these before this merges:

1. Pin the commit in the comment (`/run-pr-eval sha=<40-char>`). Confirm that SHA is on the PR. Fail if checkout does not match. Do not call `pulls.get` for a live HEAD after the comment check.
2. Refuse fork heads, or put an `environment:` approval gate in front of secrets (same idea as `external-fork-tests.yml`). Pin the SHA at approval time. Do not fetch the SHA again after approval.
3. Split the work into two jobs:
   - Eval job: may need LLM secrets if we call real models, but `permissions: {}` and `GITHUB_TOKEN: ""`. It must not write to the PR.
   - Post job: has `issues: write` / `pull-requests: write` only. It posts a sanitized report from artifacts. It must not check out PR code and must not get LLM secrets.
4. Do not run this on a persistent self-hosted runner. Vendor the eval script in-repo (or pin a checksum). Treat `$HOME` as untrusted after any eval.
5. Pin the model URL and key name from workflow vars. Do not let the PR TOML choose them.
6. Write logs only under `$GITHUB_WORKSPACE`. Redact secrets. Kill leftover processes in `if: always()`. Treat report markers and the trigger comment as untrusted. Allow only `provider=rits` or `provider=litellm`. Pass flags, not the raw comment, into the script. Pin `github-script` and `upload-artifact` to commit SHAs.

.github/workflows/run-pr-eval.yml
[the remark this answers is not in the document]
AnkitaNaik: Addressed. The workflow now requires /run-pr-eval sha=<40-char PR head SHA>, validates that SHA against the PR head, checks out that exact SHA, and hard-fails if git rev-parse HEAD does not match. This removes the previous live-head lookup behavior.

AnkitaNaik:
Thanks for the detailed review. I addressed the main security and correctness issues in the PR evaluation workflow.

The workflow now requires /run-pr-eval sha=<40-char PR head SHA>, validates the requested SHA, checks out only that commit, and verifies git rev-parse HEAD matches. Fork PR heads are refused for this credentialed path.

I split evaluation and reporting into separate jobs. The eval job has no PR write permissions and clears `GITHUB_TOKEN` for the evaluation step; the reporting job has the PR comment permissions but does not check out PR code. The workflow now validates provider/agent/benchmark/num_tasks/task parameters in trusted workflow code, allow lists provider to rits/litellm, and passes only the normalized command to the eval script.

Logs now stay under `$GITHUB_WORKSPACE`, known secrets are redacted before artifact upload, and cleanup runs with if: always() to stop leftover local services. github-script, upload-artifact, and download-artifact are pinned by commit SHA.

I also reverted src/cuga/backend/llm/models.py to match upstream main, so this PR no longer carries the RITS ChatOpenAI alias changes. The separate unit test for RITS kwargs was adjusted/kept only as applicable to the current provider behavior.

One intentional decision: the workflow still posts the inline evaluation report, because the report renderer is coming from the trusted `cuga-eval` checkout rather than PR-controlled cuga-agent code. The comment is generated from the uploaded artifact, includes the artifact link, and truncates oversized output. I understand the concern around untrusted markdown; if reviewers still prefer artifact-only reporting, I can switch that back, but I kept the inline report to preserve the existing PR evaluation UX.

The named-user allowlist is also intentionally unchanged for this initial setup as this needs to be discussed. I agree a repo variable/team-based authorization model would be cleaner long-term, but for this PR the command is gated to the currently trusted users only.

.github/workflows/run-pr-eval.yml
AnkitaNaik: @haroldship 

Recommended fix: make the PR-eval workflow use a trusted provider configuration instead of the PR-controlled `settings.rits.toml` for secret-routing fields.

The risk is that the evaluated PR can change `settings.rits.toml` to control where `RITS_API_KEY` is sent. For example, it could change the RITS `url` to an attacker-controlled endpoint, or change `apikey_name` to another environment variable available on the runner. Since the workflow puts the PR `src/` on `PYTHONPATH` and runs with `RITS_API_KEY` in the environment, this lets the PR exfiltrate the key through configuration alone.

The proposed fix is:

1. The workflow injects provider-specific trusted environment variables:
   - `RITS_API_KEY`
   - `RITS_BASE_URL`
   - `OPENAI_API_KEY`
   - `OPENAI_BASE_URL`

2. The eval wrapper selects the provider-specific config:
   - `provider=rits` -> trusted RITS config
   - `provider=litellm` -> trusted LiteLLM/OpenAI-compatible config

3. For `provider=rits`, the eval wrapper should not let the PR checkout decide the model endpoint or key env var. It should either:
   - generate a sanitized temporary RITS config on the runner, or
   - force `RITS_BASE_URL` and a fixed key name such as `RITS_API_KEY` before model initialization.

4. The RITS path should fail closed if the trusted values are missing:
   - fail if `RITS_API_KEY` is unset
   - fail if `RITS_BASE_URL` is unset

This gives a clearer contract:

- the workflow provides trusted secrets and URLs
- the eval script selects the provider configuration
- `models.py` receives the exact environment variables it reads
- PR-controlled TOML cannot redirect secrets to a different endpoint or choose a different key name

Caveat: this constrains model selection for PR evaluation. If the trusted config pins the RITS endpoint/model, changing the model used for PR evals would require updating the VM/workflow configuration rather than relying entirely on `/run-pr-eval model_name=...` command parameters. That tradeoff is intentional for PR evaluations because the model endpoint and credential-routing fields are security-sensitive.

AnkitaNaik:
@haroldship @sami-marreed 

> Do not run this on a persistent self-hosted runner. Vendor the eval script in-repo (or pin a checksum) instead of $HOME/pr-regression-testing/....

Agreed. The eval wrapper should not be treated as mutable runner state.

The intended final flow is to make `cuga-eval` a pinned dependency of the `cuga-agent` PR-eval workflow. The PR-regression wrapper lives in the paired cuga-eval PR (https://github.com/cuga-project/cuga-eval/pull/199), so that PR should be reviewed and merged first. After it is merged, the `cuga-agent` workflow/setup can record the merged cuga-eval commit SHA and verify the runner’s local cuga-eval checkout is exactly at that SHA before executing run-pr-regression-eval.sh.

That gives us a reproducible contract: cuga-agent controls which reviewed cuga-eval commit is allowed to run, and the self-hosted runner fails closed if its local checkout has drifted.

Issue #724, opened 2026-08-26
[Security] Remediate exception-detail leaks in server responses (49 baselined sites)


haroldship:
## Issue type

[Security]

## What you want and why

Near-term security slice split out of #722. #681 shipped the `error_responses.py` helpers
and #721 added a static check with a baseline of the existing offenders. This issue is the
burn-down: replace each baselined site that still puts exception text into a caller-facing
response with a fixed message plus a logged reference code.

This is the part of #722 that carries the actual security value, and it does **not** need
the typed `AppError` model — it just applies the helpers #681 already provides. Doing it
first closes the leaks now; the architectural model (#722) can follow.

## Scope

The 49 baselined sites in `scripts/checks/exc_in_responses_baseline.json`:

- `main.py` (40), `secrets_routes.py` (2), `llm_routes.py` (2), `a2a/router.py` (2),
  `config_routes.py` (1), `draft_routes.py` (1), `knowledge_routes.py` (1).

For each: choose a fixed caller message, log the exception behind a reference code via the
`error_responses.py` helpers (`safe_http_exception` / `safe_error_response` /
`safe_error_payload`), and remove its entry from the #721 baseline. Some `400`/`422`
client-input messages are intentional and stay (marked `# noqa: exc-in-response`) rather
than being rewritten.

## How it could work

- Base on #681 (helpers live there); expect several PRs, one per router, so each stays
  reviewable. `main.py` is the bulk and can be its own PR(s).
- Each fix is verified by the #721 checker: the site drops out of the baseline and no new
  finding appears.

## Out of scope

- The typed `AppError` hierarchy and central transport handlers — that is #722.

## Links or extra context

- Depends on #681 (helpers) and #721 (baseline as the burn-down list).
- Part of the #677 CodeQL hardening chore; follows #722's split.

## Checklist

- [x] I searched existing issues and this doesn't look like a duplicate

Issue #721, opened 2026-08-26
[Security] Static check: ban exception-derived values in server response sinks


haroldship:
## Issue type

[Security]

## What you want and why

Follow-up to #681, part of the #677 CodeQL hardening chore. #681 stopped the
CodeQL-flagged handlers from leaking internal exception detail by adding
`src/cuga/backend/server/error_responses.py`, whose helpers never copy anything out of
the exception. That fix relies on **caller discipline** — `message` defaults to a safe
generic string, yet nothing structurally stops a caller passing `message=str(exc)`; the
rule lives only in the docstrings. And #681 only touched the ~11 CodeQL-flagged sites: a
surface scan found **~190 other error-emission sites** it never touched that still put
exception text into responses (e.g. `main.py` conversation/config CRUD
`detail=f"Failed to ...: {str(e)}"`, `config_routes.py`/`draft_routes.py` `detail=str(e)`,
chat-request `422` with pydantic `e.errors()`).

A static check turns the #681 rule from convention into enforcement and surfaces the
untouched sites. Highest ROI, lowest cost — the concrete first step. The broader typed
error model is tracked separately in #722.

## How it could work

A lint/CI check that flags **exception-derived values reaching response sinks**:

- **Sinks:** `HTTPException(detail=...)`, `JSONResponse(...)`, SSE `data=` frames, and A2A
  `_rpc_error(...)`.
- **Banned values:** `str(exc)` / `repr(exc)` / `f"...{exc}..."`, `exc.args`,
  `traceback.format_exc()` / `traceback.*`, and pydantic `e.errors()` / `e.json()`.
- **Mechanism:** a ruff custom rule, an `ast-grep` rule, or a targeted CI grep — **not** a
  CodeQL config entry (default setup does not read the repo config, and
  `py/stack-trace-exposure` does not honor sanitizer models, so a config entry would be
  silently ignored, cf. #633).
- **Triage / allowlist:** some `str(e)` at `400`/`422` client-input sites are intentional
  validation feedback. Those get an explicit allowlist or route through a typed carve-out
  rather than being rewritten. This per-site judgement is exactly why it was kept out of
  #681.
- Wire it into the existing lint job so it runs on every PR.

## Links or extra context

- Enforces the rule established by #681; part of the #677 CodeQL chore.
- Broader typed error model: #722.
- Safe pattern reference: `src/cuga/backend/server/error_responses.py`.
[...]

Issue #217, opened 2026-05-13
feat(cua): CUAAgent — standalone orchestrator with typed async event stream


segevshlomovIBM:
Tracking issue: #209

### Scope

A standalone wrapper for the CUA loop that doesn't require booting LangGraph. Consumed by demos, MCP server wrappers, notebooks, and any caller that wants streaming UI events.

**Files added:**
- `nodes/computer_use/cua_agent.py` — `UseCaseConfig` dataclass + `CUAAgent` class + typed `@dataclass` events.

### Public surface

- `CUAAgent(executor, use_cases).run(initial_screen, utterance)` — async generator over the full multi-phase loop.
- `CUAAgent.run_phase(state, task, config=...)` — single-phase variant for tests / custom orchestration. Accepts `config` explicitly to avoid race conditions on concurrent `run()` calls.
- Event types: `CUARunStartEvent`, `CUAPhaseBannerEvent`, `CUAPhaseStartEvent`, `CUAStepStartEvent`, `CUADecisionEvent`, `CUAActionResultEvent`, `CUAConcludeEvent`, `CUAStepCapEvent`, `CUAPhaseCompleteEvent`, `CUAErrorEvent`.

### Acceptance criteria

- [ ] `CUAAgent` reuses the *same* `CUAPlannerAgent` / `CUAAction` / `CUASetup` implementations as the LangGraph path — no duplicated step logic.
- [ ] All exceptions yield a `CUAErrorEvent` before re-raising — callers always get error context.
- [ ] Both `AgentState(**cmd.update)` reconstructions are guarded.
- [ ] Concurrent `run()` calls on a shared instance do not race (config flows through method parameters, not instance vars).
- [ ] Unit-test coverage of `_resolve()` keyword matching and error paths.

### Depends on

- CUA-1 (`CUASetup`, `CUAAction`)
- CUA-2 (`CUAPlannerAgent`)

### Design reference

See section 5.5 of the epic design (#209).

PR #769, opened 2026-09-09
fix(security): stop leaking exception text in the events layer

haroldship:
## Bug fix

Related: #724 (the tracked burn-down for this rule; its 49-site baseline predates these files,
so this PR is not in its listed scope, but closes the same class of leak in the files #724
didn't cover yet)

### Summary

13 of the 14 currently open CodeQL alerts on `main` are `py/stack-trace-exposure` (CWE-209):
an exception's text — which can carry a stack trace, file paths, or other server internals —
reaches an HTTP caller. All 13 trace back to `src/cuga/backend/events/app.py` and
`src/cuga/backend/server/events_bridge.py`, both added by #602 after #681 established the
"never put exception text in a response" rule for this codebase (`error_responses.py`) and
#721 added a static check for it. Neither file picked up either one.

- **`events_bridge.py`** (alert #217's root cause — CodeQL traces the leak here, not to
  `run_routes.py` where the alert is anchored): `forward_slash_to_events` returned
  `f"...({e}). Nothing was armed."` straight to the chat user on a down eventing service. Now
  uses `error_responses.log_error_ref` — the exception goes to the log behind a reference code,
  the reply text stays fixed.
- **`events/app.py`** (alerts #184, #185, #186, #187, #188, #189, #190, #191, #192, #193, #194,
  #196 — 12 of the 14 open alerts, all the same `except Exception as e: ... JSONResponse({"error":
  str(e)/f"...{e}"...})` shape): fixed the same way, but without importing `error_responses` —
  `events/` is deliberately decoupled from `cuga.backend.server` (one seam only, `resolve_secret`;
  enforced by `test_core_decoupled.py`), so each site now returns a fixed message plus the
  request's own `trace_id` as the reference code instead. Two more sites used `repr(e)`, which
  CodeQL didn't flag but is the same rule — fixed alongside the 12.
- **`no_exc_in_responses.py`** (#721's static check): its default scan target was
  `src/cuga/backend/server` only, so it never looked at `events/` — that's how 14 new leaks
  landed there unseen between #721 and today. Widened the default to cover `events/` too.

### Testing
- [x] `uv run python scripts/checks/no_exc_in_responses.py` — 0 findings (was 14 in `events/`)
- [x] `uv run ruff check && uv run ruff format --check` — clean
- [x] `uv run pytest tests/events/ tests/unit/test_error_responses.py tests/unit/test_no_exc_in_responses.py` — 481 passed, 27 skipped (Postgres-only)
- [x] Updated two tests that had asserted on the leaked exception text as the expected response body (`test_concierge_error_is_500_with_trace_id`, `test_connect_token_ap_failure_is_500_not_a_crash`)

Alert links: [#217](https://github.com/cuga-project/cuga-agent/security/code-scanning/217), [#184](https://github.com/cuga-project/cuga-agent/security/code-scanning/184), [#185](https://github.com/cuga-project/cuga-agent/security/code-scanning/185), [#186](https://github.com/cuga-project/cuga-agent/security/code-scanning/186), [#187](https://github.com/cuga-project/cuga-agent/security/code-scanning/187), [#188](https://github.com/cuga-project/cuga-agent/security/code-scanning/188), [#189](https://github.com/cuga-project/cuga-agent/security/code-scanning/189), [#190](https://github.com/cuga-project/cuga-agent/security/code-scanning/190), [#191](https://github.com/cuga-project/cuga-agent/security/code-scanning/191), [#192](https://github.com/cuga-project/cuga-agent/security/code-scanning/192), [#193](https://github.com/cuga-project/cuga-agent/security/code-scanning/193), [#194](https://github.com/cuga-project/cuga-agent/security/code-scanning/194), [#196](https://github.com/cuga-project/cuga-agent/security/code-scanning/196)

These are code-scanning alerts, not issues — they do not auto-close on merge via `Fixes #n`. They close on their own once CodeQL re-scans the merged fix and no longer detects the pattern.

## Summary by CodeRabbit

* **Bug Fixes**
  * Error responses no longer expose raw exception details or stack traces.
  * Responses now provide a reference ID for troubleshooting while preserving existing status codes and formats.
  * Error checks now cover both server and events API areas.
  * Improved error handling across events and server-to-events requests provides more consistent sanitized responses.


haroldship:
## Fixes Applied Successfully

Fixed 2 file(s) based on 2 CodeRabbit feedback item(s).

**Files modified:**
- `src/cuga/backend/events/app.py`
- `tests/events/test_events_api_contract.py`

**Commit:** `42b8c9f5`

The latest autofix changes are on the `fix/codeql-stack-trace-exposure` branch.

offerakrabi (approved):
## PR Review: #769 — fix(security): stop leaking exception text in the events layer

only one minor finding, it's your choice whether to fix or defer.

one more thing, do you know if we currently use those traces from http requests anywhere? @sami-marreed maybe you know?
if we are using them somewhere we should at the very least update the documentation about how you can still get them

**Merge.** The fix is correct and complete: all 14 CodeQL sites are remediated, the decoupling constraint is respected throughout, and the static check now covers `events/` so the gap cannot reopen silently.
Nothing blocks. #1 is the author's call.

|---|---|---|---|---|
| 1 | Minor | `/invoke` failure paths lack "no exception text" test | `tests/events/test_events_api_contract.py`, `src/cuga/backend/events/app.py:406-419,448-461` | The two busiest error paths in `app.py` (concierge and worker agent failures under `/invoke`) have no test asserting the exception text stays out of the response body; a future regression would be caught only by the static check, not a contract test. |

Issue #766, opened 2026-09-09
[Bug]: Playbook with empty keyword list fires on every message (AND with no keywords always matches)


offerakrabi:
## What happened

A playbook with a `KeywordTrigger` whose `value` list is empty and `operator` is `"and"` matches **every single user message** with confidence 1.0. This is the default state of a newly created playbook in the UI — clicking "Add Playbook" and saving without entering any keywords produces exactly this trigger.

The playbook's guidance gets injected into every agent response regardless of context.

## Root cause

In `src/cuga/backend/cuga_graph/policy/agent.py`, the AND keyword match condition is:

```python
if len(matched_keywords) == len(keywords):  # AND: match if ALL found
    return True, 1.0, ...
```

When `keywords = []`, this evaluates to `0 == 0 → True`.

The UI creates new playbooks with this default trigger (`PoliciesConfig.tsx`):

```ts
triggers: [{ type: "keyword", value: [], target: "intent", operator: "and" }]
```

## How to reproduce

1. Start the server in any mode.
2. Open the UI → Policies → Add Playbook → save without adding any keywords.
3. Send any message to the agent — the playbook fires on every message with confidence 1.0.

Or verify directly:

```python
keywords = []
matched = [kw for kw in keywords if kw in "hello"]
print(len(matched) == len(keywords))  # True
```

## Proposed fix

Add a guard in `_check_trigger` in `agent.py` before the AND comparison:

```python
# AND: Match if ALL keywords are found
if not keywords:
    return False, 0.0, "No keywords configured in trigger"
if len(matched_keywords) == len(keywords):
    ...
```

## Environment

- All server modes
- Any playbook created via the UI without entering keywords

## Checklist

- [x] I searched existing issues and this doesn't look like a duplicate

Issue #765, opened 2026-09-09
[Bug]: Playbooks not written to .cuga/playbooks/ when saved via the UI (filesystem sync not propagated to draft state)


offerakrabi:
## What happened

When `filesystem_sync = true` and `cuga_folder = ".cuga"` (the defaults in `settings.toml`), saving a playbook via the UI writes it to the policy DB (SQLite WAL) but never creates the corresponding `.md` file in `.cuga/playbooks/`. The filesystem and DB are permanently out of sync. On the next server restart the playbook is re-loaded from the empty folder (or not at all), not from the DB.

Note: this only surfaces when running the raw server (`uvicorn cuga.backend.server.main:app`). All CLI modes (`cuga start demo`, `demo_crm`, `manager`, etc.) set `DYNACONF_POLICY__FILESYSTEM_SYNC=false` which masks the issue.

## Root causes

Three places in `src/cuga/backend/server/main.py` independently drop the `PolicyFilesystemSync` object:

1. **Startup — `_init_policy` / `lifespan`** (~line 618): `PolicyFilesystemSync` is only created when `.cuga/` already exists at startup. A fresh install with no `.cuga/` folder gets `app_state.policy_filesystem_sync = None`, so any policy saved afterward is never written to disk.

2. **Draft state init** (~line 1068): `draft_app_state.policy_filesystem_sync` is never assigned after the draft policy system is created. It stays `None` permanently. All UI saves go through the draft route (`PATCH /api/manage/config/draft/policies`), which passes `state.policy_filesystem_sync` to `apply_policies_data_to_storage` — but that value is always `None`.

3. **`POST /api/config/policies` draft path** (~line 3333): The draft branch of `save_policies_config` hardcodes `filesystem_sync=None`, ignoring the sync object entirely.

## Proposed fix

**Fix 1:** In the startup block, always create `PolicyFilesystemSync` when `filesystem_sync_enabled` is true, before checking whether `.cuga/` exists:

```python
# Before (only created if folder exists):
elif os.path.exists(cuga_folder):
    ...
    app_state.policy_filesystem_sync = PolicyFilesystemSync(cuga_folder=cuga_folder)

# After (always created when enabled):
else:
    app_state.policy_filesystem_sync = PolicyFilesystemSync(cuga_folder=cuga_folder)
    if auto_load_enabled and os.path.exists(cuga_folder):
        ...  # load from folder
```

**Fix 2:** After initialising the draft policy system, propagate the sync object:

```python
[...]

PR #773, opened 2026-09-10
fix(policy): per-agent policy isolation — playbooks no longer bleed across agents (#767)

### LINKED ISSUE #767: [Bug]: Playbooks bleed across agents — all agents share the same policy collection with no per-agent isolation
Part of #75.

## What happened

All registered agents — including supervisor sub-agents — share the same policy collection (`cuga_policies` or `cuga_policies_draft`). There is no `agent_id` field in the policy storage schema and no per-agent filter at query time.

As a result, a playbook created for Agent A (e.g. a database query agent) will fire for Agent B (e.g. a weather agent) whenever a user message matches its trigger. The only existing isolation is tenant/instance scoping, which does not distinguish agents running within the same server process.

The workaround is to delete the playbooks — which removes them from the shared store entirely.

## Root cause

- `PolicyStoreBackend` (`local.py`, `prod.py`) stores and queries policies filtered by `tenant_id` and `instance_id` only. No `agent_id` column exists in the schema.
- The server initializes **one** `PolicyConfigurable` singleton and one policy collection for all published agents (`main.py` ~line 598). Registered agents get their own `DynamicAgentGraph` and `CombinedToolProvider` keyed by `agent_id`, but all share `app_state.policy_system`.
- `PolicyConfigurable.from_config` resolves to the singleton unless a `policy_system` object is explicitly injected into the LangGraph config — which registered agents do not do today.

## How to reproduce

1. Start the server with two registered agents (e.g. a CRM agent and a docs agent).
2. Create a playbook scoped to the CRM agent (keywords: "customer", "account").
3. Send a message containing "account" to the docs agent.
4. The CRM playbook fires and injects CRM-specific guidance into the docs agent response.

### Option A — add `agent_id` to the policy storage schema

Add an `agent_id` metadata column to `PolicyStoreBackend` and filter on it in every add/get/list/search/delete operation. Tag each policy with the agent that owns it at write time, and pass `agent_id` through `PolicyContext` and `PolicyConfigurable.from_config`.

- ✅ Complete isolation with a single shared DB
- ❌ Requires a schema change + migration for both `local` (SQLite) and `prod` (Postgres/pgvector) backends, plus updates to every storage operation signature

### Option B — separate policy collection per agent (recommended)

Give each registered agent its own `PolicyStorage` instance using a unique collection name: `cuga_policies_<agent_id>`. The server already does exactly this for draft vs published (`cuga_policies_draft`). Extend `_resolve_stream_agent` (and the graph-cache builder) to construct a per-agent `PolicyConfigurable` backed by the agent-specific collection, and inject it into the graph config.

- ✅ No schema change or migration
- ✅ Natural extension of the existing draft/published isolation pattern
- ✅ Each agent's policies are fully isolated and independently manageable
- ✅ The `cuga_folder` per-agent filesystem path (already supported by `DynamicAgentGraph`) can be wired in at the same time to complete the isolation
- ❌ Slightly more memory (one `PolicyStorage` instance per agent) — negligible in practice

**We recommend Option B** as it achieves full isolation with minimal risk: no storage migration, no backend API changes, and it follows a pattern already established in the codebase.

## Checklist

offerakrabi:
## Bug fix

Fixes #767, #765, #766

### Summary

The core problem (#767): all registered agents shared the same policy collection in storage. A playbook created for Agent A would silently fire for Agent B whenever a user message matched its trigger keyword — there was no per-agent boundary at all.

This PR gives each agent its own isolated policy collection, and along the way fixes two smaller bugs (#765, #766) that were found while working through the policy system.

---

#### #767 — Per-agent policy isolation

**Root cause:** `_resolve_stream_agent` always handed every named agent the same `app_state.policy_system` singleton. All agents queried the same `cuga_policies` collection.

**Fix:** each agent now gets its own `PolicyStorage` instance backed by a uniquely named collection: `cuga_policies_<agent_id>` (e.g. `cuga_policies_crm_agent`). This follows the same pattern the codebase already used for draft vs published isolation (`cuga_policies_draft`).

Two helpers were added to `configurable.py`:
- `get_agent_policy_collection_name(agent_id, draft)` — derives the scoped collection name, keeping `cuga-default` on the existing `cuga_policies` collection so there's no migration.
- `create_agent_policy_system(agent_id, draft, policies_data)` — creates and initialises a `PolicyConfigurable` backed by the agent-scoped collection.

`_resolve_stream_agent` now calls `create_agent_policy_system(agent_id=...)` instead of reusing the singleton. Supervisor subagents are isolated through the same path via the updated `build_agents_from_stored_subagents`. The `GET` and `POST /api/config/policies` endpoints also route to the agent-scoped collection when `X-Agent-ID` is present.

#### #765 — Filesystem sync not propagated to draft state

Playbooks saved via the UI were stored in the DB but the corresponding `.md` file under `.cuga/playbooks/` was never written. Three gaps caused this: `PolicyFilesystemSync` was only created at startup if `.cuga/` already existed, `draft_app_state.policy_filesystem_sync` was never assigned, and the draft save path hardcoded `filesystem_sync=None`. All three are fixed.

#### #766 — Empty keyword trigger matched every message

A newly created playbook (no keywords entered yet) had an AND trigger with an empty list. `0 == 0` evaluated to `True`, so it fired on every message with confidence 1.0. A guard was added to return no-match immediately when the keyword list is empty.

---

### Testing

**Automated:**
- `tests/unit/test_agent_policy_isolation.py` — 4 unit tests: collection name scoping, storage isolation between two agents, `_resolve_stream_agent` producing an agent-scoped system, and supervisor subagent isolation.
- `tests/unit/test_policy_filesystem_sync_propagation.py` — covers the startup sync initialisation path.
- `policy/tests/test_keyword_operator.py` — covers the empty-keyword guard and AND/OR operator behaviour.

**Manual (not in the PR):**

To verify end-to-end against a live server, `test_policy_isolation.py` (project root) was written and run. The script: creates two named agents (`crm-agent`, `docs-agent`), saves a keyword-triggered playbook scoped to `crm-agent` only, then streams the trigger keyword `"account"` to `docs-agent` and checks whether the CRM policy fires. It passed 3/3 consecutive runs — on the final run `docs-agent` returned a full live LLM response with 0 policies in its store and no policy bleed.

The filesystem sync fix was also verified manually: playbooks saved via the UI under `cuga start demo` now appear as `.md` files in `.cuga/playbooks/`, and playbooks created for named agents correctly do not appear there (their `filesystem_sync` is `None` by design).

## Summary by CodeRabbit

* **New Features**
  * Added agent-specific policy configurations with separate published and draft states.
  * Policy updates for named agents are applied when publishing or editing drafts.
  * Supervisor sub-agents can load the appropriate draft or published policies.
  * Policy endpoints support selecting policies by agent ID.

* **Bug Fixes**
  * Empty keyword triggers now fail safely without errors.
  * Policy filesystem synchronization initializes consistently and is shared with draft state.
  * Unknown or invalid agent IDs are rejected when accessing policy configurations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -566,6 +566,49 @@ Enabling it puts tool tracking in a **timings-only** mode unless you passed
 arguments, results or errors. Coverage matches `track_tool_calls` (registry/MCP
 tools and `@tracked_tool` functions).
 
+### Configuring the final answer
+
+Everything about the final answer lives behind four knobs — this table is the map:
+
+| What | Where | Notes |
+|---|---|---|
+| Which LLM composes it | `[agent.final_answer.model]` (model TOMLs) | existing |
+| Whether the composer runs | `features.final_answer` (mode TOMLs) | `fast` mode disables it |
+| Guidance for the composer | `[final_answer].instructions` | fills the composer prompt's instructions slot — **ignored in default Lite mode**; use `special_instructions` / SDK `final_answer="..."` there |
+| Deterministic shaping | `[final_answer].function` | dotted path to a pure `(str) -> str` — imported **and called** with the answer, so point only at trusted code |
+
+```toml
+[final_answer]
+instructions = "Answer with the bare value only — no units, no prose."
+function = "my_pkg.formats.bare"   # applied to the delivered answer, before citations
+```
+
+Or from the SDK:
+
+```python
+agent = CugaAgent(final_answer="Answer with the bare value only.")   # LLM guidance
+agent = CugaAgent(final_answer=my_fn)                                 # deterministic function
+agent = CugaAgent(final_answer=FinalAnswerConfig(instructions="...", function=my_fn))
+```
+
+The function contract: **pure and deterministic** (idempotent recommended),
+applied once per delivered answer, and it can never break delivery — any
+failure logs and returns the answer unchanged. It runs after harmony-token
+sanitization and before citation resolution; sanitization is not replaceable.
+`instructions` guide the `FinalAnswerAgent` composer (full pipeline); in the
+default CugaLite mode the answer is composed by CugaLite itself — use
+`special_instructions` there (the SDK `final_answer="..."` form does this for
+you). Deterministic contracts (e.g. benchmark answer shapes) belong in
+`function`, not `instructions` — prompts cannot guarantee an exact string.
+
+Scope notes: answers synthesized by a `CugaSupervisor` itself (as opposed to
+forwarded sub-agent answers) are not shaped in this version. The AppWorld
+benchmark path keeps its own answer pipeline — its SDK post-processing
+re-extracts the answer and overwrites `InvokeResult.answer`, so in-graph
+shaping does not survive there. The chat transcript deliberately carries the
+pre-finalize text — the function shapes `result.answer`, and is not a
+redaction mechanism.
+
 ### Knowledge Base
 
 CUGA includes a built-in knowledge base powered by LangChain and local vector stores. **Docling** is integrated for document ingestion: it parses and normalizes PDFs, Office files, HTML, Markdown, images, and other supported types before chunking and embedding, so the pipeline stays self-contained with no external document services.
--- a/scripts/checks/no_exc_in_responses.py
+++ b/scripts/checks/no_exc_in_responses.py
@@ -63,8 +63,10 @@ TRACEBACK_FUNCS = frozenset({"format_exc", "print_exc", "format_exception", "for
 # for a given expression — an `except ... as x` binding does not cross a function boundary.
 _SCOPE_BOUNDARIES = (ast.FunctionDef, ast.AsyncFunctionDef, ast.Lambda, ast.Module)
 
-# Default source tree to scan.
-DEFAULT_TARGET = "src/cuga/backend/server"
+# Default source trees to scan. `events/` builds its own responses the same way `server/`
+# does (JSONResponse/HTTPException), so it needs the same check — it was left out when this
+# check was introduced and 14 new offenders landed there unseen (see #602, #217).
+DEFAULT_TARGETS = ("src/cuga/backend/server", "src/cuga/backend/events")
 DEFAULT_BASELINE = "scripts/checks/exc_in_responses_baseline.json"
 
 
@@ -399,12 +401,12 @@ def _relpath(path: Path) -> str:
 
 def main(argv: list[str] | None = None) -> int:
     parser = argparse.ArgumentParser(description=__doc__)
-    parser.add_argument("paths", nargs="*", default=[DEFAULT_TARGET])
+    parser.add_argument("paths", nargs="*", default=list(DEFAULT_TARGETS))
     parser.add_argument("--baseline", default=DEFAULT_BASELINE)
     parser.add_argument("--update", action="store_true", help="regenerate the baseline from current sources")
     args = parser.parse_args(argv)
 
-    paths = args.paths or [DEFAULT_TARGET]
+    paths = args.paths or list(DEFAULT_TARGETS)
     by_file: dict[str, list[Violation]] = {}
     for file in _iter_py_files(paths):
         try:
--- a/src/cuga/__init__.py
+++ b/src/cuga/__init__.py
@@ -38,6 +38,8 @@ __all__ = [
     "KnowledgeClient",
     "KnowledgeEngine",
     "KnowledgeConfig",
+    # Final answer: LLM guidance + deterministic function (CugaAgent final_answer=).
+    "FinalAnswerConfig",
     # Tool shortlisting: how a large tool set is shrunk before the model sees it.
     "Shortlister",
     "ShortlisterStrategy",
@@ -54,6 +56,7 @@ __all__ = [
 # For type checkers and IDEs — not executed at runtime.
 if TYPE_CHECKING:
     from cuga.sdk import CugaAgent, CugaSupervisor, run_agent, InvokeResult
+    from cuga.backend.cuga_graph.nodes.answer.answer_function import FinalAnswerConfig
     from cuga.backend.cuga_graph.nodes.cuga_lite.tracking.tracker import tracked_tool
     from cuga.backend.cuga_graph.nodes.cuga_lite.shortlister import (
         Shortlister,
@@ -89,6 +92,10 @@ _import_map: dict = {
     "KnowledgeClient": ("cuga.backend.knowledge", "KnowledgeClient"),
     "KnowledgeEngine": ("cuga.backend.knowledge", "KnowledgeEngine"),
     "KnowledgeConfig": ("cuga.backend.knowledge.config", "KnowledgeConfig"),
+    "FinalAnswerConfig": (
+        "cuga.backend.cuga_graph.nodes.answer.answer_function",
+        "FinalAnswerConfig",
+    ),
     "Shortlister": (
         "cuga.backend.cuga_graph.nodes.cuga_lite.shortlister",
         "Shortlister",
--- a/src/cuga/backend/cuga_graph/nodes/answer/answer_function.py
+++ b/src/cuga/backend/cuga_graph/nodes/answer/answer_function.py
@@ -1,0 +1,97 @@
+"""Deterministic final-answer function — the user-owned step of finalize_answer.
+
+Configured by ``[final_answer].function`` (a dotted path to a pure
+``(str) -> str``) or injected as a live callable by the SDK
+(``CugaAgent(final_answer=my_fn)``). The injected callable wins.
+
+Contract for the user function: pure and deterministic; idempotent is
+recommended as a safety margin (the seam applies it once per delivered
+answer). It must never break answer delivery: any failure is logged and
+the answer is delivered unformatted.
+"""
+
+from dataclasses import dataclass
+from typing import Callable, Optional
+
+from loguru import logger
+
+from cuga.config import get_class, settings
+
+
+@dataclass
+class FinalAnswerConfig:
+    """SDK config for ``CugaAgent(final_answer=...)`` when combining both
+    halves: LLM guidance and a deterministic function."""
+
+    instructions: Optional[str] = None
+    function: Optional[Callable[[str], str]] = None
+
+
+def answer_function_configured() -> bool:
+    """Whether ``[final_answer].function`` names a function (non-empty str)."""
+    value = settings.final_answer.function
+    return isinstance(value, str) and bool(value.strip())
+
+
+def resolve_final_answer_instructions() -> str:
+    """``[final_answer].instructions`` from settings/env, stripped ('' when unset).
+
+    Non-string TOML values are treated as unset: config.py only warns on
+    validator failures, so this guard is what keeps answer delivery alive.
+    """
+    value = settings.final_answer.instructions
+    return value.strip() if isinstance(value, str) else ""
+
+
+def resolve_answer_function(path: str) -> Callable[[str], str]:
+    """Resolve a dotted ``module.attr`` path to a callable.
+
+    Raises with a clear message on a dotless path or a non-callable target.
+    (Callers on the answer path catch and log — a bad path is reported per
+    answer rather than crashing delivery.)
+    """
+    if "." not in path:
+        raise ValueError(f"[final_answer].function must be a dotted 'module.attr' path, got '{path}'")
+    fn = get_class(path)
+    if isinstance(fn, type):
+        # Classes are callable but calling one constructs an instance — the
+        # answer would silently never be formatted (non-str result is dropped).
+        raise TypeError(f"[final_answer].function '{path}' is a class; point it at a (str) -> str function")
+    if not callable(fn):
+        raise TypeError(f"[final_answer].function '{path}' is not callable")
+    return fn
+
+
+def apply_answer_function(state, answer_function: Optional[Callable[[str], str]] = None) -> None:
+    """Apply the configured answer function to ``state.final_answer`` in place.
+
+    No-op when the answer is empty or nothing is configured. Never raises.
+    """
+    if not getattr(state, "final_answer", None):
+        return
+    fn = answer_function
+    if fn is None:
+        configured = settings.final_answer.function
+        path = configured.strip() if isinstance(configured, str) else ""
+        if not path:
+            if configured and not isinstance(configured, str):
+                logger.warning(
+                    f"[final_answer].function is {type(configured).__name__}, expected str; ignoring"
+                )
+            return
+        try:
+            fn = resolve_answer_function(path)
+        except Exception:
+            logger.exception(f"final_answer.function {path!r} failed to resolve; delivering unformatted")
+            return
+    try:
+        formatted = fn(state.final_answer)
+    except Exception:
+        logger.exception("final answer function failed; delivering unformatted answer")
+        return
+    if isinstance(formatted, str):
+        state.final_answer = formatted
+    else:
+        logger.warning(
+            f"final answer function returned {type(formatted).__name__}, expected str; keeping original"
+        )
--- a/src/cuga/backend/cuga_graph/nodes/answer/final_answer.py
+++ b/src/cuga/backend/cuga_graph/nodes/answer/final_answer.py
@@ -1,11 +1,12 @@
 import json
-from typing import Literal, Dict, Callable
+from typing import Literal, Dict, Callable, Optional
 
 from langchain_core.messages import AIMessage
 from langgraph.types import Command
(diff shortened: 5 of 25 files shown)


offerakrabi:
@coderabbitai

Addressed in e3618ef8:

Finding 1+2 — IDOR / collection collision — Added registry validation in GET and POST /api/config/policies: unrecognised X-Agent-ID values return 404 before any collection name is derived. Updated get_agent_policy_collection_name docstring to document the slugified-ID assumption. Replaced the misleading finance.bot test assertion with an injectivity check.

Finding 3 — Lock-before-policy-replace race — patch_draft_policies for named agents now uses async with agent_draft_lock + save_draft_section_unlocked directly, keeping the lock held across both the config-store write and create_agent_policy_system. No window for a concurrent save to diverge.

Finding 4 — Supervisor stale overwrite — Dropped policies_data from create_agent_policy_system in build_agents_from_stored_subagents. Supervisor construction is now read-only; the test updated to assert len(policies) == 0 with an explanatory docstring.

Finding 5 — Unit tests may touch prod storage — Added monkeypatch.setattr(facade, "_storage_mode", lambda: "local") to the shared _isolate_local_storage_db autouse fixture in conftest.py, covering all unit tests regardless of settings.storage.mode.

Verification: uv run ruff format + uv run ruff check — all clean. tests/unit/ — 1688 passed, 3 pre-existing failures in test_sdk_citations (confirmed present on branch before our changes, unrelated to this PR).

sami-marreed:
Reviewed `ea64edc3638a5cc7448d01f90888ef114a5791ec`. I found four issues that should be fixed before merge:

| Priority | Finding |
|---|---|
| **P1** | **Collection names collide across agents.** `crm` draft and `crm-draft` published both map to `cuga_policies_crm_draft`. Both IDs are valid registry slugs. Saving the former overwrites the latter's live policies because storage replacement clears the shared collection. Use an unambiguous encoding of agent ID and draft status. [configurable.py:33–35](https://github.com/cuga-project/cuga-agent/blob/ea64edc3638a5cc7448d01f90888ef114a5791ec/src/cuga/backend/cuga_graph/policy/configurable.py#L33-L35) |
| **P1** | **A first stream can erase successfully saved policies.** `POST /api/config/policies` updates policy storage, but graph construction replaces it from the older config snapshot. Reproduction: persist a named-agent draft with `policies: []`, save a policy through `POST /api/config/policies` with `X-Agent-ID` and `X-Use-Draft`, then resolve the first draft stream. The save returns 200 and storage contains the policy before resolution; afterward storage is empty. Graph construction should preserve newer saves, and the save paths should keep config and policy storage consistent. [main.py:2563–2566](https://github.com/cuga-project/cuga-agent/blob/ea64edc3638a5cc7448d01f90888ef114a5791ec/src/cuga/backend/server/main.py#L2563-L2566) |
| **P2** | **Full draft saves leave supervisor subagents without policies.** `POST /api/manage/config/draft` persists named-agent policies only in config storage. The new supervisor loader reads the separate policy collection without populating it. Saving a full draft containing an intent guard and then building a draft supervisor referencing that agent produces a subagent with zero policies. Synchronize the named-agent collection in every supported save path before relying on it here. [supervisor_config.py:229–236](https://github.com/cuga-project/cuga-agent/blob/ea64edc3638a5cc7448d01f90888ef114a5791ec/src/cuga/supervisor_utils/supervisor_config.py#L229-L236) |
| **P2** | **Published-config fallback reads draft policies.** When a referenced subagent has no draft, its config falls back to published, but `create_agent_policy_system` still receives `draft=True`. With a populated published collection and an empty draft collection, the subagent runs without its published guards. Track which config was resolved and select the corresponding policy collection. [supervisor_config.py:210–213](https://github.com/cuga-project/cuga-agent/blob/ea64edc3638a5cc7448d01f90888ef114a5791ec/src/cuga/supervisor_utils/supervisor_config.py#L210-L213), [233–236](https://github.com/cuga-project/cuga-agent/blob/ea64edc3638a5cc7448d01f90888ef114a5791ec/src/cuga/supervisor_utils/supervisor_config.py#L233-L236) |

Validation: the 18 tests in `test_agent_policy_isolation.py`, `test_policy_filesystem_sync_propagation.py`, and `test_keyword_operator.py` passed locally with a placeholder OpenAI key for client construction. Four additional isolated regression checks reproduced the failures above. These checks exercised policy storage and route/loader behavior; they did not make live LLM calls.

sami-marreed (changes requested):
Requesting changes for the four issues documented in https://github.com/cuga-project/cuga-agent/pull/773#issuecomment-5624847438: collection-name collisions, stream construction erasing saved policies, missing supervisor policies after full draft saves, and published-config fallback reading draft policy storage. Reviewed commit ea64edc3638a5cc7448d01f90888ef114a5791ec; all four issues were reproduced with isolated regression checks.

offerakrabi:
Addressed in d30e9e68:

1. [Collection name collision — P1] — Changed draft suffix from `_draft` to
   `__draft` in get_agent_policy_collection_name. Agent 'crm-draft' published
   maps to `cuga_policies_crm_draft`; agent 'crm' draft maps to
   `cuga_policies_crm__draft`. Registry slugs are [a-z0-9-] so double
   underscore is injective. Test assertion updated + injectivity test added.

2. [Stream clears newer saves — P1] — Removed `policies_data=policies_list`
   from the create_agent_policy_system call in _resolve_stream_agent. Graph
   construction is now read-only w.r.t. policy storage, matching the existing
   supervisor subagent pattern. Test updated: asserts collection is empty at
   build time (population is the save/publish flows' responsibility).

3. [Full draft save leaves named-agent subagents without policies — P2] —
   Added a named-agent policy collection seed block in save_manage_config_draft
   (draft_routes.py), mirroring what the publish path already does. When
   raw_policies is present, calls create_agent_policy_system(draft=True,
   policies_data=...) after save_draft.

4. [Published-config fallback reads draft policies — P2] — Added resolved_draft
   variable in build_agents_from_stored_subagents. When load_draft returns None
   and falls back to load_config, resolved_draft is set to False. Passed as
   draft=resolved_draft to create_agent_policy_system.

5. [ProdEmbeddingStore regex — CodeRabbit A+B] — Tightened _SAFE_IDENTIFIER_RE
   to ^[a-z][a-z0-9_]{0,48}$ and switched to fullmatch. Enforces
   lowercase-leading identifier (PostgreSQL rejects digit-leading unquoted
   names) and caps at 49 chars so idx_{name}_embedding (14 extra bytes) stays
   within PostgreSQL's 63-byte limit, preventing silent truncation / CREATE
   INDEX IF NOT EXISTS collisions.

Tests run: uv run ruff format + ruff check — all clean.
uv run pytest tests/unit/test_agent_policy_isolation.py
              tests/unit/test_policy_filesystem_sync_propagation.py
              src/cuga/backend/cuga_graph/policy/tests/test_keyword_operator.py
→ 18 passed in 3.31s.

sami-marreed:
Re-reviewed `d30e9e68fcc8d6613c7a8fef96ad85785e325b5d`. The 18 focused PR tests and all four original regression checks now pass, and CI is green. However, three further regression checks reproduced these remaining issues:

| Priority | Finding |
|---|---|
| **P1** | **An agent named `draft` still collides with the default agent's draft collection.** `get_agent_policy_collection_name("draft", draft=False)` and `get_agent_policy_collection_name("cuga-default", draft=True)` both return `cuga_policies_draft`. `draft` is a valid registry-issued ID. Reproduction: populate the published collection for `draft` with `live-policy`, then save `default-draft` to the default agent's draft collection. Reading the named agent's published policies returns `default-draft` instead of `live-policy`. The `__draft` change fixes the original `crm`/`crm-draft` example, but the default-agent namespace must also be disjoint from every named-agent namespace. [configurable.py:31–37](https://github.com/cuga-project/cuga-agent/blob/d30e9e68fcc8d6613c7a8fef96ad85785e325b5d/src/cuga/backend/cuga_graph/policy/configurable.py#L31-L37) |
| **P2** | **Full draft saves can overwrite a newer policy PATCH.** `save_manage_config_draft` releases `agent_draft_lock` after saving config, then seeds the policy collection outside that lock. Reproduction: pause an older full draft save immediately before its collection seed; complete a newer `patch_draft_policies` request; then resume the older seed. The stored config contains the new policy, but the policy collection contains the old policy. Keep the config write and policy replacement in the same lock scope, as the PATCH path already does. [draft_routes.py:67–96](https://github.com/cuga-project/cuga-agent/blob/d30e9e68fcc8d6613c7a8fef96ad85785e325b5d/src/cuga/backend/server/manage_routes/draft_routes.py#L67-L96) |
| **P2** | **The new 49-character PostgreSQL limit rejects collection names generated for valid agent IDs.** For example, `customer-support-escalation-agent` is accepted by the registry, but its draft policy collection is `cuga_policies_customer_support_escalation_agent__draft`. `ProdPolicyStore.create_schema(384)` now raises `ValueError` for that generated name before opening a database connection. This prevents policy storage initialization for the agent in prod mode. Preserve supported collection names and derive a bounded, collision-resistant index name separately, or consistently bound the generated collection names and handle compatibility. [prod.py:32–38](https://github.com/cuga-project/cuga-agent/blob/d30e9e68fcc8d6613c7a8fef96ad85785e325b5d/src/cuga/backend/storage/embedding/prod.py#L32-L38) |

Validation used isolated local policy storage and a placeholder OpenAI key for client construction; no live LLM calls were made. The PostgreSQL identifier failure was verified during schema-object construction without connecting to PostgreSQL. The concurrency check deterministically paused the older save and confirmed that config retained the newer policy while policy storage reverted to the older one.

Requesting changes until these issues are addressed, particularly the remaining cross-agent overwrite and concurrent-save overwrite.

sami-marreed (changes requested):
Re-reviewed d30e9e68fcc8d6613c7a8fef96ad85785e325b5d. The 18 focused PR tests and four original regression checks pass, but three further checks reproduce remaining issues: the published policies of an agent named draft collide with the default draft collection; a full draft save can overwrite a newer policy PATCH after releasing its lock; and the new PostgreSQL identifier limit rejects generated collection names for valid agent IDs. Details and reproduction steps: https://github.com/cuga-project/cuga-agent/pull/773#issuecomment-5662641180. Requesting changes until these are addressed.

offerakrabi:
Addressed in `9ccf5cd4`:

1. [P1 — default draft collection collision] — Fixed agent collection naming in [`get_agent_policy_collection_name()`](src/cuga/backend/cuga_graph/policy/configurable.py:32) with an explicit named-agent namespace and deterministic digest. The named agent `draft`, punctuation variants, and long IDs can no longer collide. Default runtime storage remains pinned to `cuga_policies` and `cuga_policies_draft`. Committed in `f3d392ee`.
2. [P2 — full draft save can overwrite a newer policy PATCH] — Moved named-agent config persistence and policy replacement under the same per-agent lock in [`save_manage_config_draft()`](src/cuga/backend/server/manage_routes/draft_routes.py:54). Added deterministic concurrency coverage proving a newer PATCH wins. Committed in `1ca3e700`.
3. [P2 — valid generated collection names rejected by PostgreSQL limit] — Restored support for safe table identifiers up to PostgreSQL’s 63-byte limit and added bounded, deterministic, collision-resistant index naming in [`ProdEmbeddingStore`](src/cuga/backend/storage/embedding/prod.py:39). Added DDL and validation tests. Committed in `9ccf5cd4`.

No deferred items from the latest re-review. Earlier findings were verified as already resolved on the reviewed head and required no additional code changes.

Verification:
- `uv run ruff format` — passed; 1 file reformatted.
- `uv run ruff check --fix` — passed; 2 issues fixed, 0 remaining.
- `uv run pytest tests/unit/test_agent_policy_isolation.py tests/unit/test_draft_autosave_isolation.py tests/unit/test_prod_embedding_identifiers.py tests/unit/test_policy_filesystem_sync_propagation.py src/cuga/backend/cuga_graph/policy/tests/test_keyword_operator.py -q` — passed, 32 tests; 5 pre-existing deprecation warnings.
- Final independent review — ready to merge; no Critical, Important, or Minor findings.

The design and plan documents were not committed, per instruction. Pre-existing untracked files remain untouched.

sami-marreed:
Verdict: **Request changes** on `e21d44ee715dd142eae1970db3c482a1e447eff5`.

The three findings from the previous review are addressed. The focused suite passed locally (38 tests), and six earlier regression checks also passed. One new compatibility regression and failing checks still prevent approval.

**[P2] Preserve the configured policy collection when initializing the default runtime.**

`PolicyConfigurable.initialize()` now defaults through `get_agent_policy_collection_name()`, which hardcodes `cuga_policies` and ignores `settings.policy.collection_name`. This changes the existing documented setting's behavior. An installation with policies stored in a custom collection starts reading a different collection after upgrading, so its existing guards are no longer loaded.

Reproduction with isolated storage:

1. Set `settings.policy.collection_name = "custom_policies"`.
2. Persist an intent guard named `existing-guard` in that collection.
3. Initialize a fresh `PolicyConfigurable()` without an explicit collection argument, as the default runtime does.
4. Listing its policies returns `[]` instead of `[existing-guard]`.

Preserve the configured published collection and corresponding draft collection, or provide an explicit migration that retains the existing policies. Named-agent isolation should not silently disable the default agent's configured storage.

[configurable.py:188–190](https://github.com/cuga-project/cuga-agent/blob/e21d44ee715dd142eae1970db3c482a1e447eff5/src/cuga/backend/cuga_graph/policy/configurable.py#L188-L190), [hardcoded mapping:19–43](https://github.com/cuga-project/cuga-agent/blob/e21d44ee715dd142eae1970db3c482a1e447eff5/src/cuga/backend/cuga_graph/policy/configurable.py#L19-L43).

| Validation item | Result / required action |
|---|---|
| `Unit (tests/)` | Failing. I also reproduced all three failures in `tests/unit/test_manage_publish_sync.py` locally. The publish responses select `cuga-default` while the tests expect the requested named-agent config, and the deferred-reindex assertion also fails. Resolve the behavior/test contract and restore passing checks. [CI job](https://github.com/cuga-project/cuga-agent/actions/runs/34857239696/job/104019853781) |
| `Manager and load units` | Failing four manager API tests: `test_08_partial_tool_selection`, `test_patch_llm_draft`, `test_patch_tools_draft`, and `test_patch_policies_draft`. These were observed in CI logs; I did not rerun the live manager suite locally. [CI job](https://github.com/cuga-project/cuga-agent/actions/runs/34857239696/job/104019854124) |
| CodeQL | Reports one new high-severity alert, `SQL query built from user-controlled sources`, at `src/cuga/backend/storage/embedding/prod.py:108–109`. This is a scanner finding, not a confirmed exploitable SQL-injection issue from this review. Triage and resolve or justify it before approval. [CodeQL check](https://github.com/cuga-project/cuga-agent/runs/104020099353) |

The custom-collection regression was verified with isolated local policy storage and a placeholder OpenAI key for client construction, without live LLM calls. Keep **Request changes** until the P2 regression and outstanding checks are resolved.

sami-marreed (changes requested):
Verdict: Request changes on e21d44ee715dd142eae1970db3c482a1e447eff5. The previous three findings are addressed, but a new P2 compatibility regression remains: default runtime initialization ignores policy.collection_name and stops loading existing policies from custom collections. I reproduced this with isolated storage. Unit (tests/) and Manager and load units are failing, and CodeQL reports a new high-severity alert requiring triage; exploitability is not confirmed by this review. Full finding, reproduction, validation, and check links: https://github.com/cuga-project/cuga-agent/pull/773#issuecomment-5677351678.

offerakrabi:
Addressed in `7e83dd76`:

**1. Production fix — `configurable.py:190`**  
Added one line to read `configured_collection_name = getattr(policy_config, "collection_name", None)` then chained it into the resolution:
```python
final_collection_name = collection_name or configured_collection_name or get_agent_policy_collection_name()
```
`get_agent_policy_collection_name()` itself is untouched and still settings-independent.

**2. Test replacement — `test_agent_policy_isolation.py`**  
Replaced `test_default_policy_runtime_ignores_configured_collection_name` (which codified the regression) with four new tests:
- `test_default_policy_runtime_uses_configured_collection_name` — parametrized over `custom_policies` (honoured), `""` (fallback), `None` (fallback)
- `test_default_policy_runtime_explicit_arg_overrides_configured_collection`
- `test_default_policy_runtime_draft_unaffected_by_configured_collection`

**Validation:**
- `pytest tests/unit/test_agent_policy_isolation.py` — **19 passed**
- `ruff check` + `ruff format --check` — **clean**
- `pytest tests/unit tests/integration -m "not manual and not pgvector and not load and not e2e and not stability"` — **1891 passed**; 3 failures in `test_sdk_citations.py` are pre-existing (pass in isolation, fail only under full-suite ordering — reproduced on the branch tip before this change)

PR #776, opened 2026-09-14
fix(policy): append pydantic format instructions for LiteLLM models

mbwhite:
## Bug fix

Fixes #

### Summary

LiteLLM uses `PydanticOutputParser` for structured output, which relies on explicit JSON format instructions being present in the prompt. When those instructions were absent the model returned markdown prose and the parser raised an exception during policy conflict resolution.

**Root cause:** The system prompt was interpolated with an f-string, causing LangChain's `ChatPromptTemplate` to interpret `{` / `}` in the Pydantic JSON schema as template variables and raise `KeyError`.

**Solution:** Replace the f-string with a separate `.partial()` system message (matching the `json_mode` fallback pattern in `base_agent.py`), keeping JSON braces out of the template engine entirely. Appends `PydanticOutputParser.get_format_instructions()` so all code-paths (including LiteLLM) receive the required JSON schema hint.

### Testing

- Regression tests for LiteLLM format-instructions wiring: verifies rendered messages include the JSON schema; guards index-off-by-one; asserts LLM selecting index 2 returns the correct policy.

---
Helped-by: IBM Bob


sami-marreed:
Thanks for the fix. This addresses a real issue in the LiteLLM policy conflict-resolution path, and passing the format instructions through `.partial(...)` correctly avoids treating JSON schema braces as template variables.

Please add a regression test marked `@pytest.mark.unit` that exercises the prompt construction and parser path with a mocked LiteLLM response. It should verify that the JSON schema instructions reach the model, the prompt renders without missing-variable errors, and a response selecting the second candidate returns that policy instead of falling back to the first candidate. Please keep the real prompt/parser wiring in the test rather than mocking `BaseAgent.get_chain` entirely.

Once the regression test is added and CI passes, we can approve this PR.

mbwhite:
Apologies to the maintainers — I accidentally pushed an unrelated performance fix (C-1: skip redundant output-formatter LLM call) onto this branch. It has no business being in this PR, which should stay focused on the LiteLLM format-instructions bug. I've reverted that commit and force-pushed, so the branch is back to just the three intended commits. Sorry for the noise.

Issue #600, opened 2026-08-06
[Bug]: variables summary is persisted into every user message, duplicating ~12K chars per turn until the context window overflows


Sergey-Zeltyn:
## What happened

`call_model` appends the "Available Variables" addendum to the **last** user message — correct intent — but then writes the mutated message back into the conversation state. On the next turn that message is no longer last, a new message receives its own addendum, and the previous one **keeps the copy it was given**.

After N turns the history holds N copies of a block that is capped at 12,000 characters each but unbounded in count. Growth is linear and unrelated to how much the conversation actually contains.

### Measured on `2ff4dfb_1` (AppWorld med, gpt-oss-120b)

At the final generation the context held **16 user messages carrying a variables summary, totalling 314,992 characters** (~79K tokens — 60% of the model's 131,072-token window). MD5 of the trailing 6,000 chars of each block: **15 of the 16 are byte-identical**.

Anatomy of one such message (~17,500 chars):

```
Execution output: total_manager_threads=20 meetings_found=0 sample=[]   <- ~200 chars, the real payload
--- Summary: **Overall Status / Analysis** ...                          <- reflection text
... [Variables summary trimmed to 12000 chars]                          <- ~12,000 chars, IDENTICAL every turn
You can use these variables directly by their names.
```

Per-generation input tokens for the main thread grow almost perfectly linearly at **~+6,300/turn**:

| gen | 0 | 10 | 20 | 30 | 38 | 40 |
|---|---|---|---|---|---|---|
| input tokens | 6,923 | 30,910 | 66,034 | 99,632 | 125,608 | **400** |

The model's own output is only ~1,300 tokens/turn, so expected accumulation is ~1,100–1,300/turn. The observed 6,300/turn is roughly **5× inflated**, and the excess is the duplicated block.

**This task made zero tool calls** while consuming 1,559,914 tokens — so the growth is not tool-output flooding. It is copies of one static block.

### Consequence: the provider terminates the task

Three task-runs across two bundles died outright:

| bundle | task | total tokens | input at death | limit | match_rate when killed |
|---|---|---|---|---|---|
| med-run3 | `7574325_1` | 1,865,874 | 132,444 | 131,072 | 0.8 |
| med-run4 | `6474048_1` | 2,477,477 | 134,174 | 131,072 | 0.25 |
| med-run4 | `2ff4dfb_1` | 1,559,914 | 131,885 | 131,072 | 0.222 |

```
[...]

PR #777, opened 2026-09-15
feat(cuga-lite): native function-calling execution mode + step discipline

segevshlomovIBM:
## Feature

Related to #471 · Supersedes #472

### Summary

Two opt-in settings for CUGA Lite. Both default off; the default path is unchanged (golden test).

**`cuga_lite_execution_mode = "function_calling"`** (default `codeact`)
The model calls tools natively instead of writing Python. A new `tool_exec` node runs the `tool_calls` and replies with `ToolMessage`s — same budgets, tracker and timeout as the sandbox — then the model reads the results and either calls again or answers. Short dedicated prompt; the tools the sandbox could call are advertised via `bind_tools`.

**`cuga_lite_step_discipline = "one_tool_per_step"`** (default `off`, works in both modes)
One tool call per step is attempted (success or error); the model reads that result before the next call. CodeAct: the block's tool-call cap becomes 1 and the block's variables are kept. Function-calling: extra calls get a "deferred" reply.

Set either in `settings.toml`, a model runtime profile, `config["configurable"]`, or the SDK (`CugaAgent(execution_mode=…, step_discipline=…)`, also per `invoke()` / `stream()`). README section: *Execution mode: CodeAct or native function calling*.

**Unchanged:** codeact runs, the Supervisor graph, existing prompts, `code_executor.py`.
**Not included:** a tool-approval interrupt on the function-calling path — the mode **refuses to start** (clear error, before any tool runs) while an enabled tool-approval policy exists — or when policy storage cannot be checked (fail closed); FC runs tools in-process and sequentially; the pre-execute VERIFY gate and reflection are CodeAct-only; step discipline covers local-executor blocks only.
**Note for review:** bind mode `none` → `tools` (the executable set) happens only in function-calling mode, and only when it resolved to `none`. Past the 128-tool cap the existing bind-cap shortlister round-trip applies (`[shortlister.bind_cap] strategy = "embedding"` avoids it).

### Commits

1. native function-calling execution mode
2. step discipline, both modes
3.–5. review fixes (each reproduced first): history replayed across SDK turns as a provider-valid transcript (state crosses `model_dump()` as bare `BaseMessage` shells, so `tool_calls` / `tool_call_id` were lost); every issued call id answered before a step-limit or budget end; fail-closed on tool-approval policies; empty-reply retry; executable-set binding; CodeAct few-shots withheld in FC; variables addendum; tool_exec outer guard, block-timeout bound and batch output limit; FC steps streamed to the UI.

### Testing
- [x] Tested locally; tests pass — every `tests.yml` shard re-run locally shard by shard (Lint, Unit (graph) ×6, Unit (tests/), Events, Policy unit + e2e, SDK integration, Manager/load, Mocked load): all green except 3 `test_e2b_lite.py` failures that need the `e2b-code-interpreter` package and reproduce on `main`. 10 FC test files incl. an SDK two-turn test and a feature-off golden.
- [x] `ruff format` + `ruff check` clean.

## Summary by CodeRabbit

* **New Features**
  * Added optional native function-calling execution for Cuga Lite, with tool-call results, retries, timeouts, budgets, and multi-turn history support.
  * Added configurable one-tool-per-step discipline and per-call execution-mode overrides.
  * Added function-calling prompts with app, skill, agent, and evidence-first guidance.
  * Added streaming events for issued tool calls and their results.

* **Safety & Reliability**
  * Function-calling now fails closed when tool-approval policies cannot be verified.
  * Improved handling of malformed, missing, timed-out, and failed tool calls.

* **Documentation**
  * Documented execution modes, configuration precedence, and step-discipline settings.


src/cuga/backend/cuga_graph/nodes/cuga_agent_core/graph/shared_graph.py
```diff
@@ -58,4 +67,10 @@ def build_agent_graph(
     # Execute node returns a state update (not Command); loop back for the NL answer.
     graph.add_edge(adapter.execute_node_name, "call_model")
 
+    if tool_exec_node is not None:
+        graph.add_node("tool_exec", tool_exec_node)
+        # Same shape as the sandbox edge: the node returns a state update, and the
+        # loop closes back into call_model for the next model turn.
+        graph.add_edge("tool_exec", "call_model")
```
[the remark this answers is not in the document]
segevshlomovIBM: Fixed in d6f0b8d5: the static edge is gone; tool_exec returns Command(goto="call_model") on normal exits and Command(goto=END) on error exits. Test: test_normal_exit_routes_to_call_model_and_error_exit_to_end.

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/graph_adapter.py
```diff
@@ -239,6 +319,170 @@ def _any_execution_ran(self, state: Any) -> bool:
...
+        content = strip_harmony_tokens(normalize_assistant_text(getattr(response, "content", "")) or "")
+        reasoning = normalize_assistant_text(
+            (getattr(response, "additional_kwargs", None) or {}).get("reasoning_content")
+        )
+        if not isinstance(response, AIMessage):
+            response = AIMessage(content=content, tool_calls=tool_calls)
+
+        if budget_exhausted:
+            # No tools were bound for the grace turn, so any tool_calls are noise.
+            tool_calls, invalid_tool_calls = [], []
```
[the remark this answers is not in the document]
segevshlomovIBM: Addressed in d6f0b8d5, differently: the assistant turn is persisted verbatim and every issued id gets a synthetic error ToolMessage ("Not executed: the tool budget for this turn is spent…"), so the transcript stays provider-valid without rewriting the model's message. Same treatment on a step-limit breach. Tests: test_budget_exhausted_turn_ignores_tool_calls_and_sends_the_instruction_outbound_only, test_step_limit_breach_answers_pending_calls_before_the_error.

src/cuga/settings.toml
```diff
@@ -53,6 +52,17 @@ cuga_lite_bind_tools_include_find_tools = false  # Also bind find_tools alongsid
 cuga_lite_bind_tools_max_count = 128
 # Pad shortlister output to fill the cap. Off by default for code-execution mode.
 cuga_lite_bind_tools_pad_to_cap = false
+# Execution mode: codeact (write Python, run it in the sandbox — default) | function_calling (the model
+# issues native tool_calls, a tool_exec node runs them and replies with ToolMessages). Aliases: fc, native.
+# Per-invoke: config["configurable"]["cuga_lite_execution_mode"]. Env: DYNACONF_ADVANCED_FEATURES__CUGA_LITE_EXECUTION_MODE
+cuga_lite_execution_mode = "codeact"
+# Step discipline, both modes: off (default) | one_tool_per_step — observe one result before the next call.
+# codeact: only the first tool call in a generated block runs (per-block cap becomes 1). function_calling: only
+# the first tool_call of a turn runs; the rest are answered with a "deferred" ToolMessage. Env: DYNACONF_ADVANCED_FEATURES__CUGA_LITE_STEP_DISCIPLINE
```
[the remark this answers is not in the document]
segevshlomovIBM: Fixed in d6f0b8d5: deferral is by position (index > 0), so an error on the first call is the result the model reads next. Test: test_step_discipline_defers_by_position_even_when_the_first_call_fails.

src/cuga/backend/cuga_graph/nodes/cuga_agent_core/graph/shared_nodes.py
```diff
@@ -226,6 +226,27 @@ async def call_model(state: Any, config: RunnableConfig = None) -> Command:
...
+        # path below is skipped entirely — never blended.
+        fc_command = await adapter.execute_call_model_fc(
+            state=state,
+            config=config,
+            configurable=configurable,
+            active_model=active_model,
+            bound=bound,
+            invoke_config=invoke_config,
+            system_content=system_content,
+            modified_messages=modified_messages,
```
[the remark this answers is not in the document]
segevshlomovIBM: Fixed in d6f0b8d5: call_model passes variables_addendum to the seam and the FC turn appends it to the last user message, outbound only (#600 contract). Test: test_variables_addendum_rides_the_last_human_turn_outbound_only.

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/prepare_node.py
```diff
@@ -730,7 +738,31 @@ async def _wrapped(*args, **kwargs):
...
+                == STEP_DISCIPLINE_ONE_TOOL_PER_STEP
+            )
+            dynamic_prompt = render_fc_prompt(
+                instructions=effective_instructions,
+                special_instructions=special_instructions_final,
+                is_autonomous_subtask=settings.advanced_features.force_autonomous_mode
+                or is_autonomous_subtask,
+                step_discipline=_one_per_step,
+                fragments=_fragments,
+            )
```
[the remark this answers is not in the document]
segevshlomovIBM: Fixed in d6f0b8d5: render_fc_prompt takes skills_prompt_section and agents_prompt_section (rendered only when those tools are bound). Test: test_skill_and_agent_catalogues_are_rendered_when_given.

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/prepare_node.py
```diff
@@ -730,7 +760,33 @@ async def _wrapped(*args, **kwargs):
...
+            dynamic_prompt = render_fc_prompt(
+                instructions=effective_instructions,
+                special_instructions=special_instructions_final,
+                is_autonomous_subtask=settings.advanced_features.force_autonomous_mode
+                or is_autonomous_subtask,
+                step_discipline=_one_per_step,
+                fragments=_fragments,
+                skills_prompt_section=skills_prompt_section if skills_enabled else "",
+                agents_prompt_section=agents_prompt_section if agents_enabled else "",
+                apps=apps_for_prompt if enable_find_tools else None,
```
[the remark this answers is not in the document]
segevshlomovIBM: Fixed in 7521a97a: the prompt advertises the catalogue find_tools was built with (captured before the runtime app definitions are appended), so every advertised app_name has a mapping.

tests/unit/test_fc_sdk_fail_closed_approval.py
```diff
@@ -0,0 +1,93 @@
...
+    codeact_model = _ScriptedModel(
+        [AIMessage(content="```python\nprint(await delete_record(record_id=1))\n```")]
+    )
+    agent._model = codeact_model
+    agent._graph = None
+    agent._compiled_graph = None
+    result = await agent.invoke("delete record 1", thread_id="codeact-approval", execution_mode="codeact")
+
+    assert codeact_model.invocations == 1
+    assert "approval" in result.answer.lower(), result.answer
```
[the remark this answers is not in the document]
segevshlomovIBM: Fixed in 7521a97a: the guarded tool records its runs and the test asserts RAN == [] on both paths, and the CodeAct assertion checks the policy's own approval_message rather than the substring.

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/graph_adapter.py
```diff
@@ -239,6 +435,247 @@ def _any_execution_ran(self, state: Any) -> bool:
...
+
+        try:
+            policy_system = PolicyConfigurable.from_config(config or {})
+            policies = await policy_system.agent.storage.list_policies(
+                policy_type=PolicyType.TOOL_APPROVAL, enabled_only=True, limit=1
+            )
+            return bool(policies)
+        except Exception as exc:
+            logger.warning("[fc] could not query tool-approval policies ({}); proceeding", exc)
+            return False
```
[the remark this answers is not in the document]
segevshlomovIBM: Fixed in 7521a97a: the guard now fails closed on any verification failure (distinct FC_TOOL_APPROVAL_UNVERIFIED message, no model call). PolicyStorage.list_policies gained strict=True so a backend failure re-raises instead of returning [] — both the initialisation and the backend failure paths refuse. Tests: test_refuses_when_the_policy_query_fails, test_guard_asks_storage_strictly_so_a_backend_failure_is_not_an_empty_list, policy/tests/test_storage_strict_list.py.

src/cuga/backend/cuga_graph/nodes/cuga_agent_core/graph/shared_nodes.py
```diff
@@ -191,6 +191,22 @@ async def call_model(state: Any, config: RunnableConfig = None) -> Command:
...
+            elif msg_role == "tool":
+                # A native tool result left by a function-calling turn on this thread.
+                # Matched on ``type``: persisted history crosses the SDK boundary as bare
+                # BaseMessage shells. Rendered like an execution output so a CodeAct
+                # turn on the same thread still sees what the tool returned.
+                modified_messages.append(msg)
+                tool_name = (
+                    getattr(msg, "name", None)
+                    or (msg.get("name") if isinstance(msg, dict) else None)
+                    or "tool"
```
haroldship: @coderabbitai what will happen if there is no "name" and so the tool_name gets set to "tool"?

src/cuga/backend/cuga_graph/nodes/cuga_lite/model_runtime_profile.py
```diff
@@ -114,3 +114,160 @@ def resolve_bind_tools_fields(
...
+
+EXECUTION_MODE_CODEACT = "codeact"
+EXECUTION_MODE_FUNCTION_CALLING = "function_calling"
+
+_EXECUTION_MODE_ALIASES: Dict[str, str] = {
+    "codeact": EXECUTION_MODE_CODEACT,
+    "code_act": EXECUTION_MODE_CODEACT,
+    "function_calling": EXECUTION_MODE_FUNCTION_CALLING,
+    "function-calling": EXECUTION_MODE_FUNCTION_CALLING,
+    "fc": EXECUTION_MODE_FUNCTION_CALLING,
```
haroldship: _EXECUTION_MODE_ALIASES only has fc. The `settings.toml` comment at line 57 also lists "native". Setting `cuga_lite_execution_mode = "native"` stays on CodeAct. Can we add native here, or drop it from that comment so it matches the README?

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/graph_adapter.py
```diff
@@ -239,6 +440,252 @@ def _any_execution_ran(self, state: Any) -> bool:
...
+                    self.messages_key: final_messages + [HumanMessage(content=EMPTY_RESPONSE_CORRECTION)],
+                    "script": None,
+                    "final_answer": "",
+                    "execution_complete": False,
+                    "step_count": new_step_count,
+                    self.metadata_key: retry_meta,
+                },
+            )
+
+        if "```" in content and not budget_exhausted:
```
haroldship: `if "```"` in content treats _any_ fence as a mode violation, not just a Python block. Unlike the empty-reply path, there is no one-retry cap: only cuga_lite_max_steps. Can we accept a fenced snippet after one correction, or only flag executable-looking Python?

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/graph_adapter.py
```diff
@@ -153,6 +329,31 @@ async def resolve_bind_tools(
         configurable: dict,
         config: Any = None,
     ) -> Any:
+        # Function-calling mode is inert without advertised tools, and the shipped
+        # default is bind mode "none". Upgrade only when the *resolved* mode is none
+        # — an explicit non-none choice from configurable, a model profile or
+        # settings is respected. Additive: no effect in codeact.
+        if self._execution_mode(configurable) == EXECUTION_MODE_FUNCTION_CALLING:
```
haroldship: in function-calling mode, a failed / unsupported bind_tools still continues with the unbound model. This except returns None, then the caller does `or active_model`. Same degrade happens in `_safe_bind` at helpers/bind_tools.py:54. That run cannot do native tool calling. Can we stop with a clear error, same as the tool-approval guard?

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/graph_adapter.py
```diff
@@ -153,6 +329,31 @@ async def resolve_bind_tools(
...
+                    configurable = {
+                        **(configurable or {}),
+                        "cuga_lite_bind_tools_mode": "tools",
+                        "cuga_lite_bind_tools_tool_names": names,
+                    }
+                    logger.info(
+                        "[fc] bind_tools mode was 'none'; advertising the {} executable tool(s)", len(names)
+                    )
+                else:
+                    configurable = {**(configurable or {}), "cuga_lite_bind_tools_mode": "all"}
```
haroldship: When the executable-name list is empty we upgrade bind mode to all. That advertises tools tool_exec may not be able to run. Can we bind nothing or fail the turn instead of opening the full catalogue?

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/tool_exec_node.py
```diff
@@ -0,0 +1,363 @@
...
+                },
+            )
+        except Exception as exc:
+            # Mirrors the sandbox's outer guard: a failure outside the per-call
+            # handling must still report the budget the batch spent, or the
+            # checkpoint keeps the pre-batch counts (see _budget_updates).
+            logger.error(f"tool_exec failed: {exc}")
+            return core_create_error_command(
+                adapter,
+                adapter.get_messages(state),
```
haroldship: On the outer-guard path we pass `adapter.get_messages(state)` and drop `results`. If the batch already ran, we `persist tool_calls` with no `ToolMessages`. Instead, keep `results` and invent replies for any missing ids, like the step-limit path?

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/graph_adapter.py
```diff
@@ -40,6 +57,165 @@
...
+    """
+    out: List[ToolMessage] = []
+    for index, call in enumerate(list(calls or []) + list(invalid or [])):
+        call_id = (str(call.get("id") or "") if isinstance(call, dict) else "") or f"call_{index}"
+        name = (call.get("name") if isinstance(call, dict) else None) or "unknown"
+        out.append(ToolMessage(content=reason, tool_call_id=call_id, name=str(name), status="error"))
+    return out
+
+
+def _normalize_history_for_replay(messages: List[Any]) -> List[BaseMessage]:
```
haroldship: This avoids a provider 400 by rewriting lost tool turns as user text. Turn 2 is then not a native function-calling transcript. The 2 turn test can pass even when `tool_calls` never come back. Rebuild real `AIMessage` + `ToolMessage` pairs, and assert turn 2 still has them.

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/tool_exec_node.py
```diff
@@ -0,0 +1,363 @@
...
+        track_tool_calls = configurable.get("track_tool_calls", False)
+        max_steps = configurable.get("cuga_lite_max_steps") if "cuga_lite_max_steps" in configurable else None
+        model_name = resolved_runtime_model_name(
+            configurable_llm=configurable.get("llm"), graph_default_model=getattr(adapter, "_model", None)
+        )
+        one_per_step = resolve_step_discipline(configurable, model_name) == STEP_DISCIPLINE_ONE_TOOL_PER_STEP
+        # One batch is the function-calling analogue of one code block, so the
+        # outer bound per call is the block wall clock. Registry tools keep their
+        # own inner ``tool_call_timeout``.
+        timeout = getattr(settings.advanced_features, "sandbox_execution_timeout", 30) or None
```
haroldship: Each FC call gets `sandbox_execution_timeout`, but the batch does not. N calls can take N × 30s. Either add one wait around the batch, or say in the README that the timeout is per call.

tests/unit/test_fc_sdk_two_turn.py
```diff
@@ -0,0 +1,105 @@
...
+    print("\nTURN 2 outbound:", shapes)
+
+    assert isinstance(outbound[0], SystemMessage)
+    assert all(type(m) is not BaseMessage for m in outbound), (
+        f"bare BaseMessage replayed to the provider: {shapes}"
+    )
+    ai_with_calls = [m for m in outbound if isinstance(m, AIMessage) and m.tool_calls]
+    tool_replies = [m for m in outbound if isinstance(m, ToolMessage)]
+    dangling = {c["id"] for m in ai_with_calls for c in m.tool_calls} - {t.tool_call_id for t in tool_replies}
+    assert not dangling, f"tool_calls without a ToolMessage reply: {dangling} in {shapes}"
```
haroldship: The dangling-id check passes if turn 1’s `tool_calls` were stripped. Can we also assert outbound turn 2 still has `tool_calls` and a matching `ToolMessage` for `call_1`?

src/cuga/backend/cuga_graph/nodes/cuga_lite/adapter/prepare_node.py
```diff
@@ -763,6 +823,11 @@ async def _wrapped(*args, **kwargs):
                 len(dynamic_prompt),
             )
         else:
+            if _execution_mode == EXECUTION_MODE_FUNCTION_CALLING:
+                logger.warning(
```
haroldship: When `_static_prompt` is set, FC uses it verbatim and only logs a warning. A CodeAct-shaped prompt will tell the model to write fences, which we then treat those as violations. Either refuse that combo altogether, or ignore the static prompt in FC mode.

haroldship (changes requested):
Thanks for the careful opt-in design. Default CodeAct stays off `tool_exec`, the approval guard fails closed, and every tool-call id gets a reply on the happy path. I would not merge function-calling as-is.

#### Please fix before merge

1. "native" is documented in settings.toml but not in `_EXECUTION_MODE_ALIASES`, so that spelling stays on CodeAct.
2. Any ``` in a final answer is treated as a mode violation until `cuga_lite_max_steps`, not one retry.
3. Failed / unsupported `bind_tools` still continues unbound. Fail closed, like the approval guard.
4. An empty executable-name list upgrades bind mode to all and can advertise tools `tool_exec` cannot run.
5. The `tool_exec` outer-guard path drops results after tools may already have run, leaving unanswered `tool_calls`.
6. SDK turn-2 replay is provider-valid (no 400), not a native `AIMessage` / `ToolMessage` transcript. The two-turn test can pass on the flatten path.


#### Non-blocking

7. Timeout is per call, not per batch — tighten the code or the README.
8. A static prompt plus FC only warns; a CodeAct-shaped prompt will fight the mode.
9. Assert that turn 2 still has tool_calls and a matching ToolMessage.
4. Also rebase — this branch is behind main.

Issue #616, opened 2026-08-10
[Feature]: Autonomous tasks with tools


sami-marreed:
Part of #611.

## What you want and why

Evaluate CUGA on autonomous multi-tool tasks: full/partial schemas, todos/reflection, long context/summarization, cost/small models, AppWorld agent-quality issues.

## How it could work

Capability matrix as nested Features or linked agent issues (#272 partial schemas, #574 summarization, AppWorld leaves).

## Links or extra context

AppWorld is primary public bench inside Continuous epic

Issue #611, opened 2026-08-10
[Epic] Continuous evaluation & regression


sami-marreed:
## What you want and why

Replace the scattered evaluation portfolio (mega-epic #239, #347, harness work, AppWorld, policy measurement) with one outcome epic for **continuous evaluation and regression** of CUGA: harness, dashboard, PR/regression automation, capability matrix (autonomous tool use, policies, KPIs), and AppWorld as the primary public bench.

## How it could work

Epic → Feature track parents → leaf issues (sub-of-sub):

1. [`[Design]: Improvement targets for continuous evaluation`](https://github.com/cuga-project/cuga-agent/issues/613)
2. [`[Feature]: cuga-eval harness`](https://github.com/cuga-project/cuga-agent/issues/614) — scripts, bundles, model profiles, Langfuse/export, concurrency
3. [`[Feature]: Continuous reporting, dashboard, and regression`](https://github.com/cuga-project/cuga-agent/issues/615) — absorbs #347 (#478–#483); baselines/comparative via #482
4. [`[Feature]: [Autonomous tasks with tools]`](https://github.com/cuga-project/cuga-agent/issues/616) — schemas, todos/reflection, summarization, cost/small models, AppWorld agent-quality leaves
5. [`[Feature]: Selecting tools from a large number of tools`](https://github.com/cuga-project/cuga-agent/issues/617)
6. [`[Feature]: CUGA tasks under policies`](https://github.com/cuga-project/cuga-agent/issues/618) — Oak / policy-on measurement (product policies remain #75)
7. `[Feature]: Evaluation KPIs` — tokens, latency, consistency, pass-rate
8. `[Feature]: Long-term evaluation extensions` — BACKLOG (Knowledge, Supervisor, runtime spawning, Memory); review with @Sergey-Zeltyn before staffing

## Links or extra context

- Supersedes planning under #239 and #347
- Sibling epics: Vakra evaluation, Tau-bench (tau2 + tau3)
- Policy product delivery: #75

PR #780, opened 2026-09-15
fix(cuga-lite): stop advertising filesystem helpers the executor never injects

### LINKED ISSUE #778: [Bug]: the executor prompt advertises filesystem helpers that are absent in a default run
Part of #616 (epic #611).

### What happened

`advanced_features.enable_filesystem_tools` ships **`false`** in `settings.toml`, so
`read_file`, `write_file` and `list_files` are **not** in the execution context of a
default run.

The CugaLite executor prompt describes them unconditionally, and three of the bundled
`find_tools` few-shot turns demonstrate them — one says outright *"The filesystem tools
are already available - no discovery needed"*. A demonstration gets imitated, so the
model calls them and gets `NameError`.

This is a **shipped-default** defect: every deployment that has not explicitly enabled
filesystem tools is being told about tools that do not exist.

Measured on AppWorld `test_challenge` hard (195 tasks, `aws/gpt-oss-120b`):

| | |
|---|---:|
| `NameError` on these three names | **114** |
| tasks affected | **48** (a quarter of the set) |
| of those, tasks that failed | **30** |

Smaller models are hit hardest: they follow the few-shot examples literally rather than
probing for the tool first.

### How to reproduce

1. Leave `advanced_features.enable_filesystem_tools` at its shipped default (`false`).
2. Give CugaLite any task whose natural solution writes a file — "save the report to
   `report.md`", "download these attachments and list them".
3. The model emits `await write_file("./report.md", content)`.
4. Execution fails with `NameError: name 'write_file' is not defined`.

The prompt text is in
`src/cuga/backend/cuga_graph/nodes/cuga_lite/prompts/mcp_prompt.jinja2`; the few-shot
turns are the bundled `find_tools` examples.

### Environment

`origin/main` @ `8315704c1`. Reproduced with `aws/gpt-oss-120b` via LiteLLM; the prompt
defect is model-independent.

### Notes

Related but distinct from #583 (closed), which was about injected filesystem tools not
going through `make_tool_awaitable` — that is the *enabled* path. This is the *disabled*
path: the tools are correctly absent, and the prompt claims otherwise.

`run_command` is a separate capability under `enable_shell_tool` and the two flags are
currently conflated in the prompt.

sami-marreed:
Closes #778. Part of #616 (epic #611).

## The problem

`advanced_features.enable_filesystem_tools` ships **`false`**, so `read_file`,
`write_file` and `list_files` are absent from a default run's execution context. The
executor prompt described them unconditionally, and three bundled `find_tools` few-shot
turns *demonstrated* them — one saying outright *"The filesystem tools are already
available - no discovery needed"*.

A demonstration gets imitated. The model called them and got `NameError`.

Measured on AppWorld `test_challenge` hard (195 tasks, `aws/gpt-oss-120b`):

| | |
|---|---:|
| `NameError` on these three names | **114** |
| tasks affected | **48** of 195 |
| of those, tasks that failed | **30** |

After this change, on the same 195 tasks: **0**.

## What changed

Three small pieces, all keyed on the flag that decides whether the tools exist:

1. **`create_mcp_prompt` takes `enable_filesystem_tools`** and gates the two
   helper-specific bullets on it. `run_command` stays under `enable_shell_tool` — they
   are separate capabilities and the prompt had conflated them.
2. **`drop_examples_using_absent_helpers`** removes few-shot turns that demonstrate the
   helpers when they are disabled. Gating the prose but leaving the examples would still
   teach a call that raises.
3. **`prepare_node` derives the flag from the resolved runtime bundle**, not from
   `settings`. The prompt then describes what was *actually* injected, so the two cannot
   drift.

## Why this shape

The alternative — always injecting the helpers — changes what a default deployment can
do to the host filesystem, which is a security decision, not a prompt bug fix. Describing
only what is present is the conservative half, and it is the half that is unambiguously
wrong today.

## Tests

`tests/test_filesystem_helper_gating.py`, 6 cases: the bullets and the few-shot turns are
absent when the flag is off and present when it is on, `run_command` tracks
`enable_shell_tool` independently, and the prompt follows the runtime bundle rather than
settings.

## Not AppWorld-specific

No benchmark-specific wording. The flag is the shipped default, so this affects every
deployment that has not explicitly turned filesystem tools on.

## Summary by CodeRabbit

* **Improvements**
  * Shell and filesystem tools now follow the active runtime configuration for each request.
  * Prompts and execution guidance accurately reflect available shell and filesystem capabilities, including remote operations.
  * Examples requiring unavailable helpers are automatically filtered, while compatible examples remain available.
  * Filesystem-disabled workflows no longer display irrelevant filesystem instructions or tool descriptions.


Sergey-Zeltyn:
Thanks for this — the diagnosis is right, the 114→0 NameError measurement is convincing,
and deriving the prompt flag from the resolved runtime bundle "so the two cannot drift"
is exactly the right principle. I ran the new filter and the template against the shipped
assets while reviewing; one question and a few gaps below, roughly by impact.

## Main question

**1. Was the surviving few-shot transcript reviewed? (`prompt_utils.py:808`)**
Per-turn dropping is clearly the intended mechanism, and it does zero out the NameErrors.
But the bundled set is one paired 8-turn conversation, and running the filter on it with
fs off leaves `[user, user, user, user, assistant]`: three orphaned "Execution output:"
turns for code the model never sees, the fixture's only `find_tools` call demonstration
deleted (which was the point of this few-shot set), and the kept final answer still
claiming "The drafted report was saved to ./recruitment_report.md" — so the default
prefix now teaches claiming file writes without demonstrating any work, and sends
consecutive same-role messages (which strict role-alternation providers reject). The
114→0 measurement can't see any of this.
If the residue is a known trade-off, fine — but I'd suggest either dropping the whole
conversation when any turn uses an absent helper (`return []`, also less code), or
shipping an fs-free variant of the fixture so the fs-off default keeps a coherent
`find_tools` demonstration.

## Should fix

The "describe only what is present" invariant still leaks through three channels this PR
doesn't touch:

**2. `execution_policy.py:53-58`** (cuga_agent_core/policy): the split-execution note
names `read_file`/`write_file`/`list_files` unconditionally, and `split_execution_active`
fires on remote shell alone — so `enable_shell_tool=true` + `sandbox_mode="native"` (the
shipped default) + fs off renders those names into the prompt while the helpers raise
`NameError`. The new test misses it because it renders with empty special_instructions.

**3. `local_sandbox_executor.py:276` / `native_sandbox_executor.py:310`**: `run_command`'s
tool description still says "Write scripts with `write_file` before running them" and
"`read_file` accepts both", and it reaches the prompt via Current Available Tools on
every shell-on/fs-off run.

**4. `prepare_node.py:755`**: `enable_shell_tool` is passed from **raw settings** while
injection consults `_runtime_backends.shell` — the exact asymmetry line 756 fixes for
filesystem, and the PR's own no-drift principle applied one line down but not here. With
`sandbox_mode="e2b"` (or opensandbox with `opensandbox_sandbox=false`), `run_command` is
advertised but never injected. Should be `_runtime_backends.shell != "none"`.

## Also worth fixing

**5. Filter covers 3 of 8 helpers, via a third copy of the name list
(`prompt_utils.py:807`)**: `create_filesystem_tools` injects 8 tools; the tuple lists 3,
so a user-supplied `mcp_few_shot_examples` turn demonstrating `edit_file`/`search_files`
etc. with fs off still teaches a `NameError`. `server/main.py:3561` already carries a
full private copy, and the new test's `HELPERS` is a fourth. Exporting a single
`FILESYSTEM_TOOL_NAMES` constant from `executors/filesystem` and using it in all four
places would keep them from drifting.

**6. The new jinja gate is nested inside `{% if enable_shell_tool %}`
(`mcp_prompt.jinja2:134`)**: fs-on/shell-off is a supported combination
(`resolve_runtime_backends` computes the flags independently) — verified by rendering:
all 8 helpers are injected but write-before-run, the indentation rule, and
workspace-path guidance never render. The filesystem bullet needs its own top-level
conditional.

**7. Static-prompt path bypasses the filter (`prepare_node.py:737` vs `:791`)**: already
flagged by CodeRabbit; agreed, and the proposed hoist (filter once where
`few_shot_examples` is finalized) is right. One nuance: this is currently latent —
`sdk.py` and `entry_graph.py` both pass `prompt=None` — so it bites only direct
`create_cuga_lite_graph(prompt=...)` callers. The skipped branch even logs "few-shot
injection skipped" and then ships the list anyway.

## Minor

- **Substring matching false-positives** (`prompt_utils.py:808`): a prose turn saying
  "do not use read_file" and a call to `s3_read_files_batch` are both silently dropped
  (verified). Matching `await read_file(` or the actual injected-name set would fix both
  modes.
- **Tests don't cover the wiring**: `test_bundled_find_tools_examples_would_be_filtered`
  never runs the filter — running it on the loaded fixture would surface point 1
  immediately, and nothing tests the `prepare_node` call site (where point 7 lives).
- `_runtime_backends.filesystem != "none"` is now spelled inline in three places; a
  `filesystem_enabled` local (or property on `RuntimeBackends`) would keep the gates
  from drifting — and is the natural home for the fix in point 4.

Sergey-Zeltyn (changes requested):
see comment above

sami-marreed:
Addressed the review points in the latest push:

1. **Few-shot conversation integrity:** when a transcript calls an unavailable runtime helper, the entire transcript is dropped. This prevents orphaned execution outputs, consecutive same-role messages introduced by filtering, and unsupported final file-write claims. With filesystem tools disabled, the bundled filesystem-dependent conversation is omitted.
2. **Split-execution guidance:** the note describes only the remote capabilities that are actually available, using the same resolved shell/filesystem backends as tool injection.
3. **Shell tool descriptions:** removed filesystem-helper assumptions from the local, native, and OpenSandbox `run_command` descriptions.
4. **Shell availability:** prompt gating now uses the resolved shell backend instead of raw settings. The unconditional `mkdir` shell example is also gated.
5. **All eight filesystem helpers:** injection, few-shot filtering, server tool listing, and tests use a shared filesystem tool registry.
6. **Independent filesystem guidance:** filesystem instructions render with shell disabled; write-before-run guidance requires both capabilities.
7. **Static prompts:** few-shot filtering runs before the static/dynamic prompt branch, and the static-path log now reflects the filtered chat prefix.

Also replaced substring matching with complete helper-call matching, consolidated the local availability flags, and added tests against the shipped fixture and the real prepare-node wiring.

**Validation**
- 157 tests passed across the focused regression and related suites.
- Ruff check, Ruff format check, and `git diff --check` passed.
- Two filesystem tests failed and also failed unchanged on the original PR commit (`bfa309e50`): `test_shared_assets_seed_into_empty_thread_workspace` and `test_download_upload_callable_but_not_structured_tools`. These concern workspace seeding and upload paths, respectively.

Sergey-Zeltyn:
Re-verified at the current head (`037ed2b0`) — re-ran the filter against the bundled
fixture, rendered the template in all four shell×fs combinations, and diffed the
rendered prompts against both the original commit and pre-PR. This was a strong second
round: of my earlier points, **1, 2, 3, 4, 7 and all three minors are properly fixed**,
CI is green, and preserving the both-capabilities render byte-identical to pre-PR was
exactly the right instinct. Two remaining asks — nothing I'd call blocking.

**1. The default config now runs with zero few-shot examples.** Whole-transcript
dropping is the right mechanism (verified: the bundled conversation → `[]` with fs
off, intact with only shell off). But find_tools-on + fs-off — the shipped default —
now has no few-shot at all, losing the only `find_tools` demonstration, and
`test_find_tools_and_bundled_examples_follow_separate_flags` asserts the empty result,
which codifies the gap rather than closing it. An fs-free variant of the fixture (same
task shape: `find_tools` → direct tool calls → answer, no file writes) would restore
the demonstration where it's most needed.

**2. The fs-only guidance gap is back — this time on purpose.** `2fabd605` re-nests
the "Sandbox working directory" bullet under the shell gate, so fs-on/shell-off again
injects all 8 helpers with no usage guidance (no write-before-anything, no column-0
rule, no workspace-path guidance). I understand why — the un-nested bullet is what
destabilized the stability runs — but the paragraph that caused trouble is
*shell-flavored* (write-before-**run**, `run_command(...)`, `uv run`). A short
shell-free variant for the fs-only render (how to use
`write_file`/`read_file`/`list_files`, the column-0 rule, relative workspace paths —
with every run/script sentence still shell-gated) would close the gap without
re-exposing the text that caused narration. If you'd rather accept the gap, a one-line
template comment saying so would stop the next person from "fixing" it back.

Sergey-Zeltyn (commented):
Please answer two remaining points above  - either correct or explain why it is reasonable not to correct it

PR #781, opened 2026-09-15
fix(registry): stop an endpoint that rejects every argument set the same way

### LINKED ISSUE #779: [Bug]: an endpoint that rejects every argument set with the same error is never stopped
Part of #616 (epic #611). Follow-up to #599, which shipped the guard these two defects live in.

### What happened

The rejected-call guard from #599 keys on **endpoint + exact arguments**. An agent that
varies one argument on each attempt therefore never repeats a signature and is never
stopped, even when the failure has nothing to do with the arguments.

Measured on AppWorld `test_challenge` hard (195 tasks, `aws/gpt-oss-120b`):

| Rejection message | Count |
|---|---:|
| "The payment card has expired" | **311** |
| "The promo code is not valid" | 138 |
| "Invalid password reset code" | 50 |

In the worst single task: **46 identical failures across six different argument
expressions**, with the guard firing 33 times and **never blocking**. An expired card is
not fixable by editing the request — the agent needed to pick a different card or a
different tool, and nothing in the loop told it so.

### Second, related defect: the tally is cleared by an unrelated success

`record_success` clears **every** counter globally. The Amazon flow interleaves a
successful `add_to_cart` between failing `place_order` attempts, so any endpoint-level
tally resets before it can reach a threshold. Across 60 AppWorld hard tasks this was
almost inert: **270 "card has expired" rejections, 3 blocks.**

The two kinds of claim need different clearing rules:

- *"this exact call failed"* — any state change anywhere can make it valid (the
  precondition fix often lives in another app), so a success clears it globally. Correct
  as-is.
- *"this endpoint rejects every argument set with this error"* — a success **elsewhere**
  is no evidence against that, so it should only be cleared by a success on the same
  endpoint.

### How to reproduce

1. Any task where a precondition outside the request is wrong — an expired payment card,
   an invalid promo code, a consumed reset token.
2. Let the agent retry, varying one argument each time (it does this naturally).
3. The signature differs every attempt, so `rejected_call_block_after` never trips and
   the agent loops until the step limit.
4. Add any successful mutating call on another endpoint between attempts and even a
   working endpoint-level tally is reset.

### Environment

`origin/main` @ `8315704c1`, `src/cuga/backend/tools_env/registry/registry/rejected_call_guard.py`.
Reproduced with `aws/gpt-oss-120b`; the guard logic is model-independent.

### Notes

Related: #704 (closed) and #445 are about the agent giving up too early; this is the
opposite failure — never giving up on an approach that cannot work.

sami-marreed:
Closes #779. Follow-up to #599. Part of #616 (epic #611).

Two commits, kept together because the second is what makes the first fire.

## 1. A third tier, keyed on the error rather than the arguments

The guard from #599 keys on **endpoint + exact arguments**. An agent that varies one
argument per attempt never repeats a signature, so it is never stopped — even when the
failure has nothing to do with the arguments.

AppWorld `test_challenge` hard, 195 tasks, `aws/gpt-oss-120b`:

| Rejection | Count |
|---|---:|
| "The payment card has expired" | **311** |
| "The promo code is not valid" | 138 |
| "Invalid password reset code" | 50 |

Worst single task: **46 identical failures across six different argument expressions**,
guard fired 33 times, blocked **0**.

New tier keyed on `(endpoint, error shape)`. Once
`rejected_call_distinct_args_block_after` (default **3**) *different* argument sets draw
the same error from one endpoint, further calls are refused with a directive to fix the
precondition or find another tool — at that point the arguments demonstrably are not the
problem.

The error shape drops quoted values and numbers, so *"card 176 has expired"* and *"card
178 has expired"* count as one problem, while a genuinely different error starts its own
count.

Both existing tiers, the success-clears-counters rule and the `/api/reset` boundary are
unchanged. `0` disables the new tier.

Also fixes `_endpoint_prefix` serializing non-JSON argument types, which `signature()`
already handled with `default=str`.

## 2. Do not clear the endpoint tally on an unrelated success

On its own the new tier was nearly inert: across 60 AppWorld hard tasks it saw **270**
"card has expired" rejections and blocked **3** times. `record_success` clears every
counter globally, and the Amazon flow interleaves a successful `add_to_cart` between
failing `place_order` attempts — so the tally reset before it could reach the threshold.

The two kinds of tier claim different things, so they must clear differently:

- **Per-signature** — *"this exact call failed."* Any state change anywhere can make it
  valid; the precondition fix often lives in another app. A success still clears these
  globally. **Unchanged.**
- **Endpoint** — *"this endpoint rejects every argument set with this error."* A success
  elsewhere is no evidence against that, so only a success on the **same endpoint**
  clears it.

`record_success` now takes the function name and agent id. Without a function name the
endpoint tier is left intact rather than cleared blind.

## Measured effect

On the 195-task hard run the endpoint tier fires **111** times where the old guard fired
0. Stated honestly: expired-card rejections fell only 523 → 483. It ends the loop rather
than steering the agent to a different tool — worth having, but it is the weaker half of
the pair.

## Tests

`tests/test_rejected_call_guard.py`, 37 passing. Covers: varying arguments still trips the
new tier, a different error starts its own count, numbers and quoted values are normalised
out, a same-endpoint success clears it, a different-endpoint success does **not**, missing
function name leaves it intact, `0` disables it, and both original tiers behave exactly as
before.

## Defaults

`rejected_call_distinct_args_block_after = 3` is the only new setting. Existing
`rejected_call_escalate_after` / `rejected_call_block_after` are untouched.

## Summary by CodeRabbit

- **New Features**
  - Added protection against repeatedly calling an endpoint that returns the same error across different inputs.
  - After three matching failures, further calls are paused with guidance to correct the issue or choose another tool.
  - The threshold is configurable, and setting it to zero disables this protection.

- **Bug Fixes**
  - Rejection tracking now distinguishes individual functions and agents, improving accuracy when managing blocked calls.
  - Successful calls and resets clear relevant rejection state appropriately.


Sergey-Zeltyn (commented):
I replayed the endpoint tier's semantics (same normalized error shape + ≥3 distinct argument sets, tally cleared only by a same-endpoint success) over the **`cuga_mistral_chal`** bundle (AppWorld `test_challenge`, Mistral-Medium-3.5-128B, 417 task-runs, 276 strict passes; the #599 signature tier was live in that run) and checked every cluster by hand. The tier has real false positives there: **8 of 276 passing trajectories (~3%) would have been blocked on the endpoint their winning fix needed.**

### Definite false positives — the successful fix was an argument change on the endpoint the tier locks

| Task (score 1.0) | What happened | Why the block is wrong |
|---|---|---|
| `b71318e_3` | 13 consecutive `POST /venmo/transactions` rejections, distinct receivers, "balance does not have $300" | Tier fires at #3 and refuses the rest. The fix was adding `payment_card_id=2` **on the same endpoint** → 6× "Sent money". The directive "trying another value will not help" is exactly backwards here |
| `b0934aa_1` | 11 `POST /amazon/orders` rejections while **enumerating payment cards** (249, 247, 251, …), "doesn't have enough balance" | This is the PR's flagship error family — and card enumeration *was* the solution: a later card succeeded on the same endpoint. (gpt-oss fails this task every run; Mistral passes it precisely by arg-sweeping) |
| `2ff4dfb_1/2/3` (all 3 runs) | 9–14 `POST /phone/alarms` 422 validation errors on `repeat_days`; the code comments show deliberate format probing ("Let me try day names or numbers") | 422 validation errors are argument errors **by definition** — same shape, distinct args, and varying the args is the only possible fix |
| `906f2c7_3`, `a53a8fd_2`, `d8e490b_3` | 3–4 `GET /file_system/directory` rejections with `directory_path="./"` and distinct substring args, same "Directory not available" shape | Fixed by switching to `directory_path="/"` — same endpoint; the corrected call would be refused |

### Near-misses (passed only by lucky interleaving)

- `3fcc458_1`: download loop over 22 songs, 5 "already downloaded" errors on distinct song ids. Max failure streak happened to be 2, so same-endpoint successes kept clearing the tally — 3 *consecutive* already-downloaded songs kills the rest of the loop.
- `4441ee9_1` ("already unstarred" ×4) and `7434096_1` (returns sweep): same idempotency-sweep class — benign expected errors while iterating a collection; pass/fail depends on element ordering.
- `36705ae_3`: 4× gmail reply 409 ("attachment file does not exist"); the agent fixed the precondition (created the zip) exactly as the directive instructs, then retried the *same endpoint*.

### Two design concerns

1. **The lockout is unrecoverable within a task.** Only a same-endpoint success clears the tally, but a blocked endpoint can never execute, so it can never succeed. Even the agent that follows the directive and fixes the precondition elsewhere (the `36705ae_3` pattern) finds its retry refused. Suggest one of: make the refusal a one-shot advisory that lets the next call through, add a decay/allow-a-probe-after-N-refusals rule, or clear the endpoint tally on a subsequent success anywhere in the same app.
2. **The premise "N distinct arg sets + same error ⇒ arguments are not the problem" fails for two whole error classes**: (a) 422 validation shapes, where the arguments are definitionally the problem, and (b) resource-enumeration errors (insufficient balance / expired card), where sweeping cards is a legitimate — and in this bundle *winning* — strategy. Exempting shapes that contain `Validation error` (or a field name) and/or not counting a shape whose distinct-arg set is still growing productively would remove most of the 8 FPs while keeping the 46-identical-failures pathology this PR rightly targets.

Caveat: this is the Mistral challenge split, not the gpt-oss hard split the PR was measured on — models differ in how much *productive* arg-sweeping they do, and Mistral's sweeps are exactly what lets it pass tasks gpt-oss loops on. So the tier's measured benefit on gpt-oss and its FP cost on Mistral are both real; the threshold/exemption design should account for both.

Sergey-Zeltyn:
Follow-up to my review comment: I ran the same replay over the **gpt-oss-120b** bundles (82 bundle dirs, 1,561 passing trajectories) — the same model family this PR was measured on. The false positives are fewer in rate (~0.7–0.8% of passing runs vs ~3% on Mistral) but qualitatively worse: **they hit the PR's own flagship error shapes inside winning strategies.**

### Verified false positives in passing gpt-oss runs

| Task | Runs affected | What happens |
|---|---|---|
| `7574325_1` (change Venmo password) | **7 separate passing runs** (3–22 rejections each) | Agent extracts stale codes from old emails → "Invalid password reset code" with a **different code each attempt** (distinct args, same shape). It then requests a fresh code — a *different* endpoint, so nothing clears — and finally succeeds **on the same `reset_password` endpoint**. The tier locks at the 3rd wrong code, so the correct fresh code is refused. This is the motivation table's #3 error (50 occurrences): on these runs the retry-with-fresh-code loop is how the task ever passes |
| `6474048_1` (buy microwave) | 1 | 3 distinct cards → "The payment card has expired" (the flagship error), then the code filters for a non-expired card → success. The threshold is reached at exactly the 3rd card; the valid-card call is the 4th → refused |
| `b0934aa_1` (board game) | 1 | 4× "doesn't have enough balance" while enumerating cards → later success on the same endpoint (same pattern as the Mistral run of this task) |
| `277d81d_1` (mark inbox read) | 1 | Sweep of 16 threads, outcome sequence `OK FAIL FAIL FAIL OK OK FAIL FAIL FAIL OK OK FAIL×5` — a streak of 5 "already marked as read" errors with distinct thread ids. The tier fires at the 4th call and refuses the ~12 remaining threads |
| `ccf4b82_1`, `bcb9696_1` | 2 | id-enumeration errors ("card id # does not exist" ×3, "attachment id # does not exist" ×4) and a `directory_path="./"` probe — same families as on Mistral (likely FPs, not step-verified) |

Total: 10 verified + ~3 likely, out of 1,561 passing gpt-oss runs.

### Two observations beyond the Mistral data

1. **The unrecoverable-lockout concern is now demonstrated on real data.** `7574325_1` does exactly what the block directive instructs — fixes the precondition with a different call (requests a fresh reset code) — and its retry lands on the locked endpoint, which nothing can ever clear. Under this PR, all 7 passing runs of that task flip to failure.
2. **There is a false-negative counterpart to the shape normalization.** In `5a83b05_1`, a double-slash path bug fails `DELETE /file_system/file` for 10+ files, but each file's path is embedded *unquoted* in the message, so every file becomes its own shape with a single arg signature — the tier never sees the actual "endpoint rejects every argument set" pathology it was built for. The normalization is simultaneously too coarse for enumeration errors (codes/cards/ids, where distinct args are the productive search dimension) and too fine for path-bearing messages (where the args genuinely don't matter).

Together with the Mistral findings, the FP mechanism is model-independent and concentrated in three legitimate patterns: argument search (cards, codes, path formats), idempotency sweeps ("already read/downloaded/unstarred"), and fix-precondition-then-retry. A validation/enumeration exemption plus some way for a locked endpoint to recover (one-shot advisory, decay, or clearing on a same-app success) would address nearly all of them while keeping the 46-identical-failures case this PR rightly targets.

PR #782, opened 2026-09-15
fix(cuga-lite): scope the VERIFY ungrounded rule to facts, not choices

sami-marreed:
Amends the gate merged in #739. Part of #616 (epic #611).

## The problem

Rule 1 was written for **invented facts** — an id or an amount the agent made up — but
reads as *"any literal"*. The verifier flagged `quantity=1`, `clear_cart_first=True`,
`snooze_minutes=0`, `page_limit=20` and regex patterns as ungrounded, and **the model
complied**. Dropping `clear_cart_first` shipped a pre-seeded cart as a **7-item order**;
another task lost `repeat_days`.

## Evidence

Measured over **6,421 trajectory files / 40,841 real gate decisions** (38,928 `ok`,
1,913 `revise`, 259 fail-open), of which 1,234 cite rule 1:

| What was flagged | Count | Should it be? |
|---|---:|---|
| Genuine ungrounded ids, dates, emails, names | **972 (79%)** | Yes — must keep being caught |
| Booleans and flags (`clear_cart_first`, `enabled`, `vibration`) | 108 | No |
| Pagination and code constants (`page_limit`, `skip`, regex, `%H:%M`) | 57 | No |
| Quantity-style (`quantity=1`, `snooze_minutes=0`) | **148 across 53 tasks** | No |

## What changed

Two commits.

**1. Scope rule 1 to values that assert something about the world.** It now covers a
resolved argument that *identifies a record or states a quantity* — an id, a name, an
email, an amount, a date — rather than any fixed literal. Adds "a value that appears
verbatim in the most recent execution output is grounded", which the old wording missed
when a block did not re-read it. Rule 4 now requires the task to **explicitly** restrict
which fields to change, so a field that follows from the task is not "extra". The
`Do NOT flag` list gains booleans and flags, documented enum options and defaults,
names/paths the agent is creating, and code constants.

**2. A tie-breaker for quantities.** Rule 1 names "an amount" while the carve-out excludes
choices, and `quantity=1` satisfies both. In the measured data the model resolved that
conflict by flagging — the largest remaining false-revise family. A blanket carve-out
would be wrong in the other direction: `quantity=12` derived from *"9 friends + 3
roommates"* is a **correct** catch. So the rule separates a number the task **implies**
(a count of people, items or shares, which must match the rows it came from) from one it
neither states nor implies, where the API needs a value and the agent is picking one.

## The change is one-directional

Every edit is strictly **more permissive** — it removes things from scope, never adds. The
79% of revises that catch genuine ungrounded ids, dates, emails and names all remain in
rule 1's wording, so there is no path by which this makes the gate miss something it
catches today.

## Disclosure

An **earlier version of this was reverted** on the `feat/pre-execute-verify` branch before
merge (`a1cd607f`). The revert commit records no reason, and I do not have one — if you
had a specific objection, it should be re-applied to this version. What is here is not
that patch: it narrows rule 1 to *identifies a record or states a quantity* rather than
enumerating fact types, adds the execution-output grounding clause, tightens rule 4, and
adds the quantity tie-breaker, which the reverted version did not have.

## Blast radius

Prompt-only. `pre_execute_verify_enabled` ships `false`, so only deployments that opt in
see any change.

## Tests

`test_pre_execute_verify.py`, 104 passing — 6 new, pinning each rule change against the
rendered prompt so a future edit cannot silently drop them.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation of write operations by evaluating actual resolved values instead of source text.
  - More accurately checks identifiers, quantities, dates, amounts, and task-derived counts against available results.
  - Recognizes values from the latest execution output as validly grounded.
  - Reduces false positives for documented choices, code constants, defaults, and regular expressions.
  - Applies extra-field and aggregation checks only when explicitly required by the task.

- **Tests**
  - Added coverage for grounding, quantitative validation, defaults, field restrictions, and aggregation behavior.


Sergey-Zeltyn:
The false-revise data is convincing and the quantity carve-out is exactly how this
should be done — scoped, with an implied-vs-chosen guard and its own worked example.
My concerns are that the other carve-outs didn't get the same care, and that the one
mechanical clause in the PR is defeated by the context assembly. Ordered by impact.

## Main issue

**1. Head-truncation drops exactly the output the new grounding clause relies on
(`verify_system.jinja2:14` + `context_management_utils.py:23`)**
Execution output reaches the verifier only inside Conversation history
(`sandbox_node.py:335-348` appends it to `chat_messages`), and
`truncate_text_for_context` keeps `trimmed[:max_chars]` — the **head**. So once history
exceeds the ~30k verify limit, "the most recent execution output" is precisely the text
removed from the prompt, and the clause goes inert on the long-history runs it was
written for (`pre_execute.py:35-39` itself notes the largest write blocks have the
longest histories) — the same false revise survives, hidden by short-history repros.
In the untruncated case the sentence is redundant: the output is already in
Conversation history, which rule 1 already exempts. CodeRabbit flagged the truncation
half of this too.
*Suggested fix:* pass the latest execution output as its own section of the verify
context, as `prepare_reflection_context` already does (`coder_agent_output`) — or
truncate history from the front. Either makes the sentence true; today it isn't.

## The carve-outs — this is a gate on writes, so each widened permission is a miss path

**2. "An enum option the API documents" / "the API's own documented default"
(`verify_system.jinja2:38-39`) are unverifiable from the verifier's seat.**
`verify_user.jinja2` passes task / history / variables / code / write-args — no tool
schemas. The model cannot test documented-ness, so it can only assume any option-shaped
literal *is* documented, widening the carve-out to "any enum-shaped string": an invented
`status='approved'` on a tool that only allows pending/accepted gates ok, untraceably.
Either pass the mutating tool's parameter schema into the context, or word the
carve-out in terms of evidence the prompt actually contains.

**3. "Booleans and flags" and "a name or path the agent is creating" are unconditional
(`verify_system.jinja2:37-38`).** Your own quantity bullet shows you knew blanket
carve-outs are wrong ("quantity=12 derived from 9 friends + 3 roommates is a correct
catch") — but these two got no such guard. `add_to_cart(clear_cart_first=True)` against
"without emptying my cart" is now unflaggable (rule 2 only compares retrieved values,
and your own data shows the old gate caught booleans); `create_playlist(name="Workout
Mix")` for a task naming "Morning Run" is excused as "a name the agent is creating" —
the retained line-32 bullet's condition (*the user asked for that name*) was the
protection, and the new bullet supersedes it. Also worth considering: booleans are
deterministically identifiable in `describe_write_arguments` (one `isinstance` on the
folded constant), so that filter could live in code instead of prose.

**4. Rule 1's rescope (`:9`) drops ungrounded categorical strings entirely.** Old text
covered any fixed number or string; the new enumerated list (id, name, email, amount,
date) leaves status values, phone numbers, addresses, and URLs to no rule at all —
`update_order(id=123, status='cancelled')` for "mark order 123 as shipped" has no rule
left to fire (and bullet 38 actively excuses it as an enum choice).

**5. Rule 4's escape hatch (`:27`) swallows the rule.** "A field whose value follows
from … retrieved data is not 'extra'" exempts by value-provenance — but nearly every
extra field an agent writes is echoed from something it read (e.g. a stale
`status='pending'` copied into an "only change the title" update). And
"clearly restricted" → "explicitly restricts" reads as excluding implicit one-field
restrictions ("update the email").

## Consistency

**6. `pre_execute.py:158`**: `verify_blocked_message` still says "Rewrite the block so
each write argument is grounded in retrieved data" — the blanket instruction the new
carve-outs repudiate, on the path where your PR comment says "the model complied" at
the cost of two tasks. It should mirror the new scope.

**7. No precedence between `:44` and `:14`**: an implied count "stays in scope", but
the digit it was wrongly replaced with (`quantity=1`) virtually always appears verbatim
in the latest output (`"page": 1`), which line 14 declares grounded. Two rules, opposite
verdicts, no arbiter — your own measurement (models resolve rule conflicts
unpredictably) is the reason this matters.

Sergey-Zeltyn (commented):
see comments above

PR #783, opened 2026-09-15
perf(policy): skip output-formatter LLM call for clean plain-text responses

mbwhite:
## Feature

Closes #

### Summary

The `output_formatter` policy (e.g. `"Plain English responses"`) uses a
`natural_language` trigger with value `"any agent response"` and
`threshold = 0.7`. Because this phrase is semantically close to every agent
response, the vector-search match fires unconditionally. When the response is
already clean plain text or markdown the formatter LLM simply echoes it back
unchanged — wasting ~1 LLM call (~4 s, ~$0.007) per request for zero benefit.

**Evidence** (experiment run `20260915_1418`): log lines 7885–7934, verbatim
LLM reply: *"The response is already in clean plain text… returned unchanged."*

**Root cause:** `_enact_format_output()` in `PolicyEnactment` invoked the
formatter LLM for `format_type="markdown"` regardless of whether the response
contained raw JSON or was already human-readable.

**Fix (Fix A):**
- Add a module-level `_RAW_JSON_PATTERN` regex and `_response_needs_formatting()`
  helper in `enactment.py`.
- Add an early-return guard in `_enact_format_output()`: when `format_type` is
  `"markdown"` **and** the response does not start with `{` / `[`, return
  `(None, None)` immediately — no LLM call made.
- Only raw-JSON-looking responses reach the formatter LLM, which is the only
  case that genuinely needs reformatting.

**Note — Fix B (out of scope):** Narrowing the policy trigger from
`natural_language` to `keyword` (so the matching LLM call also doesn't fire) is
a policy-config change, not a framework change, and is left to individual
deployments.

**Expected improvement:** Eliminates ~1 LLM call (~4 s, ~$0.007) per request
for well-formed plain-text / markdown responses.

### Testing

- Updated `test_markdown_prompt_prioritizes_format_config_over_preserve_rules`
  to use a JSON-containing response (the guard must not fire there — that test
  exercises prompt content, not trigger logic).
- Added `test_markdown_skips_llm_for_clean_plain_text` — asserts the formatter
  LLM is not called when the response is already clean plain text.

---
Helped-by: IBM Bob

## Summary by CodeRabbit

* **Performance**
  * Improved markdown output handling by skipping unnecessary formatting for clean plain-text responses when no content changes are required.
  * Continued processing responses containing raw JSON through the formatter.
  
* **Safety**
  * Ensured policies that redact, replace, withhold, block, remove, or filter content continue applying those changes, even when the response is already clean plain text.

* **Tests**
  * Added coverage for content-modifying policies and clean plain-text responses.


mbwhite:
Thanks for the review @coderabbitai.

**Security comment (CWE-200):** Fixed in the follow-up commit. Added `_CONTENT_MODIFYING_PATTERN` and `_is_presentation_only()` — the early-return guard now requires the `format_config` to contain no redaction/replacement/blocking keywords before skipping the LLM. Policies containing `replace`, `redact`, `withhold`, `block`, `remove`, or `filter` always run the formatter. Added `test_markdown_content_modifying_policy_runs_llm_for_plain_text` covering all 6 keywords.

**Missing `@pytest.mark.unit`:** Already present on `test_markdown_prompt_prioritizes_format_config_over_preserve_rules` (added via an external edit before this comment arrived).

offerakrabi (commented):
## PR Review: #783 — perf(policy): skip output-formatter LLM call for clean plain-text responses

@mbwhite thanks for this, i like the idea of trying to save llm calls.
however, if you look at the first finding, the way the formatter decides whether it is 'presentation-only' is problematic, it can cause policy instructions to be ignored.
the way i see it, unless we find a better way to define/check for 'presentation-only' it is better to make one more LLM call than to ignore a policy instruction, a couple of examples:

```
Numbers must be written as words

Policy instruction: “Don't use digits; write numbers using words.”
Original response: “The total is 42.”
Expected: “The total is forty-two.”
Actual: “The total is 42.”
Translate the response

Policy instruction: “Translate the response to French.”
Original response: “The order is ready.”
Expected: “La commande est prête.”
Actual: “The order is ready.”
```

Previously raised items: the named redaction/replacement bypass is guarded, and the changed test already has its unit marker; the broader instruction bypass remains as #1.

**Needs work.** Two critical items; the fast path changes formatter semantics and the PR currently breaks a default-CI unit test.
Approve once #1 preserves every configured transformation and #2 passes.

|---|---|---|---|---|
| 1 | Critical | Formatter instructions silently bypassed | `src/cuga/backend/cuga_graph/policy/enactment.py:_is_presentation_only` | Any markdown instruction lacking six hard-coded verbs is skipped for plain-text output → valid transformations such as spelling out digits or adding required sections never run. |
| 2 | Critical | Existing formatter test now fails | `tests/unit/test_langfuse_tracing.py:test_output_formatter_ainvoke_receives_callbacks` | The new early return bypasses the mocked invocation expected by an existing unit test → the PR's default unit-test shard fails. |
| 3 | Minor | JSON detector matches ordinary Markdown | `src/cuga/backend/cuga_graph/policy/enactment.py:_RAW_JSON_PATTERN` | A citation, link, bracketed aside, or code line beginning with `[` or `{` is treated as raw JSON → the formatter call and its latency remain for valid Markdown. |

## Findings

## 1 — Formatter instructions silently bypassed (Critical)
`src/cuga/backend/cuga_graph/policy/enactment.py:38-57`

The fast path calls a policy “presentation-only” whenever its free-form instructions omit six verbs, then leaves every non-JSON response unchanged. Markdown policies accept arbitrary instructions, so “write numbers as words” applied to “Revenue is 42” now returns “Revenue is 42” without invoking the formatter. Existing policies that require headings, citations, translation, tone, or unlisted redaction wording are bypassed the same way.

---

## 2 — Existing formatter test now fails (Critical)
`tests/unit/test_langfuse_tracing.py:361-399`

The existing tracing test configures “Use bullet points” and expects the formatter invocation to receive callbacks. The new guard returns before that invocation for its plain-text response, so `metadata` is `None` and the test fails at line 396. Running this test on the PR head reproduces the failure; [`tests/unit/`](tests/unit) is collected by the default `unit-b` CI shard.

---

## Questions

1. Should the optimization be limited to an explicit policy capability that guarantees the response is already compliant, rather than inferring a no-op from arbitrary natural-language instructions?

## Notes

- The PR body contains `Closes #` without an issue number, so no linked issue or acceptance criteria were available.
- The new formatter tests pass (`4 passed`), but the combined relevant unit run reports `1 failed, 30 passed`; the SDK policy subset could not initialize locally because no OpenAI credentials were configured.

PR #786, opened 2026-09-18
fix(events): Deleting an agent (that was loaded from a roster yaml) in the Manage UX didn't stick.

anupamamurthi:
Deleting an agent in the Manage UX didn't stick — the roster YAML was re-imported on every boot, and a missing row looks identical to one that was never seeded, so it came back.

The YAML is baked into the image and under git, so it can't be the live source of truth. It's a first-boot seed; after that the DB wins. The seeder now fingerprints the roster and skips re-seeding while that hash is unchanged. Edit the YAML on purpose and it re-applies, which is still how you ship a roster update to a live deployment.

Deletes also apply immediately: delete_agent strips the ref from the supervisor's subAgents, which bumps the config version and invalidates the cached supervisor. And locally, events_up.sh now sets the preserve flag, since cuga start demo was wiping cuga.db on every boot anyway.

All behind the events switch, so core CUGA is unchanged. Review also turned up a test calling reset_config_db() that was deleting your real src/cuga/dbs/cuga.db — conftest isolates it now.

## Summary by CodeRabbit

- **New Features**
  - Roster imports now run only when the roster changes, preserving dashboard edits across restarts.
  - Changed rosters continue to update applicable agents while protecting unrelated manual edits.
  - Deleted agents are automatically removed from supervisor references.

- **Bug Fixes**
  - Improved startup resilience when optional environment values or tunnel logs are missing.
  - Supervisor updates are protected from being overwritten during concurrent agent deletions.
  - Startup configuration now supports preserving existing configurations according to the selected environment.


anupamamurthi:
I created a PR without description first and CodeRabbit dinged it saying poor readability :D 

Anyway, this is a small bug that surfaced while testing events/channels workstream - I added/edited/deleted an agent and the changes were persisted as expected but given that I was loading / prioritizing the roster yaml, on a restart, the deleted agents will re-appear.

This PR fixes this experience and hardens it.

